PR #309, opened 2022-02-21
First sketch of a github action for building

bosilca:
For now this will be used as a playground for adding support to Github actions.


.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,62 @@
...
+
+#env:
+
+jobs:
+  build:
+    runs-on: [self-hosted, Linux]
+    strategy:
+      fail-fast: false
+      matrix:
+        build_type : [ Debug ]
```
therault, line 13: We don't want to test Release at this time? Or Profiling etc...?

This could come in an additional PR to move forward.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,79 @@
+# For most projects, this workflow file will not need changing; you simply need
```
therault, line 1: I thought you had removed this? Or this is just not activated in the github interface?

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @bosilca
```
abouteiller, line 1: do we need to get more people here?

abouteiller (approved):

.github/workflows/build_cmake.yml
bosilca: It was taking too long on my machine, so I removed everything but one build. Once we have more runners we can enable a wider range of tests.

.github/CODEOWNERS
bosilca: There are two aspects that need to be considered. First, it is a security risk to allow people to change what is executed on our runners. Second, it remains a critical infrastructure that we want to modify carefully. That being said, this simple means that any code altering this part will need my blessing for now, and as we are getting more familiar with the runners we can increase this list.

therault (approved):

PR #310, opened 2022-02-24
Add an Open MP target device [BBT#199]

abouteiller:
This is a transfer of https://bitbucket.org/icldistcomp/parsec/pull-requests/199

* Rename gpu_* in device_cuda.[ch]
* Add a core dgemm that executes on hosted MIC with OpenMP target
* Add an OpenMP target body to LLT
* Configury to make it compile in target mode with icc

How to try/use:

Compile PaRSEC using either
* Intel ICC [17.xx](http://17.xxx/) (not 18.xx) with MPSS 3.8 → executes on KNC Mic
* gcc > 7.3 with the Open ACC PTX enabled → executes on Nvidia
* gcc > 7.3 with the Mic emulation device enabled → executes on Host in Mic emulation mode


abouteiller:
Need to be updated to latest PaRSEC.

PR #314, opened 2022-02-24
[BBT#493] New API for parsec_init w/o cores parameter

abouteiller:
https://bitbucket.org/icldistcomp/parsec/pull-requests/493

- [x] add info comm=%p to permit selection of comm before init
- [ ] add parsec_info_create, use parsec_info_set to pass-in arguments to parsec before parsec_init
- [x] Dplasma corresponding request
- [x] rename setenv_mca_param_int to something less obnoxious
- [x] include definition of setenv_mca_param_int from public API
- [ ] decide what to do with argc, argv
    * We will keep argc/argv so that command line parameters can be 'cleaned-up' and return only argc/argv that PaRSEC didn't consume, it should be valid to pass NULL, NULL.
- [x] update copyrights

### DIFF
--- a/examples/Ex00_StartStop.c
+++ b/examples/Ex00_StartStop.c
@@ -7,7 +7,7 @@
  *
  * [mpicc|cc] -o Ex00_StartStop Ex00_StartStop.c `pkg-config --cflags --libs parsec` -lpthread -lm -lrt [-ldl -lcudart]
  *
- * @version 3.0
+ * @version 4.0
  * @email parsec-users@icl.utk.edu
  *
  */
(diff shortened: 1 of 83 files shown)


bosilca:
Discussion on 10/07/22: Add support for a simple API (based on the MPI Info) and we pass this as the last argument to `parsec_init`. We keep the pointers to argc and argv, but we do not change them. Look at issue #431 for more info.

therault:
the info object should be a local that is passed explicitly to the `set` functions, and to the `parsec_init` call.

the developer should have a simple way of listing the info keys.

bosilca:
As discussed on 03/31/23 we need to merge this with #443 and get them in.

PR #315, opened 2022-02-24
[BBT#536] Add AMD RoCM/HIP device

abouteiller:
https://bitbucket.org/icldistcomp/parsec/pull-requests/536

- [ ] This is in a working state as tested on Spock
- [ ] For the CI: none is defined by default, so the PR fails. We should change the JenkinsFile to do both CUDA and HIP independently
- [ ] For the CI: add a HIP tester directly in PaRSEC (rather than testing with DPLASMA)
- [ ] For the user: even if we don’t have the HIP framework, we fail with the error. We should check if HIP is available before failing (i.e. if the user did not define a preference, but there is only CUDA or only HIP or none of them, CMake should succeed with the only available choice per default. Only if there is both CUDA and HIP and the user did not specify a preference should we fail).
- [ ] We want to split the hip/cuda functions from the main generated file so we don't have include files conflicts in the generated code


bosilca:
Will be merged on Monday.

PR #316, opened 2022-02-24
[BBT#582] bugfix/atomic lifo: The offsetof was incorrect leading to lifo padding being wrong in external lifo

abouteiller:
https://bitbucket.org/icldistcomp/parsec/pull-requests/582

atomic lifo: The offsetoff was computed incorrectly because it didn't account for alignment and that caused static initializers to overspill

This would cause DPLASMA ctest to fail in spectacular ways. 

Tested with C11 128bit atomics and 64bit CAS (gcc/7.3.0)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1008,7 +1008,11 @@ check_type_size("parsec_lifo_t" PARSEC_LIFO_OPAQUE_SIZEOF)
 include(CheckStructureFieldOffset)
 # NB: the structure below *needs* to be kept in sync with the structure
 # in lifo-external.h
-check_structure_field_offset("lifo_private" "struct{parsec_object_t super;uint8_t alignment;char lifo_private;}" PARSEC_LIFO_HEAD_OFFSET)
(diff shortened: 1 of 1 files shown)


devreal (approved):

bosilca:
I don't think there is anything wrong with the offset of the private field, the root cause of the issue is lack of alignment of the high-level structure. Here is the reason: the internal struct uses __int128 the external lifo struct must be aligned by 128. Unfortunately, the current detection test will have the struct aligned to 64, and as a result the alignment will be incorrect and thus the size of the external struct. The correct solution would be specify the alignment of the external LIFO struct, but as the C standard does not currently define a standard way, we need another solution. As this external version of the LIFO structure is supposed to be used with external compilers (aka. not used by PaRSEC's detection), we should be careful with the types used. But, I think we can accept this patch, as we already have the int128 in the definition of this struct in the exposed header file, we can use it in the alignment detection.

bosilca (approved):

Issue #295, opened 2021-06-30
Cray `ftn` cannot accept LLVM flags


abouteiller:
**[Original report](https://bitbucket.org/icldistcomp/parsec/issue/295) by me.**

----------------------------------------

## Problem

Fortran flags are wrong on Spock

## Details

In another instance of the wrongness of detecting compilation flags only with CC, when using Cray Fortran compiler \(as seen on OLCF Spock\), we get the following error

```shell
[ 10%] Building Fortran object parsec/CMakeFiles/parsec.dir/fortran/parsecf.F90.o
ftn_driver.exe: invalid option -- '3'
ftn-2105 crayftn: ERROR in command line
  "-g" is an invalid command-line option.
ftn-2103 crayftn: WARNING in command line
  The -W all option is not supported or invalid and will be ignored.
ftn-2103 crayftn: WARNING in command line
  The -W extra option is not supported or invalid and will be ignored.
make[2]: *** [parsec/CMakeFiles/parsec.dir/build.make:1553: parsec/CMakeFiles/parsec.dir/fortran/parsecf.F90.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1719: parsec/CMakeFiles/parsec.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
 ~/parsec/dplasma/parsec/spock.debug  master ?1 ·························································································· bouteilla@login1  15:29:47 ─╮
❯ ftn --version                                                                                                                                                           ─╯
Cray Fortran : Version 11.0.4
```

## Mitigation

`configure --disable-fortran`

## Solution

We should test individually all flags with the fortran compiler, including ‘obvious’ ones like `-g`, `-W` etc.

‌

‌

abouteiller:
fix merged

PR #326, opened 2022-03-02
Per-language compiler flags

therault:
Use generator expressions to only apply compiler flags for the
language they have been tested with. Try all languages for each
flag, when it makes sense. Provide a new function to do so
in a convenient way.

Closes issue #295


bosilca (approved):

abouteiller (approved):

abouteiller:
Same changes need to be carried to DPLASMA

abouteiller:
@therault  you need to sign your comments `git ci --amend -s`

PR #328, opened 2022-03-02
Hicma/recursive

bosilca:
Fix the parsec handling of recursive DAGs.

First call the associated callback, the complete the initiator task and only then release the temporary data.

Thanks @omor1 for finding the bug and proposing a solution.

### DIFF
--- a/parsec/recursive.h
+++ b/parsec/recursive.h
@@ -12,57 +12,59 @@
 #include "parsec/mca/device/device.h"
 #include "parsec/data_dist/matrix/matrix.h"
 
-typedef struct parsec_recursive_cb_data_s {
+typedef struct parsec_recursive_callback_s parsec_recursive_callback_t;
+typedef void (*parsec_recursive_callback_f)(parsec_taskpool_t*, const parsec_recursive_cb_data_t* );
+
+struct parsec_recursive_callback_s {
     parsec_execution_stream_t    *es;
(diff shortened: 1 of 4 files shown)


parsec/recursive.h
```diff
@@ -12,42 +12,44 @@
...
+    rc = __parsec_complete_execution(data->es, data->task);
 
-    for(i=0; i<data->nbdesc; i++){
+    for( i = 0; i < data->nbdesc; i++ ) {
         parsec_tiled_matrix_destroy( (parsec_tiled_matrix_t*)(data->desc[i]) );
         free( data->desc[i] );
     }
 
-    data->destruct( tp );
+    data->callback( tp, data );
```
omor1, line 38: This must be called before `__parsec_complete_execution` or you can get some nasty races where `data->task` is no longer valid inside of `data->callback`.

omor1 (changes requested):

parsec/recursive.h
bosilca: I cannot find the patch I sent you for this a while back. Do you have it around ?

omor1: You can cherry-pick 1e15cdd3c2c82e5e4b32d5c85e44936442df67fe from my fork.

omor1 (approved):
Looks good to me!

Issue #278, opened 2021-01-08
Missing support when task class execution scope from a .. b where b < a


abouteiller:
**[Original report](https://bitbucket.org/icldistcomp/parsec/issue/278) by Qinglei Cao (Bitbucket: [Qinglei_Cao](https://bitbucket.org/Qinglei_Cao), ).**

----------------------------------------

If the execution scope of a task class is defined like this:

```
 281 potrf_dtrsm(m, k) [high_priority = on]
 282 
 283 // Execution space
 284 k = 0 .. descA->mt-2
 285 m = [ mi = 0 .. %{ return analysis->trsm_num[k]-1; %} ] %{ return analysis->trsm[k][mi]; %}
```

where `analysis->trsm_num[k]` could be 0 \(mi = 0 .. -1\), and in this case no task exist for this iteration k.

However, it seems it’s not the case, at least in the internal\_init when counting the number of local tasks. For instance,

```
static int HiCMA_dpotrf_L_2flow_sparse_potrf_dtrsm_internal_init(parsec_execution_stream_t * es,
                                                                 __parsec_HiCMA_dpotrf_L_2flow_sparse_potrf_dtrsm_task_t
                                                                 * this_task)
{
...........

    for (k = __jdf2c_k_start; k <= __jdf2c_k_end; k += __jdf2c_k_inc) {
        assignments.k.value = k;
        {                       /* block for the local variable 'mi' */
            int __jdf2c__loop_mi = 1;
            int mi;
            for (assignments.ldef[0].value = mi = 0; __jdf2c__loop_mi == 1; assignments.ldef[0].value = mi += 1) {      /* Arbitrary iterator on mi */
                if (mi == HiCMA_dpotrf_L_2flow_sparse_potrf_dtrsm_inline_c_expr26_line_285(__parsec_tp, &assignments))
                    __jdf2c__loop_mi = 0;       /* Execute once only when reaching the end */
...........
}
```

BTW, this issue also happens in the startup for each taskclass.

abouteiller:
pr has been merged, can this be closed @therault ?

PR #329, opened 2022-03-02
Use normal for loops to iterate over local index variables when they are a range.

therault:
Use 'normal' for loops to iterate over local index variables that are a range.
This enables the possibility to have empty ranges, while the previous
code would not (issue #278).


bosilca:
The generated loops look  little cleaner and use less temporaries, but I can't see any other impact on the code. It seems the loop index still needs to be increasing in value in the loop, preventing loops going backward on the value range.

therault:
More metaprogramming would be desirable in jdf2c.c: all these loops should be generated using a same function that gives all the loop the same bounds, approaches, etc...

bosilca (approved):

PR #333, opened 2022-03-07
GitHub runner with spack

bosilca:
A step toward a spack-based github runner for parsec.

### DIFF
--- a/.github/CI/spack_setup.sh
+++ b/.github/CI/spack_setup.sh
@@ -1,0 +1,75 @@
+# This file should be "sourced" into your environment
+# to set up the spack repository
+
+SPACK_DIR=${HOME}/spack
+
+sload () {
+  # Installs a software package and loads it into the environment
+  spack install --reuse --fail-fast $@
+  spack load --first $@
+}
+
+slocation () {
+  # Returns the installation directory of a given software package
+  HASH=`spack find --no-groups --loaded -l $@ | head -1 | awk '{print $1}'`
+  spack location -i /$HASH
+}
+
+# Only do the heavy lifting once per github_action invocation
+if [ ${GITHUB_ACTION} != "setup" ]; then
+  source $SPACK_DIR/share/spack/setup-env.sh
+  spack env activate ${RUNNER_ENV}
+  return
+fi
+
+save_dir=`pwd`
+#
+# Only update the spack clone once a day
+#
+echo "::group::Spack environment setup"
+if [ -r ${SPACK_DIR}/.git/FETCH_HEAD ]; then
+  last_update=`stat -c %Y ${SPACK_DIR}/.git/FETCH_HEAD`
+  current=`date -d now "+%s"`
+  if [ $(((current - last_update) / 86400)) -gt 1 ]; then
+    echo "Last update ${last_update}, current ${current}: Do git pull spack"
+    cd $SPACK_DIR && git pull
+  else
+    echo "git pull was less than one day ago"
+  fi
+else
+  echo "git clone spack"
+  git clone https://github.com/spack/spack $SPACK_DIR || true
+  # We do a git pull to create the .git/FETCH_HEAD for the check
+  # next time the script is executed.
+  cd $SPACK_DIR && git pull && git status
+fi
+
+echo "Load spack environment"
+source $SPACK_DIR/share/spack/setup-env.sh
+
+spack external find
+spack compiler find
+
+# Start with a fresh env every time
+spack env remove -y ${RUNNER_ENV}
+
+cd ${save_dir}
+mkdir ${RUNNER_ENV} || true
+
+# Show the known envs
+if spack env list | grep ${RUNNER_ENV}; then
+  if diff ${GITHUB_WORKSPACE}/.github/CI/${RUNNER_ENV}.yaml ${SPACK_DIR}/var/spack/environments/${RUNNER_ENV}/spack.yaml > /dev/null; then
+    echo "Update spack env ${RUNNER_ENV}"
+    cp ${GITHUB_WORKSPACE}/.github/CI/${RUNNER_ENV}.yaml ${SPACK_DIR}/var/spack/environments/${RUNNER_ENV}/spack.yaml
+  fi
+  spack env update ${RUNNER_ENV}
+else
+  echo "Create spack env ${RUNNER_ENV}"
+  spack env create ${RUNNER_ENV} ${GITHUB_WORKSPACE}/.github/CI/${RUNNER_ENV}.yaml
+fi
+spack env activate ${RUNNER_ENV}
+spack concretize --force
+spack install --reuse
+
+echo "::endgroup::"
+
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,4 @@
-* @bosilca
+* @ICLDisco/staff
 contrib/pregen_flex_bison.tar @ICLDisco/Staff
+.github/workflows @bosilca
+
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -1,26 +1,123 @@
-name: Linux/MacOS Build
+name: CI Build
 
-on: [push, pull_request]
+# Triggers on push and branches on the master
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
 
 #env:
 
+# Allows to run this workflow manually from the Actions tab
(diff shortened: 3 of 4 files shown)


.github/CODEOWNERS
```diff
@@ -1,2 +1,3 @@
-* @bosilca
+* @staff
```
abouteiller, line 1: Should be @ICLDisco/staff

therault: The github pull request interface shows the following warning:


This CODEOWNERS file contains errors:
    Unknown owner on line 1: make sure @staff exists and has write access to the repository

    * @staff

therault (approved):
No other comment, the rest looks good.

How to add CUDA to that? There is no spack package for CUDA I assume?

devreal:
There is a warning in the test:

```
Run source .github/CI/spack_setup.sh
fatal: destination path '/home/github_runner/spack' already exists and is not an empty directory.
```

Also, the spack env installs OTF2 but  `PARSEC_PROF_TRACE` is disabled. I think we should build with at least one profiler enabled, ideally testing all three (none, built-in, otf2)

It also seems that HWLOC is not detected even though HWLOC is installed through spack:

```
Module LFQ not selectable: does not have HWLOC
```

devreal (approved):

.github/workflows/build_cmake.yml
```diff
@@ -1,54 +1,150 @@
...
+
+    name: "Type = ${{ matrix.build_type }} shared=${{ matrix.shared_type }} profiling=${{matrix.profiling}}"
+    env:
+      BUILD_DIRECTORY : "${{github.workspace}}/build/${{ matrix.build_type }}/shared_${{matrix.shared_type}}/profile_${{matrix.profiling}}"
+      INSTALL_DIRECTORY : "${{github.workspace}}/install/${{ matrix.build_type }}/shared_${{matrix.shared_type}}/profile_${{matrix.profiling}}"
+      RUNNER_ENV : github_runner
+      BUILD_CONFIG : >
+        -G Ninja
+        -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+        -DPARSEC_DEBUG_NOISIER=OFF
```
abouteiller, line 109: This is the default so we don't have to set it

.github/workflows/build_cmake.yml
```diff
@@ -1,54 +1,150 @@
...
     env:
-      BUILD_DIRECTORY : "${{github.workspace}}/build/${{ matrix.build_type }}/shared_${{matrix.shared_type}}"
-      INSTALL_DIRECTORY : "${{github.workspace}}/install/${{ matrix.build_type }}/shared_${{matrix.shared_type}}"
+      BUILD_DIRECTORY : "${{github.workspace}}/build/${{ matrix.build_type }}/shared_${{matrix.shared_type}}/profile_${{matrix.profiling}}"
+      INSTALL_DIRECTORY : "${{github.workspace}}/install/${{ matrix.build_type }}/shared_${{matrix.shared_type}}/profile_${{matrix.profiling}}"
+      RUNNER_ENV : github_runner
       BUILD_CONFIG : >
         -G Ninja
         -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+        -DPARSEC_DEBUG_NOISIER=ON
```
abouteiller, line 33: add -DPARSEC_DEBUG_PARANOID=ON

abouteiller (approved):

abouteiller:
> No other comment, the rest looks good.
> 
> How to add CUDA to that? There is no spack package for CUDA I assume?

Yes there is, it's as simple as adding Cuda to the list. The hard part is testing because we don't have the hardware on the runners.

G-Ragghianti:
> Yes there is, it's as simple as adding Cuda to the list. The hard part is testing because we don't have the hardware on the runners.

For the running that I set up, it has cuda installed under /usr/local/, and you are welcome to install your own cuda with spack.  The runner also should have access to the GPU hardware (V100 on leconte).

G-Ragghianti:
> There is a warning in the test:
> 
> ```
> Run source .github/CI/spack_setup.sh
> fatal: destination path '/home/github_runner/spack' already exists and is not an empty directory.
> ```

The spack_setup script ignores this error message (by doing "git clone || true").  It is just due to git seeing that the directory already exists from a  previous run. This is what you want so that you don't have to install all the spack dependencies from scratch each time.

G-Ragghianti:
> It also seems that HWLOC is not detected even though HWLOC is installed through spack:
> 
> ```
> Could NOT find HWLOC; Options depending on HWLOC will be disabled (missing: HWLOC_LIBRARY HWLOC_INCLUDE_DIR) 
> ```

If you are loading a library dependency from spack but are not building your actual code within spack, then you will have to do whatever is necessary to get your configure/build process to detect the library.  It will only magically find the library if you are using the spack-provided compiler wrappers from within spack.

bosilca:
> The spack_setup script ignores this error message (by doing "git clone || true"). It is just due to git seeing that the directory already exists from a previous run. This is what you want so that you don't have to install all the spack dependencies from scratch each time.

This is only partially true because if the directory exists and is not empty (which include a partial pull from a prior operation) the status of the clone repository is undetermined. It is not clear if this is what's going on here, but bootstraping the spack from that directory fails.

G-Ragghianti:
> This is only partially true because if the directory exists and is not empty (which include a partial pull from a prior operation) the status of the clone repository is undetermined. It is not clear if this is what's going on here, but bootstraping the spack from that directory fails.

Theoretically, this could happen on the very first time a spack job is executed on the github runner container, but I've not seen that happen.  I'd like to see some log output so I can tell if that's the case.  You can also delete the spack install and let spack_setup reclone it again the next time it runs, but I don't have any reason to think that is necessary in this case.

G-Ragghianti:
I noticed that your spack_setup.sh differs significantly from the one I was using.  I tested it and found that the error that is causing it to exit prematurely is on line 29 with `spack env list | grep ${RUNNER_ENV}`.  When $RUNNER_ENV is not found, "grep" will exit with non-zero status and cause the build to short-circuit. The $RUNNER_ENV is not found because it is created later in your script.

.github/CI/spack_setup.sh
```diff
@@ -0,0 +1,56 @@
...
+  echo "git clone spack"
+  git clone https://github.com/spack/spack $DIR || true
+  cd $DIR && git status
+fi
+
+echo "Load spack environment"
+source $DIR/share/spack/setup-env.sh
+
+# Show the known envs
+spack env list | grep ${RUNNER_ENV}
```
G-Ragghianti, line 29: Error is here because grep doesn't find the environment string (hasn't been created yet).

G-Ragghianti:
If you want to reproduce the test, I just ran a small wrapper like this:

``
#!/bin/bash -e
export RUNNER_ENV=github_runner
source spack_setup.sh
``

.github/CI/spack_setup.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ $? -ne 0 ]; then
```
G-Ragghianti, line 30: This also doesn't work because the github runner process runs within a shell that will exit automatically on any non-zero return code (like "bash -e").

.github/CI/spack_setup.sh
bosilca: I'm not sure I agree with this. The script executes in the environment generated by the yaml file, and the RUNNER_ENV environment variable is setup in that file.

G-Ragghianti: Yes, but if you do "command | grep asdf" and "asdf" is not found, grep will return non-zero.  This causes the shell to exit in this case.

```EXIT STATUS
       Normally,  the  exit  status is 0 if selected lines are found and 1 otherwise.  But the exit status is 2 if an
       error occurred, unless the -q or --quiet or --silent option is used and  a  selected  line  is  found.   Note,
       however,  that  POSIX only mandates, for programs such as grep, cmp, and diff, that the exit status in case of
       error be greater than 1; it is therefore advisable, for the sake of portability, to use logic that  tests  for
       this general condition instead of strict equality with 2.
```

.github/CI/spack_setup.sh
bosilca: got it, I need to test the return directly with the command.

bosilca:
This is now ready for primetime. Long story short, having a named environment is not the best approach when you expect to update some of the base packages. Thus, removing the environment each time, guarantee that it is concretized only with the packages we expect.

PR #340, opened 2022-03-10
Clean strings usages in CMake.

bosilca:
When we compare a CMake variable to a string, there is no need to pass
it as a string (aka "${}"), we can just use it directly as a named
variable.


cmake_modules/ParsecCompilerFlags.cmake
```diff
@@ -187,9 +186,9 @@ endif()
 if(CMAKE_Fortran_COMPILER_WORKS)
   get_filename_component (Fortran_COMPILER_NAME ${CMAKE_Fortran_COMPILER} NAME)
   #  message(STATUS "Fortran Compiler ${Fortran_COMPILER_NAME} id is ${CMAKE_Fortran_COMPILER_ID}")
-  if(${CMAKE_Fortran_COMPILER_ID} STREQUAL "GNU")
+  if(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
     # gfortran or g77
-    if(${Fortran_COMPILER_NAME} MATCHES g77)
+    if("${Fortran_COMPILER_NAME}" MATCHES g77)
```
devreal, line 191: Looks to me as if `MATCHES` works with variable names too: https://cmake.org/cmake/help/latest/command/if.html#variable-expansion

parsec/parsec.c
```diff
@@ -701,11 +698,17 @@ parsec_context_t* parsec_init( int nb_cores, int* pargc, char** pargv[] )
 #if defined(PARSEC_PROF_TRACE)
     if( (0 != strncasecmp(parsec_enable_profiling, "<none>", 6)) && (0 == parsec_profiling_init( profiling_id )) ) {
         int i, l;
-        char *cmdline_info = basename(parsec_app_name);
+        char *cmdline_info = NULL;
 
         /* Use either the app name (argv[0]) or the user provided filename */
         if( 0 == strncmp(parsec_enable_profiling, "<app>", 5) ) {
+            /* Specialize the profiling filename to avoid collision with other instances */
```
devreal, line 705: This seems unrelated and seems to supersede https://github.com/ICLDisco/parsec/pull/339/files. On purpose?

abouteiller (commented):
There is also an instance in FindCython that got missed

parsec/parsec.c
bosilca: gosh, I had a dirty tree so I mixes this PR with the patch I made for profiling issue.

devreal (approved):

parsec/fortran/CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
 include(FortranCInterface)
 FortranCInterface_HEADER(f2c_mangle.h
                              MACRO_NAMESPACE "PARSEC_F2C_"
-                             SYMBOL_NAMESPACE "parsec_f2c_"
-                             SYMBOLS my_sub parsecf:my_sub)
```
devreal, line 5: Why was this line removed?

bosilca: Because it was incorrect, we were not providing the my_sub symbol.

devreal (approved):

abouteiller (approved):

PR #344, opened 2022-03-12
Reshape test: variable used for polling should be volatile

devreal:


bosilca (approved):

abouteiller:
Don't think that fix is entirely correct https://www.kernel.org/doc/Documentation/process/volatile-considered-harmful.rst

Should have the memory yield instruction in the loop instead.

@ICLDisco/staff

devreal:
I agree that volatile is not the perfect choice here, I am not worried about performance in this case though. And we do not have something like `cpu_relax()` in PaRSEC, nor can we use `_Atomic` as a replacement for `volatile`. PaRSEC uses `volatile` in many places where `_Atomic` would be a better choice (lifo, list, taskpool, arena, barrier, rwlock, even in the C code generated from JDF). If we  want to go down the route of adding `cpu_relax()` I'm all for it, but it likely requires architecture-specific implementations. We can peek at the kernel code for that, of course. Or just bite the bullet and transition to requiring C11 atomics and allow for `_Atomic` to be used directly.

PR #347, opened 2022-03-14
Apply -g3 to DEBUG only, set default config to Release

abouteiller:
When we changed the way compiler flags are applied, we changed what config receive what flags. This is a proposal PR to restore some (but not all) of the flag scoping we had before:

New:
- default config set to release

Revert: 
- apply -g3 to DEBUG config only

Keep:
- -Wall -Wextra applies to all configs

### DIFF
--- a/cmake_modules/ParsecCompilerFlags.cmake
+++ b/cmake_modules/ParsecCompilerFlags.cmake
@@ -128,15 +128,15 @@ endif(NOT CMAKE_SYSTEM_NAME MATCHES "Windows")
 #
 
 # add gdb symbols in debug and relwithdebinfo, g3 for macro support when available
-check_and_set_compiler_option(OPTION "-g3" NAME PARSEC_HAVE_G3 COMMENT "List of compile options used to enable highest level of debugging (e.g. -g3)")
+check_and_set_compiler_option(OPTION "-g3" NAME PARSEC_HAVE_G3 COMMENT "List of compile options used to enable highest level of debugging (e.g. -g3)" CONFIG DEBUG)
 
 # Some compilers produce better debugging outputs with Og vs O0
 # but this should only be used in RelWithDebInfo mode.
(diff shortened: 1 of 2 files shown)


cmake_modules/ParsecCompilerFlags.cmake
```diff
@@ -128,15 +128,15 @@ endif(NOT CMAKE_SYSTEM_NAME MATCHES "Windows")
 #
 
 # add gdb symbols in debug and relwithdebinfo, g3 for macro support when available
-check_and_set_compiler_option(OPTION "-g3" NAME PARSEC_HAVE_G3 COMMENT "List of compile options used to enable highest level of debugging (e.g. -g3)")
+check_and_set_compiler_option(OPTION "-g3" NAME PARSEC_HAVE_G3 COMMENT "List of compile options used to enable highest level of debugging (e.g. -g3)" CONFIG DEBUG)
```
devreal, line 131: Why only enable `-g3` for debug builds? I don't think this is harmful for RelWithDebInfo builds

abouteiller: I knew somebody would ask :) the real reason is that the `check_and_set_compiler_option` doesn't take a list of CONFIG in the parameter, but a single one, and didn't felt like doing more cmake today.

devreal: OK, I can live with that. This explains why I had pretty decent debug symbols in my `Release` builds. We might want to keep that after all to make debugging even in a release build.

abouteiller: We still get normal -g in RelWithDebInfo, and we definitely should not get -g in Release (that is the difference between RelWithDebInfo and Release, after all)

devreal: Yes and no. RelWithDebInfo disables most optimizations. So bugs appear in Release mode and not in RelWithDebInfo.

bosilca (approved):

PR #351, opened 2022-03-17
Handle error in CUDA/HIP module init and configurable max_streams

therault:
A misconfiguration on leconte makes none of the device capable of allocating a cuda stream.

This leads to a SEGFAULT, because we try to free things that have not been allocated.

With this patch, the devices are just removed (with a warning), but the execution proceeds.

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_component.c
+++ b/parsec/mca/device/cuda/device_cuda_component.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2020 The University of Tennessee and The University
+ * Copyright (c) 2010-2022 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  */
@@ -35,7 +35,7 @@ static int device_cuda_component_register(void);
 
 /* mca params */
 int parsec_device_cuda_enabled_index, parsec_device_cuda_enabled;
-int parsec_cuda_sort_pending = 0;
+int parsec_cuda_sort_pending = 0, parsec_cuda_max_streams = PARSEC_GPU_MAX_STREAMS;
 int parsec_cuda_memory_block_size, parsec_cuda_memory_percentage, parsec_cuda_memory_number_of_blocks;
 char* parsec_cuda_lib_path = NULL;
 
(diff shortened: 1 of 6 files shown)


abouteiller (approved):

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -363,25 +363,29 @@ parsec_cuda_module_init( int dev_id, parsec_device_module_t** module )
...
     if(-1 == len)
-        gpu_device->super.name = "";
+        gpu_device->super.name = strdup("");
     gpu_device->data_avail_epoch = 0;
 
-    gpu_device->max_exec_streams = PARSEC_MAX_STREAMS;
+    gpu_device->max_exec_streams = 0; // We will increment this as streams are succesfully initialized
     gpu_device->exec_stream =
-        (parsec_gpu_exec_stream_t**)malloc(gpu_device->max_exec_streams * sizeof(parsec_gpu_exec_stream_t*));
+        (parsec_gpu_exec_stream_t**)malloc(PARSEC_MAX_STREAMS * sizeof(parsec_gpu_exec_stream_t*));
```
bosilca, line 371: Why are we removing the capability to dynamically define the number of execution streams per device ?

bosilca: This was part of the infrastructure to allow the number of exec_streams to become an MCA param. I would prefer to complete the work than remove it.

abouteiller: I updated the PR to add configurable max_streams

abouteiller (approved):

parsec/mca/device/cuda/device_cuda_component.c
```diff
@@ -189,9 +195,6 @@ static int device_cuda_component_register(void)
     (void)parsec_mca_param_reg_int_name("device_cuda", "max_number_of_ejected_data",
                                         "Sets up the maximum number of blocks that can be ejected from GPU memory",
                                         false, false, MAX_PARAM_COUNT, &parsec_gpu_d2h_max_flows);
-    (void)parsec_mca_param_reg_int_name("device_cuda", "sort_pending_tasks",
```
bosilca, line 192: Move back to original order.

bosilca (approved):

PR #356, opened 2022-03-22
Fix atomics on macosX not working properly (missing header)

abouteiller:
* Fix atomics on macosX not working properly (missing header): we now fallback to using gcc-builtins if we don't have c11; all modern OSX compilers have them, so we removed the macosx specific atomics altogether. 

* Removes the PARSEC_OSX parsec_config.h define and use `__APPLE__` instead

* Change bind thread logic so that we use the OSX specific thread binding by default (even when we have hwloc, because hwloc binding doesn't work on OSX), issue some warning messages if user tries to change default behavior, as the backend doesn't have these capabilities. 

* Also, the conditional for using 128bit atomics must be on the CAS128 capability, not the availability of int128_t (many platforms have the int128, but not the CAS128). Having a single downstream variable rather than an upstream conditional avoids making mistakes and configurations that are non-workable. 

(this is not a fix for the error with c11 atomics on gcc-11.2 osx, I just stumbled upon this problem when forcing other atomic modes for testing)


parsec/include/parsec/sys/atomic-macosx.h
```diff
@@ -44,17 +39,6 @@ int parsec_atomic_cas_int64( volatile int64_t* location,
     return OSAtomicCompareAndSwap64( old_value, new_value, location );
 }
 
-#if defined(PARSEC_ATOMIC_USE_GCC_128_BUILTINS)
```
bosilca, line 47: clang/llvm/gcc/icc on OSX all provide __sync_bool_compare_and_swap, and this intrinsic can be used in addition to the default OSX atomic support. Instead of removing it we should make sure it is in use when available.

CMakeLists.txt
```diff
@@ -296,18 +296,7 @@ include (ParsecCompilerFlags)
 include (cmake_modules/CheckAtomicIntrinsic.cmake)
 if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
   set( PARSEC_OSX 1 CACHE INTERNAL "Compile on MAC OS X")
```
bosilca, line 298: I don't think we need PARSEC_OSX anymore, as far as I see all remaining uses can rely on the default \_\_APPLE__ instead.

parsec/include/parsec/sys/atomic.h
```diff
@@ -38,13 +38,15 @@ BEGIN_C_DECLS
...
  */
+#    if defined(PARSEC_OSX)
+#       include <AvailabilityMacros.h>
+#    endif
 #    if defined(__FUJITSU)
 #      undef PARSEC_ATOMIC_USE_XLC_32_BUILTINS
 #    endif
 #    if defined(PARSEC_ATOMIC_USE_XLC_32_BUILTINS)
 #      include "atomic-xlc.h"
-#    elif defined(PARSEC_OSX)
```
bosilca, line 46: unnecessary changes.

bosilca (approved):

parsec/include/parsec/sys/atomic-macosx.h
abouteiller: I hesitated, but we didn't do this before, on Apple we would force disable the int128 type altogether, and this code would never actually be used. 

If we are using the GCC builtins for the 128 bit, why not use them all around and just drop the OSX  specific OSAtomics? (that is, drop down to the atomic-gcc.h). Is there a convincing performance advantage for maintaining this code?

parsec/include/parsec/sys/atomic.h
abouteiller: Not: if we found OSX, but not the right OSX version, we would define no atomics at all (because we would not consider the other #elif cases for the outer conditional when the inner conditional fails).

parsec/include/parsec/sys/atomic-macosx.h
bosilca: That makes more sense. When atomic support was nascent we needed to deal with what was available, but today, as atomic support become mainstream, we don't necessarily need all the special cases.

Let's remove the OSX specific atomic support.

bosilca (approved):

parsec/bindthread.c
```diff
@@ -19,7 +19,8 @@
 #elif defined(PARSEC_HAVE_SCHED_SETAFFINITY)
 #  include <linux/unistd.h>
 #  include <sched.h>
-#elif defined(PARSEC_OSX)
+#endif  /* define(PARSEC_HAVE_HWLOC) */
+#if defined(__APPLE__)
```
devreal, line 23: What happens if hwloc starts supporting OS X? We'll have both the OS X-specific includes / code paths below and the hwloc code. In other words, why replace the `elif` with `endif`, `if`?

abouteiller: We can compile with hwloc today (that's actually how I run it), all features are supported, that is we can work on cpusets etc to detect problematic binding conditions, but the bind operation itself returns an error and doesn't do anything. 

That's why we have a specific code path to use the OS specific binding operation (it's not quite a real bind, but an approximation of it that will mark a preference for the target cpu, and an aversion for migration, but that's the best that can be done on OSX).

PR #360, opened 2022-03-25
Remove never compiled in '64bit' lifo implementation

abouteiller:
This was never compiled in, and some rumors say it is not working anyways.

### DIFF
--- a/parsec/class/lifo.h
+++ b/parsec/class/lifo.h
@@ -448,125 +448,8 @@ LIFO_STATIC_INLINE parsec_list_item_t* parsec_lifo_try_pop( parsec_lifo_t* lifo 
     return item;
 }
 
-#elif defined(PARSEC_USE_64BIT_LOCKFREE_LIST)
 
-/* Add one element to the LIFO. We will return the last head of the list
- * to allow the upper level to detect if this element is the first one in the
- * list (if the list was empty before this operation).
- */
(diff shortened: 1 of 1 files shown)


devreal (approved):

abouteiller:
The discussion I remember is that it was not completely immune to ABA when the counter rolls over. But honestly I do not remember more than that.

devreal:
Here is the PR from back then: https://bitbucket.org/icldistcomp/parsec/pull-requests/264

I vaguely remember debugging the old 64bit implementation and seeing it fail with ICC and GCC. My memory is extreme fuzzy on that though (I don't see how this code can be correct on 64bit machines, but maybe I have to read it more carefully). Anyway, it's legacy and not used on 99% of machines. Plus we have a fallback on a lock-based implementation that is at least correct.

bosilca:
That discussion mostly talks about the performance of the new implementation, but it does not justify the need for it.

I'm setting up a new rpi 32 bits to test this code, I'll get back to you in a few days.

bosilca:
We are certainly not immune to bugs, but what are the chances that a piece of code suddenly stops working after a decade of good and loyal services?

Anyway, with the current prevalence of C11-atomis (even on the rpi) we are completely covered on our current needs for efficient LIFO implementation, we can remove the old LIFO support.

bosilca (approved):

PR #363, opened 2022-03-29
Fix/many small updates

bosilca:
Cleaning my repos by pushing all fixes that were waiting for a better opportunity to get fixed.


devreal (approved):

therault (approved):
Looks good to me.

NB: checks do not pass, because of two things
  - bug introduced in github actions during the bison changes
  - commits are not signed?

bosilca:
I don't understand why these commits are not signed. I'll look into it, and only merge once this is fixed.

For the github action topic, I'm working on it. The docker will soon be updated to use a recent version of git, and I also added a protection around the action checker for older versions of git (in #333).

PR #372, opened 2022-04-09
[RFC] profiling tools: more efficient cross-stream event matching

omor1:
Performance when matching events that start in one profiling stream and end in a different one is abysmal: O(_m_ \* _n_) for _m_ such events and _n_ total events in the worst case. This has largely not been a problem, since the only time this occurs is when using GPUs, and there are relatively few such events (i.e. _m_ is fairly small).

In trying to profile my ongoing work with LCI, this is no longer the case. Communication events almost always start in the progress thread and end in the communication thread, or vice-versa. This results in many such events, so the current behavior is infeasible.

This code relies on an observation by @bosilca that most start events are immediately followed by their end events. This means it is possible to construct a list of end events that are _possible_ matches for cross-stream events. If the matching event for a start event is not found in its own stream, this list can be searched to determine where to start the search in other streams. Since the list is naturally sorted by timestamp, a binary search can be used—I rely on `bsearch` and a slight hack that stores the last array element in the search key to implement a "fuzzy" timestamp search on a per-event-key basis. This list is constructed lazily when it is needed and only once per stream.

Search performance should be O(_m_ \* log _m_) for _m_ events matched in a different stream. Constructing the lists(s) is O(_n_).

Current code is WIP for early review—do not merge.

### DIFF
--- a/tools/profiling/dbpreader.c
+++ b/tools/profiling/dbpreader.c
@@ -16,6 +16,7 @@
 #include <sys/mman.h>
 #endif
 #include <sys/types.h>
+#include <pthread.h>
 #include <fcntl.h>
 #include <stdarg.h>
 
@@ -214,10 +215,28 @@ int dbp_file_translate_local_dico_to_global(const dbp_file_t *file, int lid)
    (EVENT_HAS_INFO((dbp_event)->native) ?                   \
     (dbp_object)->parent->dico_keys[(dbp_object)->dico_map[BASE_KEY((dbp_event)->native->event.key)]].keylen : 0))
 
+typedef struct {
+    uint64_t timestamp;
+    off_t    offset;
+} event_cache_item_t;
+
+typedef struct {
+    event_cache_item_t *items;
+    size_t              len;
+    size_t              size;
+} event_cache_key_t;
+
+typedef struct {
+    pthread_mutex_t    mtx;
+    event_cache_key_t *keys;
+    int                done;
+} event_cache_t;
+
 struct dbp_thread {
     const parsec_profiling_stream_t *profile;
     dbp_file_t                      *file;
     dbp_info_t                      *infos;
+    event_cache_t                    cache;
     int                              nb_infos;
 };
 
@@ -361,7 +380,7 @@ static const dbp_event_t *dbp_iterator_next_buffer(dbp_event_iterator_t *it)
     return dbp_iterator_current(it);
 }
 
-const dbp_event_t *dbp_iterator_next(dbp_event_iterator_t *it)
+static const dbp_event_t *dbp_iterator_next_in_buffer(dbp_event_iterator_t *it)
 {
     size_t elen;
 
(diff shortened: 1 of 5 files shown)


tools/profiling/python/pbt2ptt.pyx
```diff
@@ -44,7 +44,8 @@ import traceback
 from parsec_trace_tables import * # the pure Python classes
 from common_utils import *
 
-multiprocess_io_cap = 9 # this seems to be a good default on ICL machines
+#multiprocess_io_cap = 9 # this seems to be a good default on ICL machines
+multiprocess_io_cap = 128
```
omor1, line 48: Whoops, this shouldn't have been included—it will be removed in the next iteration I push. I just modified this for better performance on the machines where I do the analysis.

omor1:
I note in the last commit:

> There could be some edge cases where the search finds a potential match in a stream, but the true match is on a different stream; in this case, the current code still traverses the entirety of the mistaken stream. If this becomes an issue, this could be prevented by only comparing against entries in the cache. The cache may need to store some additional metadata (e.g. position/index in event buffer) for this to work; we are currently very fuzzy and simply an iterator to the start of the event buffer containing the potentially-matching event. This is essentially a time/space tradeoff.

I can work on solving this if you think it's necessary @bosilca. This is primarily an issue if unmatched end events with the searched-for key are distributed across many profiling streams—this isn't the case for LCI, so I haven't considered it deeply.

tools/profiling/dbpreader.c
```diff
@@ -378,55 +417,188 @@ void dbp_iterator_delete(dbp_event_iterator_t *it)
...
+    e1 = dbp_iterator_current( i1 );
+
+    /* iterator 2 points to next event */
+    i2 = dbp_iterator_new_from_thread( thr );
+    e2 = dbp_iterator_next( i2 );
+
+    while( NULL != e2 ) {
+        key2 = dbp_event_get_key(e2);
+
+        /* if e2 is end event, but e1 doesn't match, then e2 out of order */
```
therault, line 463: 'out of order' is confusing in the comment here. I would replace with 'then store e2 position in the cache'.

therault (approved):

therault:
@bosilca the CI tests are failing with profiling=ON because spack fails installing gmp-6.2.1...

I don't know why gmp-6.2.1 is required for this setup. Is this a know problem that has been fixed already? Other PRs don't have this issue, and I don't see anything in the PR that adds the dependency to gmp-6.2.1 (and to be honest, I have no idea why we need gmp-6.2.1...)

tools/profiling/dbpreader.c
omor1: That makes sense, I'll do that in a bit once I have access to the machine I'm running on again.

omor1:
> @bosilca the CI tests are failing with profiling=ON because spack fails installing gmp-6.2.1...
> 
> I don't know why gmp-6.2.1 is required for this setup. Is this a know problem that has been fixed already? Other PRs don't have this issue, and I don't see anything in the PR that adds the dependency to gmp-6.2.1 (and to be honest, I have no idea why we need gmp-6.2.1...)

I think it's a prerequisite for GCC. That said, other PRs _are_ failing with this—see #374 for an example.

omor1:
@therault I haven't run with any PaRSEC GPU codes, but @bosilca mentioned when I was working on this that this problem has come up before in that context—GPU computations begin and end on different PaRSEC profiling streams, so you get mismatching events of this nature. If you have a set of profiles, I can see how performance improves for them, or you could test it yourself if you so want.

bosilca:
As far as I see it's the same issue in all cases, a broken spack recipe, that use an older version of libtool (instead of calling libtoolize). I don't have access to the docker capsule, so I can't really fix it. The other alternative is to fix the version of the packages we are installing with spack, to avoid this issue.

bosilca (approved):

abouteiller:
What version do we need, it should be pretty easy to push

PR #374, opened 2022-04-25
flex: on some machines, flex cannot work if parsec/utils is not created

abouteiller:
in the build directory before invocation, so lets create it/

reference: https://gitlab.kitware.com/cmake/cmake/-/issues/19529

### DIFF
--- a/cmake_modules/pregen_flex_bison.cmake
+++ b/cmake_modules/pregen_flex_bison.cmake
@@ -49,6 +49,7 @@ if(archive)
     # Check for problems: stray files in build directory
     execute_process(
       COMMAND git ls-files --error-unmatch ${srcdir}/${source}
+      WORKING_DIRECTORY ${srcdir}
       RESULT_VARIABLE ret
       OUTPUT_VARIABLE stdout
       ERROR_VARIABLE stderr
@@ -61,6 +62,7 @@ if(archive)
     # Check for modifications
     execute_process(
       COMMAND git status --porcelain -- ${srcdir}/${source}
+      WORKING_DIRECTORY ${srcdir}
       RESULT_VARIABLE ret
       OUTPUT_VARIABLE stdout
       ERROR_VARIABLE stderr
--- a/parsec/CMakeLists.txt
+++ b/parsec/CMakeLists.txt
@@ -42,10 +42,12 @@ set(BASE_SOURCES
 
 if(FLEX_FOUND AND BISON_FOUND)
   # generate in the build dir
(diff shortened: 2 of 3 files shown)


abouteiller:
An alternate solution could be to generate the .l.c in `buildir/parsec` instead of `builddir/parsec/utils`, since the `CMakeLists.txt` is in `parsec`, we have a guarantee that it will exist in the builddir.

bosilca:
I don't think this patch is enough, as it missed another directory used for flex/bison files in parsec/interfaces/ptg/ptg-compiler. We need to add the following patch:

```diff
diff --git a/parsec/interfaces/ptg/ptg-compiler/CMakeLists.txt b/parsec/interfaces/ptg/ptg-compiler/CMakeLists.txt
index e693fe98c..76ccf9ec6 100644
--- a/parsec/interfaces/ptg/ptg-compiler/CMakeLists.txt
+++ b/parsec/interfaces/ptg/ptg-compiler/CMakeLists.txt
@@ -8,6 +8,8 @@ IF(NOT CMAKE_CROSSCOMPILING)
     FLEX_TARGET(parsec_flex parsec.l  ${CMAKE_CURRENT_BINARY_DIR}/parsec.l.c)
     ADD_FLEX_BISON_DEPENDENCY(parsec_flex parsec_yacc)
 
+    # generate in the pregen dir
+    file(MAKE_DIRECTORY "${PARSEC_PREGEN_FLEX_BISON_DIR}/parsec/interfaces/ptg/ptg-compiler")
     # Generate in the pregen dir, add to the non-automated rule to update the archive
     BISON_TARGET(pregen_parsec_yacc parsec.y ${PARSEC_PREGEN_FLEX_BISON_DIR}/parsec/interfaces/ptg/ptg-compiler/parsec.y.c COMPILE_FLAGS -l)
     FLEX_TARGET(pregen_parsec_flex parsec.l ${PARSEC_PREGEN_FLEX_BISON_DIR}/parsec/interfaces/ptg/ptg-compiler/parsec.l.c COMPILE_FLAGS --noline)
```

bosilca (changes requested):
I left a comment in the "Conversation" section.

abouteiller:
> I left a comment in the "Conversation" section.

Did your changes

@devreal can you try as you are the only one with the bad flex? especially with the non-automatic rule 
`cmake --build . --target parsec_pregen_flex_bison` (follow the on-screen instructions to not commit the changes)

bosilca (approved):

devreal (approved):
Works for me now, thank you!

PR #375, opened 2022-04-27
Do now allow spack to be updated automatically.

bosilca:


therault (approved):
Maybe add a comment why we need to remove ${HOME}/.spack ?

But it looks ok to me.

PR #376, opened 2022-04-27
profiling: disambiguation between certain MPI events

### LINKED ISSUE #373: Profiling: MPI_DATA_PLD_SND/MPI_DATA_PLD_RCV will mismatch end event
## Describe the bug
The `MPI_DATA_PLD_SND` and `MPI_DATA_PLD_RCV` profiling events use dep index as the event id. There can be many data transfers occurring concurrently, including multiple for the same dep index for different task classes or different instances of the same task class. An end event `E` is considered to match a start event `S` when:
1. `S` and `E` have the same base key i.e. are the same event type
2. `S` and `E` have the same event identifier
3. `S` and `E` have the same taskpool id
4. `S` has a timestamp before that of `E`

In particular, when matching the _next_ matching event in the stream is assumed to be the correct match. Suppose a profiling stream has the following events:
1. Data Send Start, event id = 0 [start send for dep 0 of task A]
2. Data Send Start, event id = 0 [start send for dep 0 of task B]
3. Data Send End, event id = 0 [end send for dep 0 of task A]
4. Data Send End, event id = 0 [end send for dep 0 of task B]

Then both start events will complete with the end profiling event 3, rendering the end completion times bogus.

### To Reproduce
1. Profile communication for some PaRSEC application
2. Convert to PTT/HDF5
3. Analyze:
```python
import parsec_trace_tables as ptt
import pandas as pd
trace = ptt.from_hdf("/path/to/profile.hdf5")
trace.events.loc[lambda df: df.type == trace.event_types.at["MPI_DATA_PLD_SND"]].groupby("node_id").apply(
    lambda df: pd.Series(
        {"end": len(df.end.unique()), "begin": len(df.begin.unique()), "len": len(df)}
    )
)
```

Looking at an existing profile, I see:
| node_id |   end |  begin |    len |
|---------|-------|--------|--------|
| 0       | 44498 |  67122 |  67122 |
| 1       | 88903 | 133728 | 133728 |
| 2       | 22619 |  34049 |  34049 |
| 3       | 22701 |  33599 |  33599 |

Note that around 1/3 of the events share an end timestamp—these are mismatched!

### Expected behavior
Matching should find the correct end event; it must be possible to disambiguate between concurrent events with the same dep index. I believe that other communication profiling events do this correctly: activations are funneled through the comm thread and complete in-order (blocking `MPI_Send` is used), so using a \`0' tag is valid, though not ideal; and the data control event is similarly funneled, though it uses a monotonically incrementing counter to disambiguate events. The communication engine branch I'm using is somewhat better about this, using an atomically incrementing event id for both activations and data control events, but data send/recv is still bugged.

I'd like to keep the dep index around in some form—it's useful for analysis of _which_ data was communicated—but it is not suitable for the event id. It should probably be added to the event info, if possible.

## Environment (please complete the following information):
 - PaRSEC version: recent
 - OS: Rocky Linux 8.5 [exec], macOS 12.3.1 [analysis]
 - Compiler: AMD clang version 12.0.0
 - MPI version: N/A

omor1:
`MPI_DATA_PLD_SND` and `MPI_DATA_PLD_RCV` use the dep index as profiling event ID. Depending on communication ordering, this can result in mismatches with the end event.
Ensure that events use unique IDs: `(peer << 32) | (tag + dep_index)`

Fixes #373.

### DIFF
--- a/parsec/remote_dep.h
+++ b/parsec/remote_dep.h
@@ -372,28 +372,18 @@ void remote_deps_allocation_init(int np, int max_output_deps);
 void remote_deps_allocation_fini(void);
 
 typedef struct {
-    int rank_src;  // 0
-    int rank_dst;  // 4
-    uint64_t tid;  // 8
-    uint32_t tpid;  // 16
-    uint32_t tcid;  // 20-
-    int32_t msg_size; // 24
-    int32_t padding;  // 28 -- this field is not necessary, but the structure will be padded
-                      //       by the compiler due to the uint64_t field. It is declared here
-                      //       just to be consistent with the conversion string.
+    int rank_src;  //  0
+    int rank_dst;  //  4
+    uint64_t tid;  //  8
+    uint32_t tpid; // 16
+    uint32_t tcid; // 20
+    int msg_size;  // 24
+    int dep;       // 28
 } parsec_profile_remote_dep_mpi_info_t; // 32 bytes
 
-
 #ifdef PARSEC_PROF_TRACE
-extern int MPI_Activate_sk, MPI_Activate_ek;
-extern int MPI_Data_ctl_sk, MPI_Data_ctl_ek;
-extern int MPI_Data_plds_sk, MPI_Data_plds_ek;
-extern int MPI_Data_pldr_sk, MPI_Data_pldr_ek;
-extern int activate_cb_trace_sk, activate_cb_trace_ek;
-extern int put_cb_trace_sk, put_cb_trace_ek;
-
-// TODO: how to replace call to MPI_Pack_size?
-#define TAKE_TIME_WITH_INFO(PROF, KEY, I, src, dst, rdw, nbdtt, dtt, comm) \
+#define TAKE_TIME_WITH_INFO(PROF, KEY, I, k, src, dst, rdw, nbdtt, dtt) \
+  do {                                                                  \
     if( parsec_profile_enabled ) {                                      \
         parsec_profile_remote_dep_mpi_info_t __info;                    \
         parsec_taskpool_t *__tp = parsec_taskpool_lookup( (rdw).taskpool_id ); \
@@ -404,15 +394,17 @@ extern int put_cb_trace_sk, put_cb_trace_ek;
         __info.tcid = (rdw).task_class_id;                              \
         __info.tid  = __tc->key_functions->key_hash(                    \
                              __tc->make_key(__tp, (rdw).locals), NULL); \
-        MPI_Pack_size(nbdtt, dtt, comm, &__info.msg_size);              \
+        parsec_ce.pack_size(&parsec_ce, nbdtt, dtt, &__info.msg_size);  \
+        __info.dep = (k);                                               \
         PARSEC_PROFILING_TRACE((PROF), (KEY), (I),                      \
                                PROFILE_OBJECT_ID_NULL, &__info);        \
-    }
+    }                                                                   \
+  } while (0)
 
 #define TAKE_TIME(PROF, KEY, I) PARSEC_PROFILING_TRACE((PROF), (KEY), (I), PROFILE_OBJECT_ID_NULL, NULL)
 
 #else
-#define TAKE_TIME_WITH_INFO(PROF, KEY, I, src, dst, rdw, count, dtt, comm) do {} while(0)
+#define TAKE_TIME_WITH_INFO(PROF, KEY, I, k, src, dst, rdw, nbdtt, dtt) do {} while(0)
 #define TAKE_TIME(PROF, KEY, I) do {} while(0)
 #endif  /* PARSEC_PROF_TRACE */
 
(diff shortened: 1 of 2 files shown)


bosilca:
I think the root of the idea is correct, use the `tag + dep_index` to differentiate the communications. However, I don't think that adding the peer is correct, because this leads to a lack of uniqueness of these identifiers across the entire job (same target process would receive the same tag with potentially the same number of dependencies from different peers). What would always remain unique is the combination of local rank and `tag + dep_index`

omor1:
> However, I don't think that adding the peer is correct, because this leads to a lack of uniqueness of these identifiers across the entire job (same target process would receive the same tag with potentially the same number of dependencies from different peers).

The tag is determined by the sender when it sends the activation message, unless I've misunderstood the code. The combination of sender rank and `tag + dep_index` is unique for `MPI_DATA_PLD_RCV`. For `MPI_DATA_PLD_SND` it is sufficient to use `tag + dep_index`, but it might make sense to use the same event id as for `MPI_DATA_PLD_RCV` to make associating the events easier in analysis; I think I have this slightly backwards as I use the receiver rank on the sender's side. I'll fix that.

omor1:
Actually I've realized that the `MPI_Status` in `remote_dep_mpi_put_end_cb` is from an `MPI_Isend`, so none of the fields are valid—in particular, `status->MPI_TAG` is not valid, which was being used even before this PR.

I'm not sure how otherwise to retrieve the tag at the sender in `remote_dep_mpi_put_end_cb`; it's not valid in `status`, it's not recorded in `cb`, and `cb->deps` doesn't have it either. In `remote_dep_mpi_put_start` the tag is retrieved from `item->cmd.activate.task.tag`, but `item` is freed at the end of the function, so we lose the tag thereafter. I can add a field to `cb` for the tag—this doesn't actually use any more memory on LP64 systems, since `parsec_comm_callback_t` is padded by 4 bytes anyway.

bosilca:
go for that, add the tag into the parsec_comm_callback_t struct.

bosilca (dismissed):

bosilca:
I meant to the `remote_dep_cb_data_t` struct.

omor1:
> I meant to the `remote_dep_cb_data_t` struct.

At the time the PR was started the comm engine revamp hadn't been merged, so `parsec_comm_callback_t` was correct. I haven't rebased and fixed this for the comm engine yet.

therault (approved):
Looks good to me. Only improvement I would suggest: define a macro or a static inline function to compute an event_id from the rank, k and the tag. There are at least 3 places where you compute `(((uint64_t)(es->virtual_process->parsec_context->my_rank) << 32) | (uint64_t)(tag + k));` and an explicit function to do so would simplify changing the formula if that is ever necessary.

Other comment: tag is an overloaded term, between the parsec_ce tag and the MPI tag. The first comment ('tag of the associated MPI communication') could be misinterpreted.

bosilca:
why not simply using the tag provided by the `next_tag` and store that in your own structure ? This will save an atomic, and the tag will be unique for an entire cycle (they will be reused once we run out of them)

omor1:
> why not simply using the tag provided by the `next_tag` and store that in your own structure ? This will save an atomic, and the tag will be unique for an entire cycle (they will be reused once we run out of them)

I explained that—`next_tag` is implemented in the MPI funneled comm engine, `parsec_mpi_funnelled.c`, while the comm events are generated by the comm engine user, `remote_dep_mpi.c`. The high-level has no concept of per-communication tags and doesn't care that the comm engine whether implements `send_am`, `put`, or `get`, with tags, while the comm engine doesn't know anything about what data is being transported.

I also just noticed:
```C
    TAKE_TIME_WITH_INFO(es->es_profile, MPI_Activate_sk,0,
                        es->virtual_process->parsec_context->my_rank,
                        peer, deps->msg, position, MPI_PACKED, MPI_COMM_WORLD);
    parsec_ce.send_am(&parsec_ce, REMOTE_DEP_ACTIVATE_TAG, peer, packed_buffer, position);
    TAKE_TIME(es->es_profile, MPI_Activate_ek, event_id);
```

`event_id` != `0`, these events won't match! Also only end events for `MPI_DATA_CTL` are ever generated... I'll put those fixes in my PR as well.

bosilca:
@omor1 please fix the conflicts and the left over comments, I would like to merge this.

omor1:
> @omor1 please fix the conflicts and the left over comments, I would like to merge this.

Please give me direction on how I should proceed—I explained above why your suggested approach doesn't work.

omor1:
> Also only end events for `MPI_DATA_CTL` are ever generated... I'll put those fixes in my PR as well.

Ditto for `MPI_DATA_PLD_RCV`, things got really screwy at some point. I'll try to get the fixes in today.

omor1:
Oops, foiled by macro replacement. Fix incoming...

bosilca (approved):

omor1:
I'll note that an improvement that would save 4 bytes per activation message would be to use a different structure for those messages. This also still has an issue where packed activation messages only record the info for the first activation in the buffer—in my branch I've been additionally generating a "pack" event for each activation packed, which lets me match up activations.

PR #377, opened 2022-04-29
Perfetto profiling

therault:
Add a simple converter from PaRSEC Binary Profile files into a Perfetto trace file.

See https://ui.perfetto.dev/ to visualize such traces.

### DIFF
--- a/tools/profiling/CMakeLists.txt
+++ b/tools/profiling/CMakeLists.txt
@@ -41,5 +41,42 @@ if(Graphviz_FOUND)
 
 endif(Graphviz_FOUND)
 
-Add_Subdirectory(python)
+if(CMAKE_CXX_COMPILER)
+  include(FetchContent)
+  FetchContent_Declare(perfetto_source
+                       URL https://android.googlesource.com/platform/external/perfetto/+archive/refs/heads/releases/v23.x/sdk.tar.gz
+                       CONFIGURE_COMMAND ""
+                       UPDATE_COMMAND ""
+                       BUILD_COMMAND ""
+                       INSTALL_COMMAND ""
+                       BUILD_BYPRODUCTS perfetto.h perfetto.cc)
+  FetchContent_MakeAvailable(perfetto_source)
+  FetchContent_GetProperties(perfetto_source SOURCE_DIR PERFETTO_SOURCE_DIR)
 
+  if( EXISTS ${PERFETTO_SOURCE_DIR}/perfetto.h AND EXISTS ${PERFETTO_SOURCE_DIR}/perfetto.cc )
+    message(STATUS "Found Perfetto library sources in ${PERFETTO_SOURCE_DIR}")
+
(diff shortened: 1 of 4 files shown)


tools/profiling/dbp2pftrace.cc
```diff
@@ -0,0 +1,327 @@
+//
+// Created by herault on 4/19/22.
```
devreal, line 2: Should there be a proper copyright header?

tools/profiling/CMakeLists.txt
```diff
@@ -41,5 +41,42 @@ if(Graphviz_FOUND)
 
 endif(Graphviz_FOUND)
 
-Add_Subdirectory(python)
+if(CMAKE_CXX_COMPILER)
```
bosilca, line 44: The perfetto content seems to be downloaded in our source directory, and will confuse git and possibly us with new untracked files. Maybe we should have a 3rd-party directory where we keep our external dependencies.

tools/profiling/dbp2pftrace.cc
```diff
@@ -0,0 +1,330 @@
...
+void StopTracing(std::unique_ptr<perfetto::TracingSession> tracing_session, int fd)
+{
+    // Make sure the last event is closed for this example.
+    perfetto::TrackEvent::Flush();
+
+    // Stop tracing and read the trace data.
+    tracing_session->StopBlocking();
+    std::vector<char> trace_data(tracing_session->ReadTraceBlocking());
+
+    // Write the result into a file.
```
bosilca, line 84: Are these leftover comments from their example ?

tools/profiling/dbp2pftrace.cc
```diff
@@ -0,0 +1,330 @@
...
+}
+
+void StopTracing(std::unique_ptr<perfetto::TracingSession> tracing_session, int fd)
+{
+    // Make sure the last event is closed for this example.
+    perfetto::TrackEvent::Flush();
+
+    // Stop tracing and read the trace data.
+    tracing_session->StopBlocking();
+    std::vector<char> trace_data(tracing_session->ReadTraceBlocking());
```
bosilca, line 82: Why is there a read of the events ?

tools/profiling/CMakeLists.txt
therault: I have tested again, FetchContent_Declare / FetchContent_Populate download and expand the tar.gz file into the `${PROJECT_BINARY_DIR}/_deps/perfetto_source-src/` directory. 

The call to `  FetchContent_GetProperties(perfetto_source SOURCE_DIR PERFETTO_SOURCE_DIR)` a few lines after that allow us to find exactly where it is, but when I tested it is always in the build directory

therault:
The Perfetto SDK is a *tracing* interface: there is no way to guarantee, when converting a file into another, that all events will be kept in the new trace. I have updated this branch to a more recent Perfetto while trying to make it work and understanding the behavior of the tracing library, this code might be re-usable if we create a new MCA framework to have alternative tracing mechanisms.

Meanwhile, the trace conversion approach needs to either go one level lower, at the Protobuf level, or to keep generating a JSON file.

PR #379, opened 2022-05-07
Hash table: count used buckets only when needed

devreal:
The used-bucket count is used to remove old tables from the search
path once they are empty. However, keeping the count for the current
head requires an atomic operation for each insert and removal for a
value that is otherwise unused while the table is the current head.
Instead, count the number of used buckets during resize so that we
can remove the table once it is empty.

### DIFF
--- a/parsec/class/parsec_hash_table.c
+++ b/parsec/class/parsec_hash_table.c
@@ -252,15 +252,25 @@ static void parsec_hash_table_resize(parsec_hash_table_t *ht)
 {
     parsec_atomic_lock_t unlocked = PARSEC_ATOMIC_UNLOCKED;
     parsec_hash_table_head_t *head;
-    int nb_bits = ht->rw_hash->nb_bits + 1;
+    parsec_hash_table_head_t *old_head = ht->rw_hash;
+    int nb_bits = old_head->nb_bits + 1;
     assert(nb_bits < 32);
+
+    /* count the number of used buckets */
+    int32_t used_buckets = 0;
+    for (size_t i = 0; i < (1ULL << old_head->nb_bits); ++i) {
+        if (NULL != old_head->buckets[i].first_item) {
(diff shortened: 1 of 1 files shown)


bosilca (approved):

bosilca:
I'm not sure I agree with the description of the current behavior, the atomic update on the number of used buckets is only done when a bucket is first, respectively last, used. So the number of atomic operations is not linear with the number of stored items but with the number of buckets.

parsec/class/parsec_hash_table.c
```diff
@@ -252,15 +252,23 @@ static void parsec_hash_table_resize(parsec_hash_table_t *ht)
 {
     parsec_atomic_lock_t unlocked = PARSEC_ATOMIC_UNLOCKED;
     parsec_hash_table_head_t *head;
-    int nb_bits = ht->rw_hash->nb_bits + 1;
+    parsec_hash_table_head_t *old_head = ht->rw_hash;
+    int nb_bits = old_head->nb_bits + 1;
     assert(nb_bits < 32);
 
+    /* count the number of used buckets */
```
bosilca, line 259: There seems to be an implicit assumption that `used_buckets` is 0. If that's really the case add an assert.

devreal: Good point, I changed the code to count into a local variable and set the `used_bucket` member once. It isn't changed we count the used buckets here and then it's only decremented afterwards.

devreal:
The truth is somewhere in between: the number of atomic operations depends on the fill level of the buckets. If a buckets keeps flipping between empty and having a single element then there is an atomic operation for each insert/remove. If there are collisions, then the atomic operations are less frequent.

devreal:
Rebased onto master with revamp merged.

therault (approved):

PR #387, opened 2022-06-10
Make the PUSHOUT and other DTD GPU concepts generic

abouteiller:
Make the PUSHOUT and other DTD GPU concepts generic so that we can compile w/o cuda (and with rocm, when rocm PR merge)


parsec/interfaces/dtd/insert_function.c
```diff
@@ -2213,9 +2204,20 @@ static parsec_hook_return_t parsec_dtd_cuda_task_submit(parsec_execution_stream_
     }
     parsec_device_load[dev_index] += (float)gpu_task->load;
 
-    return parsec_cuda_kernel_scheduler(es, gpu_task, dev_index);
-}
+    parsec_device_module_t *device = parsec_mca_device_get(dev_index);
+    assert(NULL != device);
+    switch(device->type) {
```
abouteiller, line 2209: switch form here because we will have to add the rocm case

bosilca (approved):

PR #389, opened 2022-06-15
Add a load-balancing skew so that we favor locality up to a configurable limit

abouteiller:
Add a load-balancing skew so that we favor locality up to a configurable limit, where we then favor load balancing

### DIFF
--- a/parsec/mca/device/device.c
+++ b/parsec/mca/device/device.c
@@ -58,6 +58,14 @@ float *parsec_device_dweight = NULL;
 float *parsec_device_tweight = NULL;
 
 /**
+ * Load balance skew we are willing to accept to favor RO data reuse
+ * on GPU: a value of 20% means that we will schedule tasks on the preferred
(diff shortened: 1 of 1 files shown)


parsec/mca/device/device.c
```diff
@@ -57,6 +57,14 @@ float *parsec_device_sweight = NULL;
 float *parsec_device_dweight = NULL;
 float *parsec_device_tweight = NULL;
 
+/**
+ * Load balance skew we are willing to accept to favor RO data reuse
+ * on GPU: a value of 20% means that we will schedule tasks on the preferred
+ * GPU except if it is loaded 1.2 times as much as the best load balance option
+ */
+static int parsec_device_load_balance_skew = 20;
+static float load_balance_skew = 1.2f; // don't forget to change me if you change the default above!
```
bosilca, line 66: can we always compute the floating point skew and don't have a default value ?

bosilca (approved):

PR #390, opened 2022-06-16
profiling tools: improve large buffer performance

omor1:
Cross-stream-matching caches only stored a single entry per buffer,
iterating over each event in the buffer to find the matching event.
This is impractical for large buffers such as 2 MiB (x86 huge pages).
This commit adds metadata to each cache entry for event position and
index so that we store each possibly-matching event in the cache, jump
directly to the first possibly-matching event, and iterate only over
those events.

This commit also has minor fixes and improvements for streams without
any matching events and similar scenarios.

### DIFF
--- a/tools/profiling/dbpreader.c
+++ b/tools/profiling/dbpreader.c
@@ -218,6 +218,7 @@ int dbp_file_translate_local_dico_to_global(const dbp_file_t *file, int lid)
 typedef struct {
     uint64_t timestamp;
     off_t    offset;
+    int64_t  event_idx;
 } event_cache_item_t;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


omor1:
Don't merge yet—I want to double-check this works correctly!

tools/profiling/dbpreader.c
```diff
@@ -327,18 +329,24 @@ const dbp_event_t *dbp_iterator_current(dbp_event_iterator_t *it)
     return &it->current_event;
 }
 
-static const dbp_event_t *dbp_iterator_set_offset(dbp_event_iterator_t *it, off_t offset)
+static const dbp_event_t *dbp_iterator_set_offset(dbp_event_iterator_t *it,
```
bosilca, line 332: why do we need 3 variables to point us to a position in a stream ?

omor1: 1) which buffer
2) position inside buffer
3) index of this position (this isn't really needed, but can't be calculated)

bosilca: I was referring to `offset`, `event_pos` and `event_idx`. How do I come up with a tuple `offset, event_pos, event_idx` that makes sense, and how is the function reacting when I provide incorrect values ?

omor1: I don't think I can (easily) sanity-check `offset`, unless the offset of the last valid buffer in the file is stored somewhere in the header.

omor1: I have no way of checking whether the `event_pos` and `event_idx` agree with _each-other_, but that's the entire reason I need to cache both.

bosilca: ok, so in addition to moving to a new profiling page (using offset), you also move to a specific event (event_pos in bytes and event_idx in logical position). In fact, one could have used a single offset (using modulo to compute the relative displacement from the beginning of the current page).

Moreover, a short description would have saved time to read the code. So, this is all perfectly fine, for as long as the function does some checks and it is not called `dbp_iterator_set_offset`, but uses a more clear name, something that indicates that it moves to a specific event.

omor1: > In fact, one could have used a single offset (using modulo to compute the relative displacement from the beginning of the current page).

I think I'm a bit unclear on what you mean by this—events can have different sizes due to info, so you can't easily associate event position and index without iterating over the prior events in the buffer.

bosilca: There are two offsets in use here: `offset` which is relative to the file and increases in increments of profiling page size, and `event_pos` which is relative to the starting of the current page and which is increment using the real size of the events that are read. These two offset are complementary and because the page size is of fixed size, and the other is always smaller than the page size, can be combined together.

omor1: Oh I see what you mean, clever! I was thinking that since `event_pos` is relative not to the start of the page but to the event buffer (i.e. after the header) that it could cause issues, but you're right, it wouldn't. I'll implement that.

therault (approved):

bosilca (approved):

PR #392, opened 2022-06-17
Update to a newer spack compiler

bosilca:
Correctly set the environment variables before cmake invocation

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -60,7 +60,8 @@ jobs:
       # The CMake binaries on the Github Actions machines are (as of this writing) 3.12
       run: |
         source ${{github.workspace}}/.github/CI/spack_setup.sh
-        cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE $BUILD_CONFIG CC=gcc CXX=g++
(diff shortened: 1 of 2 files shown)


therault (approved):

abouteiller (approved):

.github/workflows/build_cmake.yml
```diff
@@ -60,7 +60,8 @@ jobs:
       # The CMake binaries on the Github Actions machines are (as of this writing) 3.12
       run: |
         source ${{github.workspace}}/.github/CI/spack_setup.sh
-        cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE $BUILD_CONFIG CC=gcc CXX=g++
+        echo CC=gcc CXX=g++ FC=gfortran cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE $BUILD_CONFIG
+        CC=gcc CXX=g++ FC=gfortran cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE $BUILD_CONFIG
```
devreal, line 64: Maybe a nitpick, but isn't the canonical way of specifying the compiler in cmake to use `-DCMAKE_C_COMPILER=gcc` instead of `CC=gcc`?

bosilca: The environment variable takes precedence if the -DCMAKE_C_COMPILER is not specified. For all practical purposes, in our case the two are equivalent.

PR #399, opened 2022-07-22
Execution stream keeps the highest priority task for local execution.

bosilca:
Instead of pushing the task into the scheduler and then asking for
something to execute, an execution stream will keep the highest priority
tasks as the next ready task. In general this will improve the data
reuse, but it is possible that it will prevent the scheduler to follow
the critical path (because the highest priority task might be less
important that tasks currently available in the scheduler).

### DIFF
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2019 The University of Tennessee and The University
+ * Copyright (c) 2009-2022 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  */
@@ -366,7 +366,7 @@ int __parsec_schedule_vp(parsec_execution_stream_t* es,
     assert( (NULL == es) || (parsec_my_execution_stream() == es) );
 #endif  /* defined(PARSEC_DEBUG_PARANOID) */
 
-    if( NULL == es ) {
+    if( NULL == es || !parsec_runtime_keep_highest_priority_task ) {
         for(int vp = 0; vp < es->virtual_process->parsec_context->nb_vp; vp++ ) {
             parsec_task_t* ring = task_rings[vp];
             if( NULL == ring ) continue;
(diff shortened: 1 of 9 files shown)


parsec/scheduling.c
```diff
@@ -484,7 +533,7 @@ int __parsec_context_wait( parsec_execution_stream_t* es )
     uint64_t misses_in_a_row;
     parsec_context_t* parsec_context = es->virtual_process->parsec_context;
     int32_t my_barrier_counter = parsec_context->__parsec_internal_finalization_counter;
-    parsec_task_t* task;
+    parsec_task_t* task, *last_local_task = NULL /* for debug purposes */;
```
devreal, line 536: `last_local_task` seems unused

parsec/scheduling.c
```diff
@@ -322,14 +327,58 @@ int __parsec_schedule(parsec_execution_stream_t* es,
...
+    int ret = 0;
+
+    for(int vp = 0; vp < es->virtual_process->parsec_context->nb_vp; vp++ ) {
+        parsec_task_t* ring = task_rings[vp];
+        if( NULL == ring ) continue;
+
+        target_es = vps[vp]->execution_streams[0];
+
+        if( vp == es->virtual_process->vp_id ) {
+            if( NULL == es->next_task ) {
```
devreal, line 363: Can this function ever be called from the comm thread? If so, we might lose tasks because the comm thread uses the same `es` as thread 0

parsec/scheduling.c
bosilca: it is temporary used [aka in draft mode] for debug purposes.

parsec/scheduling.c
bosilca: the comm thread, in fact no threads, should not call this function with that `es`. It can be called with NULL, or with its own copy where the `next_task` field is set to a predefined value.

abouteiller:
Looks like with this change it becomes impossible to create a deterministic scheduler because the decision of keeping the first task to oneself is taken before the scheduler gets invoked?

bosilca:
Correct. The balance here is between the possibility of one day creating a deterministic scheduler (thing we never did because of the performance implication), and allowing the data to bounce if we don't retain the highest priority task locally.

therault (commented):
Could we have an MCA parameter to disable storing in the next_task, in case we want to force everything to go in the scheduler?

bosilca:
runtime_keep_highest_priority_task is now available to control the fate of the highest priority task.

abouteiller (approved):

devreal (approved):

PR #402, opened 2022-07-29
profiling tools: fix format for padded structures

omor1:
Fix for padded structures in the profiling tools


omor1:
@bosilca I think CI is broken again

bosilca (approved):

PR #409, opened 2022-08-08
Allow the DSL to provide a task_snprintf function and use it when displaying the DOT

devreal:
Backport from the TTG branches...

Signed-off-by: Thomas Herault <herault@icl.utk.edu>

### DIFF
--- a/parsec/parsec_prof_grapher.c
+++ b/parsec/parsec_prof_grapher.c
@@ -28,6 +28,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <inttypes.h>
+#include <ctype.h>
 
 #if defined(PARSEC_PROF_GRAPHER)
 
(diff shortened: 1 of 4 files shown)


therault (approved):

parsec/parsec_prof_grapher.c
```diff
@@ -125,7 +130,11 @@ void parsec_prof_grapher_task(const parsec_task_t *context,
     if( NULL != grapher_file ) {
         char tmp[MAX_TASK_STRLEN], nmp[MAX_TASK_STRLEN];
         char sim_date[64];
-        parsec_task_snprintf(tmp, MAX_TASK_STRLEN, context);
+        if(NULL != context->task_class->task_snprintf) {
```
bosilca, line 133: Let's make it the default to use the task class accessor.

devreal: You mean we should provide `task_snprintf` callbacks for DTD and PTG by default?

bosilca: yes, always use `task_class->task_snprintf`.

devreal: So there are two ways of going with this. We're using `parsec_task_snprintf` in many places and all of them would have to be adjusted:

1) Use `parsec_task_snprintf` as the default implementation in `task_class`. That would require replacing all calls to `parsec_task_snprintf` with `task->task_class->task_snprintf`. 
2) Inside `parsec_task_snprintf`, check if `task_class->task_snprintf` is set and if not fallback to the current implementation. This would not require us to touch all the places that call `parsec_task_snprintf`.

I'm leaning to (1) although it will make this PR significantly bigger...

bosilca: (1) is perfectly fine.

therault: (1) should be implemented now.

NB: debugging still falls back on parsec_task_snprintf(). We may want to change this to always call task_class->task_snprintf (or have a more robust version that checks if it exists for debugging purpose, as Joseph suggested). For now, only parsec_prof_grapher uses task_class->task_snprintf. I think changing everything could be another PR, but if you think we should also do that change now, speak up before merging.

abouteiller (approved):

PR #410, opened 2022-08-08
Log all types of runtime-system events in task_profiler, and ensure that all time of compute threads is accounted for

devreal:
From the TTG branches...


therault (approved):

parsec/scheduling.c
```diff
@@ -411,6 +414,8 @@ int __parsec_complete_execution( parsec_execution_stream_t *es,
             parsec_taskpool_update_runtime_nbtask(tp, -1);
     }
 
+    PARSEC_PINS(es, COMPLETE_EXEC_END, task);
```
bosilca, line 417: This event was meant to compute the user side completion of a task, excluding anything related to runtime.

parsec/scheduling.c
```diff
@@ -559,7 +560,9 @@ int __parsec_context_wait( parsec_execution_stream_t* es )
         if( task != NULL ) {
             misses_in_a_row = 0;  /* reset the misses counter */
 
+            PARSEC_PINS(es, SELECT_END, task);
```
bosilca, line 563: not clear why we need to move these calls from inside to outside of this function.

bosilca (commented):
I don't understand the goal of this PR. It seems to me that it replicates a lot of events we already have, but without offering the capability to turn them on and off (as we do now).

omor1:
> It seems to me that it replicates a lot of events we already have, but without offering the capability to turn them on and off (as we do now).

For some of these I haven't figured out a way to turn on; to get the `SELECT ` event I had to create my own PINS module. On which note, is it possible to either add an MCA param to this module or separate the different events to different modules so that we can pick and choose which events to profile, instead of getting all or nothing?

devreal:
@therault I rebased the PR to resolve some conflictes. Please check that I didn't mess up anything.

bosilca:
> For some of these I haven't figured out a way to turn on; to get the `SELECT ` event I had to create my own PINS module. On which note, is it possible to either add an MCA param to this module or separate the different events to different modules so that we can pick and choose which events to profile, instead of getting all or nothing?

We came up with a solution and @devreal implemented it here.

parsec/mca/pins/task_profiler/pins_task_profiler_module.c
```diff
@@ -97,6 +174,11 @@ static void pins_fini_task_profiler(parsec_context_t *master_context)
 static void pins_thread_init_task_profiler(struct parsec_execution_stream_s * es)
 {
     parsec_pins_next_callback_t* event_cb;
+    event_cb = (parsec_pins_next_callback_t*)malloc(sizeof(parsec_pins_next_callback_t));
```
bosilca, line 177: What is the point of registering events that will not be generated ?

omor1: Are you referring to the PREPARE_INPUT events?

bosilca: We now have a mask of events that will be generated. If we move the reading the of MCA earlier in this function, we can only register the events that are useful.

omor1: Ah, I see—so rather than registering callbacks for all events, but those events never being generated due to the `parsec_pins_enable_mask`, simply don't register callbacks for those events.

This also lets a different PINS module listen to those events if it wants. The current solution has the MCA param "belong" to the task_profiler module, but it changes what PINS events are generated for _all_ PINS modules, not just itself.

bosilca (approved):

PR #391, opened 2022-06-17
TTG/termdet

therault:
This is the first PR of the TTG merge: it includes all changes related to the Termination Detection.

This PR also superseeds the PR 429 of the old bitbucket repository (https://bitbucket.org/icldistcomp/parsec/pull-requests/429)

All tests are currently passing, but only one, project_dyn in the haar-tree directory uses termination detection.

Termination Detection is a new MCA, with 3 modules: 4counter using the 4 counters algorithms, user that lets the user declare termination, and local that uses local task counter pre calculated at init time to detect local termination (as PTG was doing before).

PR #411, opened 2022-08-09
Introduce parsec_taskpool_wait and parsec_taskpool_test

devreal:
From the TTG branches...

@therault please review carefully, there were some conflicts that I had to resolve...

Signed-off-by: Thomas Herault <herault@icl.utk.edu>


devreal:
Ahh, this seems to depend on https://github.com/ICLDisco/parsec/pull/391

therault:
We should review this PR once the termination detection PR is merged.

parsec/runtime.h
```diff
@@ -322,6 +322,41 @@ int parsec_context_test( parsec_context_t* context );
...
+/**
+ * @brief Allow the main thread to temporarily join the computation
+ *    by executing one (or less) task
+ *
+ * @details
+ * Try to progress the execution context until the given taskpool reaches termination.
+ * Upon return from this function, all resources (threads and accelerators)
+ * associated with the corresponding context are left in a mode where they are
+ * active. The taskpool must be ready and registered with a started context.
+ *
```
abouteiller, line 352: Cannot the test set the task pool as completed (i.e., have the effect of an implicit wait), from that documentation I infer that the only correct code is to follow `taskpool_test` with `taskpool_wait` even when `taskpool_test` returned 0.

parsec/scheduling.c
```diff
@@ -532,6 +532,107 @@ int __parsec_task_progress( parsec_execution_stream_t* es,
...
+    PARSEC_PINS(es, SELECT_BEGIN, NULL);
+
+    if( NULL == parsec_current_scheduler ) {
+        parsec_fatal("Main thread entered parsec_taskpool_wait, while a scheduler is not selected yet!");
+        return -1;
+    }
+
+    while( tp->tdm.module->taskpool_state(tp) != PARSEC_TERM_TP_TERMINATED ) {
+        /* Here we detach all dtd taskpools registered with us */
+        parsec_detach_all_dtd_taskpool_from_context(parsec_context);
```
abouteiller, line 601: I do not understand why when we work on a specific task pool, we detach all task pools. Should we just call `parsec_dtd_detach_taskpool_from_context(tp)` here?

parsec/scheduling.c
```diff
@@ -796,3 +897,53 @@ int parsec_context_wait( parsec_context_t* context )
...
+        parsec_warning("taskpool is registered to non-started context in parsec_taskpool_wait\n");
+        return -1;
+    }
+
+    if( __parsec_context_cas_or_flag(context,
+                                     PARSEC_CONTEXT_FLAG_COMM_ACTIVE) ) {
+        (void)parsec_remote_dep_on(context);
+    }
+
+    ret = __parsec_taskpool_test( tp, context->virtual_processes[0]->execution_streams[0] );
```
abouteiller, line 946: This is the only call site, and the `es` argument is set to a "constant" that can be deduced from `tp->context`; is the second argument to `__parsec_taskpool_test` needed?

tests/taskpool_wait/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+include(ParsecCompilePTG)
```
abouteiller, line 1: All these tests should be in the directory `tests/api/taskpool_wait/`

tests/CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ if( TARGET parsec-ptgpp )
 endif( TARGET parsec-ptgpp )
 add_subdirectory(dsl/dtd)
 add_subdirectory(apps)
+add_subdirectory(taskpool_wait)
```
abouteiller, line 78: This add_subdirectory must be moved inside the `TARGET parsec-ptgpp` block above

abouteiller (changes requested):

parsec/scheduling.c
therault: I will take a DTD test that reuses a DTD taskpool in a loop, and change the parsec_context_wait to a parsec_taskpool_wait, and test.

It seems obvious that we need to detach the taskpool if it is a DTD taskpool, and not the other taskpools, but unclear what else we should do.

parsec/scheduling.c
therault: Probably it's a place where we should use `parsec_my_execution_stream()`

bosilca: done.

tests/CMakeLists.txt
bosilca: only because the single test in the directory uses both PTG andDTD taskpools.

tests/taskpool_wait/CMakeLists.txt
bosilca: done.

abouteiller:
Should this be slated for v4.0 release?

parsec/interfaces/dtd/insert_function.c
```diff
@@ -227,6 +227,7 @@ parsec_dtd_dequeue_taskpool(parsec_taskpool_t *tp)
         should_dequeue = 1;
         parsec_list_nolock_remove(tp->context->taskpool_list,
                                   (parsec_list_item_t *)tp);
+        parsec_list_item_singleton((parsec_list_item_t*)tp);
```
abouteiller, line 230: Could use the macro LIST_ITEM_SINGLETON here

parsec/data_distribution.c
```diff
@@ -53,22 +53,46 @@ parsec_data_dist_fini(void)
...
+    }
+#endif  /* defined(PARSEC_DEBUG_PARANOID) */
     dc->ht_item.key = key;
     dc->dc_id = key;
     parsec_hash_table_insert(parsec_dc_hash_table, &dc->ht_item);
     /* We do not have any function to check if there is a collision
      * so for now we replace the old entry if there is request for
      * a new registration with same id.
      */
+    parsec_inform("register dc_t %p with key %lld", dc, (long long)key);
```
abouteiller, line 82: This is way too high a debug level for this, should use parsec_debug_verbose(...)

parsec/interfaces/dtd/insert_function_internal.h
```diff
@@ -228,8 +228,10 @@ struct parsec_dtd_taskpool_s {
     int                          task_window_size;
     int32_t                      task_threshold_size;
     int                          total_tasks_to_be_exec;
-    uint32_t                     local_task_inserted;
-    uint8_t                      function_counter;
+    uint32_t                     local_task_inserted;  /* don't waste an atomic operation
```
abouteiller, line 231: I don't understand this comment

parsec/runtime.h
abouteiller: is the following mockup correct and does that description prevents it? 

```
taskpool_test(tp)
context_wait(ctx)
```

parsec/scheduling.c
```diff
@@ -532,6 +531,108 @@ int __parsec_task_progress( parsec_execution_stream_t* es,
...
+{
+    uint64_t misses_in_a_row;
+    parsec_task_t* task;
+    int nbiterations = 0, distance, rc;
+    struct timespec rqtp;
+
+    rqtp.tv_sec = 0;
+    misses_in_a_row = 1;
+
+    assert(PARSEC_THREAD_IS_MASTER(es));
```
abouteiller, line 587: don't think that assert is correct, what about recursive task pools

parsec/scheduling.c
```diff
@@ -532,6 +531,108 @@ int __parsec_task_progress( parsec_execution_stream_t* es,
     return rc;
 }
 
+static int __parsec_taskpool_test( parsec_taskpool_t* tp, parsec_execution_stream_t *es )
+{
+    parsec_context_t* parsec_context = es->virtual_process->parsec_context;
+    parsec_task_t* task;
+    int nbiterations = 0, distance, rc;
+
+    assert(PARSEC_THREAD_IS_MASTER(es));
```
abouteiller, line 540: don't think that assert is correct, what about recursive task pools
(same comment appears twice)

parsec/scheduling.c
```diff
@@ -532,6 +531,108 @@ int __parsec_task_progress( parsec_execution_stream_t* es,
...
+    PARSEC_PINS(es, SELECT_BEGIN, NULL);
+
+    if( NULL == parsec_current_scheduler ) {
+        parsec_fatal("Main thread entered parsec_taskpool_test, while a scheduler is not selected yet!");
+        return -1;
+    }
+
+    if( tp->tdm.module->taskpool_state(tp) != PARSEC_TERM_TP_TERMINATED ) {
+        /* Here we detach all dtd taskpools registered with us */
+        parsec_detach_all_dtd_taskpool_from_context(parsec_context);
```
abouteiller, line 552: Shouldn't this do the same as `test` below?
```
/* Detach the current taskpool from the associated context */
     parsec_dtd_dequeue_taskpool(tp);
     PARSEC_LIST_ITEM_SINGLETON(tp);
```

tests/api/taskpool_wait/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+include(ParsecCompilePTG)
+
+parsec_addtest_executable(C taskpool_wait SOURCES main.c dtd_tp.c)
+target_include_directories(taskpool_wait PRIVATE $<$<NOT:${PARSEC_BUILD_INPLACE}>:${CMAKE_CURRENT_SOURCE_DIR}>)
+target_ptg_sources(taskpool_wait PRIVATE "ptg_tp.jdf")
+
+if( MPI_C_FOUND )
+    parsec_addtest_cmd(shm_two_dim_band ${MPI_TEST_CMD_LIST} 8 ./taskpool_wait)
```
abouteiller, line 8: name of test target (2Dband) is wrong

abouteiller (changes requested):
Mo' review mo problems

parsec/data_distribution.c
bosilca: This is a real issue, it might or might not work we have no way to know if there will be collision or not between different data collections. So I would rather have this well visible, such that an educated user can fix their code (once it gets annoyed by the output).

parsec/interfaces/dtd/insert_function.c
bosilca: At this point I'm not sure the macro remains necessary in the code base, it always defaults to calling the `parsec_list_item_singleton ` function.

parsec/interfaces/dtd/insert_function_internal.h
bosilca: Reminder that we do not need to manipulate this field with atomic operations.

parsec/runtime.h
bosilca: This is the expected usage, as `taskpool_test` does not alter the state of the context (aka. it remains active).

parsec/scheduling.c
bosilca: We could allow usage of this function in other threads than master. That being said I'm not sure all accessors used in this function are thread safe.

parsec/scheduling.c
bosilca: With the new API we designed few days ago, the use of `parsec_detach_all_dtd_taskpool_from_context` will not be necessary.

parsec/scheduling.c
bosilca: Unlike in test I don't think we should allow any other threads to do a blocking operation on any parsec_objects. We will run out of working threads, and encounter undesired side-effects.

therault:
I'm fixing issues in this pull request. I have a question for all:

Some DSLs (DTD, but others probably will also need the feature) require that something special is done on their taskpools when the user thread enters `context_wait()` or `taskpool_wait()`, and something should also be done when leaving `context_wait()` or `taskpool_wait()`.

For `taskpool_wait()`, it's easy, because we have the taskpool, so we can add a DSL function in the taskpool that the DSL registers.

For `context_wait()`, it's not that easy, because the context does not have a list of taskpools.

The current solution (in master) is that we call some DSL-specific functions (namely `parsec_detach_all_dtd_taskpool_from_context`), and the DSL keeps a list of taskpools. 

I want to make this more generic. I see two ways:

  - the DSL registers a callback to be called in the context. If the callback is already registered, it's the responsibility of the DSL not to register the callback again, and the DSL keeps a private list of taskpools that needs to be managed, similar to what is done today (except that instead of calling `parsec_detach_all_dtd_taskpool_from_context()` directly, we call `dsl_callback_enter_wait()` for each registered DSL.
  - each taskpool has an optional `dsl_enter_wait_cb()` and `dsl_leave_wait_cb()` function, and the context keeps a list of attached taskpools, and iterates over that list every time it enters / leaves context_wait (and we can use the same callbacks for the `taskpool_wait` version).

I'm more in favor of the second approach, because it's less state to manage at the DSL level, but maybe you have a different opinion?

parsec/interfaces/dtd/insert_function.c
abouteiller: the macro does the list_item cast, that's all it does. If we think this is not useful, we should remove the macro then.

parsec/data_distribution.c
abouteiller: looking at the code, I think you are referring to the line 64: registering the same key twice (this is abnormal but harmless, `parsec_info` is adequate), on line 68, the `DEBUG_VERBOSE` should be promoted to a `parsec_warning` because this is a real coding error that will cause problems, this line is unconditionally printing the registration, this is too verbose.

parsec/interfaces/dtd/insert_function.c
bosilca: ok, let me keep the code consistent with the rest (I'll switch to PARSEC_LIST_ITEM_SINGLETON)

parsec/data_distribution.c
bosilca: ok, I will make the change. It is unfortunate that the warnings are only delivered when in heavy debug mode.

therault: See 52caa8bee925eac4c3876079444690662393f69a

parsec/interfaces/dtd/insert_function.c
therault: Already fixed in the current code

parsec/scheduling.c
therault: I think adding the feature that this function can be called from other threads than the master is outside the scope of this PR... As noted by George, there are accesses that are unsafe in this function, and we should keep the assert for now.

parsec/interfaces/dtd/parsec_dtd_data_flush.c
```diff
@@ -326,15 +322,15 @@ parsec_dtd_insert_flush_task_pair(parsec_taskpool_t *tp, parsec_dtd_tile_t *tile
         if(last_writer.task->rank != tile->rank) {
```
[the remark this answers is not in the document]
devreal: Actually every process sees that task. It's not inserted on just one process but only if the last writer of the  tile is not the same as the owner.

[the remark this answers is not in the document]
bosilca: Yes, I misread the code and @therault pointed me on the right direction. Why is the sending task missing one ref count ?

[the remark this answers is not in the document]
devreal: So here is what I found while debugging:

The flush task is created (https://github.com/ICLDisco/parsec/blob/c4cc5b01e46552cef018f90f05bd601ec6d0b350/parsec/interfaces/dtd/parsec_dtd_data_flush.c#L266) "manually" (i.e., not using `parsec_dtd_insert_task`) and then passed to `parsec_insert_dtd_flush_task` where it is retained (https://github.com/ICLDisco/parsec/blob/master/parsec/interfaces/dtd/parsec_dtd_data_flush.c#L134) and released again (https://github.com/ICLDisco/parsec/blob/master/parsec/interfaces/dtd/parsec_dtd_data_flush.c#L239) at the end of the function. That's according to plan, as far as I can see. When the second task of the pair is created, that first task is retained when it's new descendant is set (https://github.com/ICLDisco/parsec/blob/master/parsec/interfaces/dtd/parsec_dtd_data_flush.c#L185). This is good, because eventually this reference will be released once the dependency is released. However, we also release that first task right away because it's not the last writer anymore (https://github.com/ICLDisco/parsec/blob/master/parsec/interfaces/dtd/parsec_dtd_data_flush.c#L225), which is correct (we don't care about tasks anymore if they are not the last writer anymore). It's important though to keep that one reference that will be released once the dependency is released. That's the reference that is added if `is_first_flush_task` is `true`.

Maybe the original sin is that the code doesn't use `parsec_dtd_insert_task` (which hopefully would do just the right thing) but I cannot properly judge that. And I don't know why it wasn't done that way in the first place. My sense is that my solution is a hack, piled on top of more hacks. The manual ref-counting (e.g., https://github.com/ICLDisco/parsec/blob/c4cc5b01e46552cef018f90f05bd601ec6d0b350/parsec/interfaces/dtd/parsec_dtd_data_flush.c#L275) makes reasoning about the life-time hard. I'd prefer doing this properly but I'm afraid that pulling one string will unravel all others.

parsec/interfaces/dtd/insert_function.c
```diff
@@ -2222,9 +2074,9 @@ parsec_dtd_create_task_classv(parsec_dtd_taskpool_t *dtd_tp,
 
     *name_not_const = name;
     tc->task_class_type = PARSEC_TASK_CLASS_TYPE_DTD;
-    tc->task_class_id = dtd_tp->function_counter++;
```
therault, line 2225: This is suspicious and breaking profiling.

parsec/interfaces/dtd/insert_function.c
```diff
@@ -1956,10 +1808,8 @@ parsec_dtd_release_local_task(parsec_dtd_task_t *this_task)
             }
         }
         assert(this_task->super.super.super.obj_reference_count == 1);
-        parsec_taskpool_t *tp = this_task->super.taskpool;
-
         parsec_thread_mempool_free(this_task->mempool_owner, this_task);
-        parsec_taskpool_update_runtime_nbtask(tp, -1);
+// REAZUL        parsec_taskpool_update_runtime_nbtask(tp, -1);
```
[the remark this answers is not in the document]
therault: Sorry, used this while debugging, forgot to remove it. Done.

parsec/interfaces/dtd/insert_function.c
```diff
@@ -2890,7 +2741,7 @@ parsec_insert_dtd_task(parsec_task_t *__this_task)
     parsec_dtd_tile_t *tile = NULL;
 
     /* Retaining runtime_task */
-    parsec_taskpool_update_runtime_nbtask(this_task->super.taskpool, 1);
+//REAZUL    parsec_taskpool_update_runtime_nbtask(this_task->super.taskpool, 1);
```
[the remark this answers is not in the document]
therault: Sorry, used it while debugging, forgot to remove it. Done

parsec/interfaces/dtd/parsec_dtd_data_flush.c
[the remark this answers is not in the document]
therault: I believe this conversation (tile owner rank != tile last user rank) is resolved. Re-open it if I'm wrong?

therault:
The last bug report we have had on this PR was from Sameer, but his case re-uses the same flow as RW in multiple inputs of a task, and it seems the issue is also present on master. A workaround has been found (by passing the not tracking flag to additional copies of the flow), and this should be solved, but should be solved in a separate PR.

Please review the PR, so that we can merge it.

@abouteiller your last review asks for changes, I believe they are implemented, please re-review and change your status.

bosilca (approved):

PR #413, opened 2022-08-10
Don't pass the execution stream around to recursive calls.

### LINKED ISSUE #412: Recursive tasks can generate profiling events on wrong streams (profile corruption)
## Describe the bug
Recursive tasks, such as those used for large tiles in DPLASMA or dense tiles in HiCMA, work by starting a local child taskpool and attaching completion of the parent (asynchronous) task to that of the child taskpool. This is done via `parsec_recursivecall`:
```C
static inline int
parsec_recursivecall( parsec_execution_stream_t    *es,
                      parsec_task_t                *task,
                      parsec_taskpool_t            *tp,
                      parsec_recursive_callback     callback,
                      int nbdesc,
                      ... )
{
    parsec_recursive_cb_data_t *cbdata = NULL;
    /* removed for brevity */
    cbdata = (parsec_recursive_cb_data_t *) malloc(/* size calculation*/);
    cbdata->es       = es;
    cbdata->task     = task;
    cbdata->callback = callback;
    cbdata->nbdesc   = nbdesc;
    /* removed for brevity */
    parsec_taskpool_set_complete_callback( tp, &parsec_recursivecall_callback,
                                           (void *)cbdata );
    parsec_context_add_taskpool( es->virtual_process->parsec_context, tp);
    return -1;
}
```

In short, `parsec_recursivecall` allocates and fills a structure for the recursive callback data—including the execution stream `es` where the parent task executed—sets the completion callback for the child taskpool to `parsec_recursivecall_callback`, and adds the child taskpool to the parent task's context. The completion callback `parsec_recursivecall_callback` is given below:
```C
static inline int parsec_recursivecall_callback(parsec_taskpool_t* tp, void* cb_data)
{
    int i, rc = 0;
    parsec_recursive_cb_data_t* data = (parsec_recursive_cb_data_t*)cb_data;
    /* call user callback *before* we complete and release the task */
    data->callback( tp, data );
    rc = __parsec_complete_execution(data->es, data->task);
    /* removed for brevity */
    return rc;
}
```

This completion callback calls the user's custom completion callback—in DPLASMA and HiCMA this is used in `*potrf` to propagate errors—and then the standard PaRSEC execution completion function `__parsec_complete_execution`. For regular tasks, `__parsec_complete_execution` is called by the runtime after the body of the tasks executes, but for asynchronous tasks it is up to the user to call it—in this case, via the runtime-provided recursive wrapper.

`__parsec_complete_execution` does various tasks around cleaning up task execution, including potentially triggering dependency releases. While doing so, it can generate profiling events. These are written to the profiling stream associated with the execution stream argument passed to `__parsec_complete_execution`.

When `__parsec_complete_execution` is called for a normal task during `__parsec_task_progress` after it completes execution of the task body, the execution stream where it is currently executing on is passed as the argument. Inside `parsec_recursivecall_callback` however, the execution stream passed is `data->es`, which is the execution stream where the _parent_ task originally executed, not the execution stream of wherever this callback is being executed—presumably by the last task in the child taskpool to complete. This is **incorrect** and leads to the profile being corrupted.

### To Reproduce
1. Profile a recursive task
2. Be unlucky and have a corrupted profile
3. ???
4. Profit!

The likelihood of this occurring is higher with many recursive tasks mixed with many short tasks in the parent taskpool as well as with many compute cores.

### Expected behavior
Profiles shouldn't get corrupted by recursive tasks.

I don't think there's currently a way to get the _current_ execution stream from inside the completion callback. The most straightforward solution is to change that, so that PaRSEC passes the execution stream where `on_complete` is called to the callback, but that requires changes to the API, which may not be palatable. In any case, the current behavior is simply **wrong**.

## Additional context
This problem has been frustrating to debug, but I'm fairly certain I've determined the cause.

bosilca:
Thanks @omor1 for the bug report.

Fixes #412.

### DIFF
--- a/parsec/recursive.h
+++ b/parsec/recursive.h
@@ -16,7 +16,6 @@ typedef struct parsec_recursive_callback_s parsec_recursive_callback_t;
 typedef void (*parsec_recursive_callback_f)(parsec_taskpool_t*, const parsec_recursive_callback_t* );
 
 struct parsec_recursive_callback_s {
-    parsec_execution_stream_t    *es;
     parsec_task_t                *task;
     parsec_recursive_callback_f   callback;
     int nbdesc;
@@ -26,12 +25,13 @@ struct parsec_recursive_callback_s {
 static inline int parsec_recursivecall_callback(parsec_taskpool_t* tp, void* cb_data)
 {
     parsec_recursive_callback_t* data = (parsec_recursive_callback_t*)cb_data;
+    parsec_execution_stream_t *es = parsec_my_execution_stream();
     int i, rc = 0;
 
     /* first trigger the internal taskpool completion callback */
     data->callback( tp, data );
     /* then complete the generator task */
-    rc = __parsec_complete_execution(data->es, data->task);
+    rc = __parsec_complete_execution(es, data->task);
     /* and finally release the data associated with the inner taskpool */
     for( i = 0; i < data->nbdesc; i++ ) {
         parsec_tiled_matrix_destroy( (parsec_tiled_matrix_t*)(data->desc[i]) );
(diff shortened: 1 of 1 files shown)


therault (approved):

devreal (approved):
LGTM although I am not sure I grasp what recursive tasks are ^^

omor1:
Whoops, you call `parsec_context_add_taskpool( es->virtual_process->parsec_context, tp)` but don't have `es`. Can probably just change it to passing in `context`.

PR #414, opened 2022-08-12
Fix bug introduced in pr #413 (recursive missing execution stream param)

abouteiller:


therault:
Already fixed in https://github.com/ICLDisco/parsec/commit/c4dbbd61fb8a2ffb38566648159b63cf19668c8d

PR #415, opened 2022-08-12
Protect the inline functions by they device support.

bosilca:
We use the type of the BODY to add a #ifdef around the generated code
used by the generated code path for the device. This removes a large
number of warnings from DPLASMA.

Protect the code generated for the simulations using the same mechanism.


therault (approved):
Looks good to me, but CI testing fails because of a problem when parsec.y changes and we try to compile in release mode.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -6642,8 +6642,6 @@ static void jdf_generate_code_hook_cuda(const jdf_t *jdf,
                 "#endif /* PARSEC_PROF_TRACE */\n");
     }
 
-    dyld = jdf_property_get_string(body->properties, "dyld", NULL);
```
abouteiller, line 6645: This change is unrelated (and incomplete if intentional, we still have a lot of infrastructure to do the 'dyld' thing, and DPLASMA is very dependent on it)

parsec/interfaces/ptg/ptg-compiler/parsec.y
```diff
@@ -526,7 +526,24 @@ body:         BODY_START properties BODY_END
              {
                  jdf_body_t* body = new(jdf_body_t);
                  body->properties = $2;
+                 /**
+                  * Go over the list of propertiues and tag them with the type of the BODY. This
```
abouteiller, line 530: typo Properties

parsec/interfaces/ptg/ptg-compiler/parsec.y
```diff
@@ -526,7 +526,24 @@ body:         BODY_START properties BODY_END
...
+                  * allows us to protect the generated code to avoid compiler warnings.
+                  */
+                 jdf_expr_t* type_str = jdf_find_property( body->properties, "type", NULL );
```
abouteiller, line 533: Should the name be more specific than "type", for example "architecture" or "accelerator" or something else?

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
bosilca: it is intentional, the property was extracted twice without modification between them. So this call is unnecessary.

parsec/interfaces/ptg/ptg-compiler/parsec.y
bosilca: This is what we are using right now. We can change it to something more meaningful but then we need to a larger PR.

abouteiller (approved):

PR #416, opened 2022-08-12
Fix/more warnings

bosilca:
A bunch of issues/warnings were fixed in this PR.

### DIFF
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -1770,7 +1770,7 @@ remote_dep_mpi_put_end_cb(parsec_comm_engine_t *ce,
             ((remote_dep_cb_data_t *)cb_data)->k, deps, lreg, rreg);
 
 #if defined(PARSEC_PROF_TRACE)
-    parsec_profiling_ts_trace(MPI_Data_plds_ek, ((remote_dep_cb_data_t *)cb_data)->k, PROFILE_OBJECT_ID_NULL, NULL);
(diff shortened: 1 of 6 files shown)


parsec/remote_dep_mpi.c
```diff
@@ -1770,7 +1770,7 @@ remote_dep_mpi_put_end_cb(parsec_comm_engine_t *ce,
             ((remote_dep_cb_data_t *)cb_data)->k, deps, lreg, rreg);
 
 #if defined(PARSEC_PROF_TRACE)
-    parsec_profiling_ts_trace(MPI_Data_plds_ek, ((remote_dep_cb_data_t *)cb_data)->k, PROFILE_OBJECT_ID_NULL, NULL);
```
therault, line 1773: I believe we used the `_ts_` version here because that function might be called by other threads than the comm thread occasionally? With `TAKE_TIME(parsec_comm_es.es_profile, ...)` we might corrupt the profile file if that is the case.

bosilca: I don't think any other thread can call the put_end callback. Out of all callback types the start and end for data transfers are strictly limited to the communication thread.

omor1: The only one that is supposed to be able to be called from other threads (right now, anyway) is `remote_dep_nothread_send`; this generates the Activation Send events. (In my branch I also have Activation Pack and Activation Recv events; the Activation Pack is used mostly to count which events are being sent in an activation.)

devreal (approved):

abouteiller (approved):

abouteiller:
you need to resign some commits @bosilca

PR #417, opened 2022-08-12
remove an unused variable in dbpreader

abouteiller:


therault:
Have a look at https://github.com/ICLDisco/parsec/pull/416, it has this fix too

bosilca:
This is already part of #416.

PR #418, opened 2022-08-12
tools/profiling: PTT v2

omor1:
Add new 'version 2' of the Parsec Trace Table format, where instead of one giant table for events with many NA values for keys and all columns stored as `object`, there is one 'main' table with just the base event information (`node_id`, `stream_id`, `taskpool_id`, `type`, `begin`, `end`, `flags`, `id`) and an additional 'info' table for each event type.

In a `ParsecTraceTables` object there is a new field, `event_infos`; this is a `dict` mapping `event_type -> DataFrame`. Each DataFrame will have a separate set of type-specific columns (keys); each row (event) is indexed by the corresponding event in the 'main' table, `events`—notably, this means that the indices for each DataFrame in `event_infos` are _not_ linear. The expectation is that most users will look for events in the 'main' DataFrame and use the indices of these events to index into the `event_infos` DataFrames. This PR also fixes a longstanding issue where the `events` DataFrame would have duplicated indices, due to how the DataFrames are concatenated.

The default format remains 'version 1', but 'version 2' can be selected via the new `--ptt-version` argument to `profile2h5`. A second patch adds new `--complib` and `--complevel` arguments to allow PyTables to compress the generated tables—the defaults here remain the same as before, `blosc` and `0`, respectively. There is a minor addition in `pbt2ptt.pyx` that checks the return value from each process—this is to ensure that we didn't crash in e.g. the C code from `dbpreader.c`, which can happen when the profile is corrupt (see #412).

Please test that I a) haven't screwed up the 'version 1' format—I made changes that _shouldn't_ change the resulting DataFrame, but I haven't thoroughly vetted this yet—and b) the 'version 2' format does what we want it to.

### DIFF
--- a/tools/profiling/python/parsec_trace_tables.py
+++ b/tools/profiling/python/parsec_trace_tables.py
@@ -88,11 +88,12 @@ class ParsecTraceTables(object):
 
     # the init function should not ordinarily be used
     # it is better to use from_hdf(), from_native(), or autoload()
-    def __init__(self, events, event_types, event_names, event_attributes, event_convertors,
+    def __init__(self, events, event_infos, event_types, event_names, event_attributes, event_convertors,
                  nodes, streams, information, errors):
         self.__version__ = self.__class__.class_version
         # core data
         self.events = events
+        self.event_infos = event_infos
         self.event_types = event_types
         self.event_names = event_names
         self.event_attributes = event_attributes
@@ -106,13 +107,15 @@ class ParsecTraceTables(object):
                complevel=0, complib='blosc'):
         if not overwrite and os.path.exists(filename):
             return False
+        table_format = 't' if table else 'f'
         store = pd.HDFStore(filename + '.tmp', 'w', complevel=complevel, complib=complib)
         for name in ParsecTraceTables.HDF_TOP_LEVEL_NAMES:
             store.put(name, self.__dict__[name], encoding='ascii')
-        if table:
-            store.put('events', self.events, format='t', append=append)
-        else:
-            store.put('events', self.events, format='f', append=append)
+        store.put('events', self.events, format=table_format, append=append)
+        for event_type, event_info in self.event_infos.items():
+            # store each event info into events::{event_type}
+            key = 'event_infos_{}'.format(event_type)
+            store.put(key, event_info, format=table_format, append=append)
         store.close()
         # do atomic move once it's finished writing,
         # so as to allow Ctrl-Cs without secret breakage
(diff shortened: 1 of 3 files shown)


bosilca:
There is no impact on the script that use the profiling data ?

omor1:
> There is no impact on the script that use the profiling data ?

Not as long as `--ptt-version=1` is used, which is the default. There will be a new `event_infos` dict, but it will be empty. I don't think adding backwards-compatibility between v1 scripts and v2 profiles is a priority, since it's hard to have the "correct" behavior in all circumstances. There are too many ways to index a DataFrame, and any operations that result in a new DataFrame might need to be wrapped too...

omor1:
I've been testing the v2 format a bit, it seems to work correctly. Loading from disk seemed a lot faster, despite the large file size, and the data is all there, from what I can tell. It's perhaps a bit clunkier since you need to reference multiple tables, it might be worth it to have a utility function that can merge events of a particular type from the main table with its infos:
```python
def merge_events_info(trace, type):
    if isinstance(type, str):
        type = trace.event_types.at[type]
    events = trace.events.loc[lambda df: df.type == type]
    infos = trace.event_infos[type].loc[events.index]
    return pd.concat([events, infos], axis=1)
```

Alternatively, it could be possible to instead of having a giant table with all event base infos + side tables for event infos per type to have tables per type with all event infos together. I find that usually I want to examine each type of event separately.

As _another_ alternative, it could be possible to use Pandas Sparse columns and/or Nullable dtypes in the user-facing presentation—which is convenient from a usability standpoint—and convert it to/from a "native" format with dense tables per-type and "native" dtypes when storing to/loading from HDF5.

therault:
Looking at the code, I think I like the alternative version better, but it really depends on only one thing: the efficiency of join/merge/concatenate, and which one to use (if there is one that is efficient).

With one table per type, with all columns (begin, end, etc...), that forces scripts that look at all events to always join them or merge them. If that operation creates a full in-memory copy of the tables, that's not possible.

But if the join/merge/concatenate is efficient and does not copy the tables but refer to them, then it becomes easy for a script to take the format v2 and treat it as a v1, and scripts that work on a per-type basis are much simpler.

I'll see if I can do a test of pandas join/merge/concatenate.

omor1:
> If that operation creates a full in-memory copy of the tables, that's not possible.
> But if the join/merge/concatenate is efficient and does not copy the tables but refer to them

Pandas tries to reduce the number of copies and usually has different objects refer to the same underlying data, but this is usually rather opaque to the user and it isn't obvious whether an operation will create a copy or not.

bosilca (approved):

omor1:
I would have held off on merging this—I haven’t tested much that the v1 code is correct, I’m not sure anyone else has done so either.

PR #419, opened 2022-08-15
hotfix: duplicate variable d2h_max_flows

abouteiller:
https://github.com/spack/spack/issues/32125

We will need to produce a new v3.x release for this bug to appear fixed on the upstream Spack channel

### DIFF
--- a/parsec/mca/device/cuda/transfer.c
+++ b/parsec/mca/device/cuda/transfer.c
@@ -165,7 +165,7 @@ static const parsec_symbol_t symb_CUDA_d2h_task_param = {
     .flags = 0x0
 };
 
-int32_t parsec_CUDA_d2h_max_flows = 0;
+extern int32_t parsec_CUDA_d2h_max_flows;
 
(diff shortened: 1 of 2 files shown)


parsec/mca/device/cuda/transfer.c
```diff
@@ -165,7 +165,7 @@ static const parsec_symbol_t symb_CUDA_d2h_task_param = {
     .flags = 0x0
 };
 
-int32_t parsec_CUDA_d2h_max_flows = 0;
+extern int32_t parsec_CUDA_d2h_max_flows = 0;
```
devreal, line 168: We shouldn't mark variables as `extern` in compilation units. Can we put them in a header?

abouteiller: This is fixed in master (exactly as you propose), I'm just taking the path of least resistance here to hot fix v3.

bosilca:
let's call it 3.1. In fact we had a nicer naming scheme for our releases, why are we not using it here ?

Issue #396, opened 2022-07-08
Profiling: Recursive and other Asynchronous Tasks


omor1:
## Description
PaRSEC recursive tasks are currently problematic to profile, especially in a distributed context. Recursive tasks are not modeled explicitly by the DSL; instead, they are set up by
1. Generating task creates a new taskpool with only local data
2. Generating task adds it to the current context and connects its own completion to that of the child taskpool via `parsec_recursivecall`
3. Generating task returns with `PARSEC_HOOK_RETURN_ASYNC`.

From the runtime's perspective then, it is impossible to differentiate between a recursive task and any other asynchronous task. In terms of profiling, the time measured is the body of the generating task—not the actual computation; this affects all asynchronous tasks. To determine the actual time for compute, one needs to look at the compute time of tasks in the child taskpool.

Connecting between the child taskpool and parent taskpool is nontrivial, however. Taskpool IDs are not synced between nodes when generating recursive taskpools, so there are duplicate tpids on different nodes. For some computations, such as 3-flow HiCMA, it is possible to determine the relationship between the generating parent `HiCMA_dpotrf_L_3flow::potrf_dpotrf` task and the child `dpotrf_L` taskpool by virtue of knowing that all the dpotrf tasks are necessarily serial and generated in order, so that the order of tpids and the order of dpotrf tasks on each node is the same. However, for other computations, this order is not predictable a priori, so this becomes an intractable task.

### Describe the solution you'd like
The best solution, in my opinion, is to record in the info of a generating recursive task the taskpool ID of the child taskpool—that allows one to both measure only the time of execution bodies and easily find all child tasks from the parent task. This may be annoying to implement, however, since recursive tasks appear identical to other asynchronous tasks. Suggestions of how to fix this are welcome. Integrating recursive tasks more concretely into the DSL rather than the current ad hoc implementation would be one way, but is much more invasive.

### Describe alternatives you've considered
A workaround to the above might be to add a local that stores the tpid of the child taskpool—since it's a local, it will be recorded in the profiling info. This needs to be done for _every_ recursive task though.

[...]

PR #421, opened 2022-08-17
profiling: fix multiple EXEC_END events in some circumstances

omor1:
Tasks with multiple incarnations would record EXEC_END profiling events
for each incarnation. The profiling tools would match with the first
event, which is incorrect if that isn't the incarnation that was
executed. Instead, only record EXEC_END after an incarnation hook is
successfully executed.

### DIFF
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -179,13 +179,14 @@ int __parsec_execute( parsec_execution_stream_t* es,
         rc = hook( es, task );
 #if defined(PARSEC_PROF_TRACE)
         task->prof_info.task_return_code = rc;
-        PARSEC_PINS(es, EXEC_END, task);
 #endif
         if( PARSEC_HOOK_RETURN_NEXT != rc ) {
             if( PARSEC_HOOK_RETURN_ASYNC != rc ) {
                 /* Let's assume everything goes just fine */
                 task->status = PARSEC_TASK_STATUS_COMPLETE;
             }
+            /* Record EXEC_END event only for incarnation that succeeds */
+            PARSEC_PINS(es, EXEC_END, task);
             return rc;
         }
     next_chore:
@@ -197,8 +198,10 @@ int __parsec_execute( parsec_execution_stream_t* es,
                 break;
     } while(NULL != tc->incarnations[chore_id].hook);
     assert(task->status == PARSEC_TASK_STATUS_HOOK);
-    /* We're out of luck, no more chores */
+    /* Record EXEC_END event to ensure the EXEC_BEGIN is completed
+     * return code was stored in task_return_code */
     PARSEC_PINS(es, EXEC_END, task);
+    /* We're out of luck, no more chores */
     return PARSEC_HOOK_RETURN_ERROR;
 }
 


parsec/scheduling.c
```diff
@@ -179,13 +179,14 @@ int __parsec_execute( parsec_execution_stream_t* es,
...
         task->prof_info.task_return_code = rc;
-        PARSEC_PINS(es, EXEC_END, task);
 #endif
         if( PARSEC_HOOK_RETURN_NEXT != rc ) {
             if( PARSEC_HOOK_RETURN_ASYNC != rc ) {
                 /* Let's assume everything goes just fine */
                 task->status = PARSEC_TASK_STATUS_COMPLETE;
             }
+            /* Record EXEC_END event only for incarnation that succeeds */
+            PARSEC_PINS(es, EXEC_END, task);
```
bosilca, line 189: Not sure this is what we need. As an example recursive tasks will now have an ending that is based on the moment they return, not on the moment they really complete. Thus, my first reaction was to move the `EXEC_END` event in the `if` above. But if we do this then `EXEC_END` should measure the time the runtime think a task took, so we will need to generate the same event when a task is brought back from the ASYNC mode.

omor1: > recursive tasks will now have an ending that is based on the moment they return, not on the moment they really complete.

This is the current behavior; I thought it was intended and so maintained it. If we want to change this behavior then `EXEC_END` will need to be recorded in `__parsec_complete_execution`. It's a bit misleading to do that though, since that would mean that the "duration" of a recursive task and other asynchronous tasks includes time spent in the PaRSEC runtime or doing other work. See #396 for additional discussion on this topic.

abouteiller:
Isn't the problem that we match the last EXEC_END with the first EXEC_START, the result would be logical if we matched the first with the first, second with second, etc. The surfaces thus defined should never overlap.

omor1:
> Isn't the problem that we match the last EXEC_END with the first EXEC_START, the result would be logical if we matched the first with the first, second with second, etc. The surfaces thus defined should never overlap.

The problem is that for a particular <node, taskpool, task class, id> tuple, there should only ever be _one_ `EXEC_START` and `EXEC_END`. The existing code will record an `EXEC_END` for _each_ incarnation it tries until one executes successfully. The matching code will match a start event with the first matching end event it finds, so it will always match and `EXEC_START` with the `EXEC_END` of the first incarnation that it tried to execute, regardless of whether it was successful.

bosilca:
Going back to this after digging into the code for a bit. First, let me confirm that we do have the imbalance described above, one START event for multiple END events. However, all but the last END event have a return code that is not 0 (meaning they are not the final END event). So I guess the issue here is with the python profiling system, it ignores the return code in the END event, and matches the start event with one of the END (more or less randomly).
There are multiple way to fix this:

1. Ignore all END events that have the return code non-zero. This will match the START with the last END of a task, providing information about the span of the task. Good but far from perfect, especially for the visualizers (they dont usually like overlapping tasks).
2. keep the current matching of the START with the first END event, which should occur in the same stream. But if the return code of the END event is not zero, mark the event as partial with duplicate, and then for every other END event automatically create a start with the same timestamp. Then we will need to change the profiling to deal with multiple events.

omor1:
> So I guess the issue here is with the python profiling system, it ignores the return code in the END event, and matches the start event with one of the END (more or less randomly).

Not randomly—it matches with the first `EXEC_END` i.e. the first incarnation tried.

> 1. Ignore all END events that have the return code non-zero. This will match the START with the last END of a task, providing information about the span of the task. Good but far from perfect, especially for the visualizers (they dont usually like overlapping tasks).

This is problematic for some task types, e.g. it is valid for async/recursive tasks to return nonzero, since they complete, well, asynchronously. See #396 for some discussion on the topic. It's probably  better to _specifically_ ignore return code `PARSEC_HOOK_RETURN_NEXT`.

> 2. keep the current matching of the START with the first END event, which should occur in the same stream. But if the return code of the END event is not zero, mark the event as partial with duplicate, and then for every other END event automatically create a start with the same timestamp. Then we will need to change the profiling to deal with multiple events.

I'm not a fan of this, in part because there doesn't exist a mechanism to re-record an event in the current profiling code, and it violates some assumptions that profiling streams are always linear in time. This also just sounds confusing, at least how you worded it.

---

How about moving the start of `EXEC_BEGIN` to occur right before the incarnation's `hook` is called? That would result in multiple non-overlapping events for a single task, but only one would have a return code that is not `PARSEC_HOOK_RETURN_NEXT`. As a downside, the profile consumer might need to remove those extraneous events and it could increase the number of events by a lot—e.g. for DPLASMA taskpools with recursive tasks, for each task type it first checks if the tile size is too large and, if so, will split it into a subcomputation; otherwise it calls into BLAS/LAPACK. So this would double number of events for many DPLASMA computations, with those being short-duration "tasks" that didn't actually compute anything and bailed out to the next incarnation.

bosilca:
After reevaluating this a second time, I think it is the best we can do with the current infrastructure.

PR #423, opened 2022-08-17
Create a Priority Controlled Binding scheduler

therault:
This PCB scheduler uses some bits (MCA defined) of the priority word to define which 'group' of threads can schedule a task with this priority.

There is additional documentation in sched_pcb.h for more details.

The PR also includes a test that prints where tasks are executed, to check if the policy works.


omor1:
I feel like the problem with this approach is that priorities are (generally) an aspect of the algorithm that is to be used by the scheduler to make scheduling decisions, while the scheduling groups are essentially an aspect of the hardware. It's really hard then to create portable priorities, since the priorities might need to be aware of the hardware details. The current limitation that priorities are 32-bit is also problematic for stealing any of those bits for hardware scheduling groups; we already run into overflow problems with e.g. large potrf problems with small tiles.

bosilca:
As discussed on 03/31/23 we need to add a property to the PTG tasks to indicate the core/device they want to be executed on.

PR #424, opened 2022-08-19
TTG/building system

therault:
These commits are the ones missing for TTG support, which uses PaRSEC as a submodule.

There is one commit that still needs to be considered (not part of this PR):

[352014fdbf10a141fe9f4eab7e70cac6949e638d](https://github.com/TESSEorg/parsec/commit/352014fdbf10a141fe9f4eab7e70cac6949e638d) [Restore the C11 support broken in 694d1e921 and few other things]. I think it was a backport from the master, but too many things have changed to make sure.

-- We need to check if this is still needed (or what parts of it are needed).


parsec/parsec_internal.h
```diff
@@ -347,35 +348,35 @@ parsec_dependency_t *parsec_hash_find_deps(const parsec_taskpool_t *tp,
                                            parsec_execution_stream_t *es,
                                            const parsec_task_t* task);
 typedef int (parsec_update_dependency_fn_t)(parsec_taskpool_t *tp,
-                                            const parsec_task_t* restrict task,
```
devreal, line 350: Just curious: why remove the `restrict` here?

tests/CMakeLists.txt
```diff
@@ -27,7 +27,11 @@ function(parsec_addtest_executable lang target)
   set(options NODEFAULTBUILD)
   set(multiValueArgs SOURCES)
   cmake_parse_arguments(PARSEC_ADDTEST "${options}" "" "${multiValueArgs}" ${ARGN})
-  add_executable(${target} ${PARSEC_ADDTEST_SOURCES})
+  if("${CMAKE_SOURCE_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}")
```
devreal, line 30: Should we document somewhere that tests are only built in stand-alone builds?

parsec/parsec_internal.h
therault: Because this .h file is pulled in by parsec/ttg.h, and restrict is a keyword that is refused by g++ and/or other C++ compilers, even if the declaration is within an `extern "C" { ... }` block. At least, this is what I remember of this issue.

devreal: Ahh right, we need the header for `parsec_task_s` (and probably some other things). It's a bit ironic that an external project includes `parsec_internal.h` though ^^

therault: Some things will probably need cleaning, but I wouldn't say TTG is 'external' from PaRSEC... There is a shade area here :)

CMakeLists.txt
```diff
@@ -904,7 +904,7 @@ set(CTEST_CUDA_LAUNCHER_OPTIONS
       "" CACHE STRING
       "Options to pass to the CTEST_MPI_LAUNCHER to select GPU nodes")
 
-if( BUILD_TESTING )
+if( BUILD_TESTING AND "${CMAKE_SOURCE_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}" )
```
abouteiller, line 907: If you are in "BUILD_TESTING" mode and you integrate parsec as a submodule, I'd argue that you want to run the testers on PaRSEC as well.

cmake_modules/FindHWLOC.cmake
```diff
@@ -72,13 +72,14 @@ if(HWLOC_FOUND)
 
   #===============================================================================
   # Import Target ================================================================
-  if(NOT TARGET HWLOC::HWLOC)
-    add_library(HWLOC::HWLOC INTERFACE IMPORTED)
-  endif(NOT TARGET HWLOC::HWLOC)
+  if(NOT TARGET hwloc)
```
abouteiller, line 75: Why the  name change here, this is not canonical (although the all caps we had before was not either), should be `hwloc::hwloc`

cmake_modules/FindHWLOC.cmake
```diff
@@ -72,13 +72,14 @@ if(HWLOC_FOUND)
...
+    add_library(hwloc INTERFACE IMPORTED)
+  endif(NOT TARGET hwloc)
 
-  set_property(TARGET HWLOC::HWLOC PROPERTY INTERFACE_LINK_LIBRARIES "${HWLOC_LIBRARY}")
-  set_property(TARGET HWLOC::HWLOC PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${HWLOC_INCLUDE_DIR}")
-  set_property(TARGET HWLOC::HWLOC PROPERTY INTERFACE_COMPILE_OPTIONS "${HWLOC_DEFINITIONS}")
+  set_property(TARGET hwloc PROPERTY INTERFACE_LINK_LIBRARIES "${PC_HWLOC_LIB}")
```
abouteiller, line 79: This may cause problems when using a Hwloc that is not system installed (not found by Pkg-config) but is findable by other means (e.g., user set HWLOC_ROOT) as this will mix the flags found from Pkg-config with the flags found from find_library.

parsec/parsec_internal.h
bosilca: I don't think the right approach is to remove this optimization keyword. Instead I would propose we protect restrict in parsec_config or parsec_config_bottom for C++ case, and leave it's usages in the C library as intended originally.

tests/CMakeLists.txt
```diff
@@ -50,12 +60,14 @@ endfunction(parsec_addtest_executable)
 # name in that case.
 #
 function(parsec_addtest_cmd target)
```
bosilca, line 62: This looks like a very unexpected side effect (aka. the macro does not do anything if parsec is not the main project). I would rather create a special testing configuration for parsec, and add all our tests in there. The configuration does not have to be activated by the upstream project.

therault: After looking at the documentation for CMake `add_test`, I think the better option is to keep doing `add_test` unconditionally, and TTG's CMake, should define `BUILD_TESTING` to OFF when configuring parsec. I removed the conditional thing for this function.

tests/CMakeLists.txt
therault: There is a note in the comments of this macro

parsec/parsec_internal.h
therault: The macro trick is now implemented.

abouteiller: This undef of the macro will have side effects on users that include `parsec.h`

parsec/include/parsec/parsec_config_bottom.h
```diff
@@ -36,6 +36,14 @@
 #define END_C_DECLS            /* empty */
 #endif
 
+/* If we're in C++, redefine restrict to nothing */
+#if defined(c_plusplus) || defined(__cplusplus)
+#  if defined(restrict)
```
devreal, line 41: I suggest to use something along the lines of `parsec_restrict` and not fiddle with reserved keywords.

parsec/parsec_internal.h
therault: Well, it will have a side effect only if the users are using a C++ compiler...

According to stackoverflow, some C++ compilers do support `restrict` or `__restrict__`... And others don't.
I guess we should have some CMake logic to decide if HAVE_RESTRICT or not, and have a corresponding cmakedefine.

CMake has a `c_restrict` low-level C compiler feature testable via `target_compile_features`, but there is no such thing in the C++ compiler feature set. I could try to compile some code with `restrict` with the C++ compiler to see if it's supported. But the macro we define would need to be conditional to the type of compiler anyway.

Comment below if you think I should implement this solution.

devreal: Does anyone remember why restrict was added in the first place? We don't use it in other places where tp and tasks are passed, so maybe dropping it here is just ok?

bosilca (approved):

parsec/parsec_internal.h
abouteiller: This is a function pointer and it cannot be inlined, so as long as the compiler doesn't barf on the incompatible function type signatures I don't see what good the restrict does here.

bosilca:
Let's get this done. I agree with @abouteiller comment, restrict has little meaning in the function prototype, other than forcing all functions to support it. Instead, we can add it to the function implementation, allowing the compiler to do optimizations. Also, please squash all the commits into one or two (no more commits with revert).

therault:
I rebased and squashed all commits. I fixed the comments.

PR #323, opened 2022-03-02
[BBT#552] Cleanup the logic of the insert task.

bosilca:
https://bitbucket.org/icldistcomp/parsec/pull-requests/552

It overlaps with one of Thomas's PR but we will sort it out later. Right
now this is WIP, I plan to continue cleaning the handling of read
dependencies and splitting

PR #425, opened 2022-08-19
Remove startup tasks from DTD

therault:
This is the only commits from the 'various' set of commits in parsec-ttg that have not been merged into master yet.

I believe there is another branch/PR with a larger set of cleanup for DTD, so maybe we just want to drop this.

### DIFF
--- a/parsec/mca/termdet/user_trigger/termdet_user_trigger_module.c
+++ b/parsec/mca/termdet/user_trigger/termdet_user_trigger_module.c
@@ -156,7 +156,7 @@ static void parsec_termdet_user_trigger_monitor_taskpool(parsec_taskpool_t *tp,
 {
     parsec_termdet_user_trigger_monitor_t *monitor;
     assert(&parsec_termdet_user_trigger_module.module == tp->tdm.module);
-    assert(NULL == tp->tdm.monitor);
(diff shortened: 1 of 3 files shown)


bosilca:
This is indeed a part among the goodies of #323, but we can merge it as a standalone effort, and I will update my branch.

bosilca (approved):

parsec/mca/termdet/user_trigger/termdet_user_trigger_module.c
```diff
@@ -156,7 +156,7 @@ static void parsec_termdet_user_trigger_monitor_taskpool(parsec_taskpool_t *tp,
 {
     parsec_termdet_user_trigger_monitor_t *monitor;
     assert(&parsec_termdet_user_trigger_module.module == tp->tdm.module);
-    assert(NULL == tp->tdm.monitor);
```
abouteiller, line 159: should this be reinstated or removed?

therault: The patch says to remove it

abouteiller: no the patch comments it line below

PR #426, opened 2022-08-25
More data transfer statistics between devices

therault:
There are two commits. In the first commit, we just display the information we gathered but did not display, in the second we refine the type of information we keep, and display it.

### DIFF
--- a/parsec/mca/device/device.h
+++ b/parsec/mca/device/device.h
@@ -141,7 +141,7 @@ struct parsec_device_module_s {
 #if defined(PARSEC_PROF_TRACE)
     parsec_profiling_stream_t *profiling;
 #endif  /* defined(PROFILING) */
-    uint8_t data_in_array_size; /**< Current size of the data_in_from_device array. */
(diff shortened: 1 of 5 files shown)


parsec/mca/device/device.h
```diff
@@ -283,6 +285,38 @@ parsec_device_find_function(const char* function_name,
     MCA_BASE_VERSION_2_0_0, \
         "device", 2, 0, 0
 
+/**
```
bosilca, line 288: This approach seems costly and not more difficult to implement (as an example I'm not sure it is working without marking `data_in_array_size` as volatile). I think a simple approach is to allocate the `data_in_from_device` array during the device attach. If the specialized device->attach allocate it's own array then so be it, otherwise the `parsec_mca_device_attach` will allocate it as needed.

therault: Right, I didn't know that attach happens after all modules have been selected, this is a better place than dynamic sizing of this array. See commit 38b7b3e16f126731f8cef41f97d91d085ae4b56c

bosilca (approved):

PR #427, opened 2022-08-29
Fix possible access to unowned memory in `parsec_dtd_task_class_add_chore`

DSMishler:
proper count for incarnations - used defined names rather than hardcoded numbers.

Since a dtd taskpool is initialized with 7 (`PARSEC_DEV_MAX_NB_TYPE+1`) chores as space (6 if you consider that this array is null-terminated), it doesn't make sense to allow checking of up to 8 chores in `parsec_dtd_task_class_add_chore`. This could mean accessing unallocated memory. This PR is simply a patch for that issue.

### DIFF
--- a/parsec/interfaces/dtd/insert_function.c
+++ b/parsec/interfaces/dtd/insert_function.c
@@ -2434,7 +2434,7 @@ int parsec_dtd_task_class_add_chore(parsec_taskpool_t *tp,
                                     int device_type,
                                     void *function)
 {
-    __parsec_chore_t **pincarnations;
+    __parsec_chore_t *incarnations;
     int i;
 
     if( tc->task_class_type != PARSEC_TASK_CLASS_TYPE_DTD ) {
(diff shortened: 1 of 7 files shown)


parsec/interfaces/dtd/insert_function.c
```diff
@@ -2447,16 +2447,16 @@ int parsec_dtd_task_class_add_chore(parsec_taskpool_t *tp,
     parsec_dtd_task_class_t *dtd_tc = (parsec_dtd_task_class_t*)tc;
 
     /* We assume that incarnations is big enough, because it has been pre-allocated
-     * with 7 chores, as this is a DTD task class */
+     * with PARSEC_DEV_MAX_NB_TYPE+1 chores, as this is a DTD task class */
     pincarnations = (__parsec_chore_t **)&dtd_tc->super.incarnations;
```
devreal, line 2451: @bosilca why is it `pincarnations = (__parsec_chore_t **)&dtd_tc->super.incarnations;` and not `pincarnations = (__parsec_chore_t **)dtd_tc->super.incarnations;` I don't see that `dtd_tc->super.incarnations` is assigned a new value and instead all accesses have to add another dereference `(*pincarnations)[i]`...

bosilca: Fixed.

bosilca (approved):

DSMishler:
Did a forced update to re-sign my commit on my banch.

PR #428, opened 2022-08-31
Profiling dsl extension

therault:
- Extend the parsec_task_class to make the DSLs define what data to profile for tasks
 - Extend the profiling interface to allow writing the infos related to an event directly in an event buffer instead of assuming infos are always a contiguous block
 - Use the extended task class interface to have programmer-defined parameters profiled as infos for a task in DTD
 - Port PTG to the extended task class interface
 - Enable a couple of tests for PTG and DTD

### DIFF
--- a/tools/profiling/dbpreader.c
+++ b/tools/profiling/dbpreader.c
@@ -382,6 +382,12 @@ static const dbp_event_t *dbp_iterator_next_buffer(dbp_event_iterator_t *it)
     assert( it->current_events_buffer->buffer_type == PROFILING_BUFFER_TYPE_EVENTS );
 
     next_off = it->current_events_buffer->next_buffer_file_offset;
+    if(next_off == -1) {
+        it->current_event.native = NULL;
+        it->current_events_buffer = NULL;
+        return NULL;
+    }
+
     return dbp_iterator_set_offset(it, next_off);
 }
 
--- a/parsec/interfaces/dtd/insert_function.c
+++ b/parsec/interfaces/dtd/insert_function.c
@@ -830,6 +830,28 @@ parsec_dtd_add_profiling_info_generic(parsec_taskpool_t *tp,
     free(str);
 }
 
+void *parsec_dtd_task_profile_info(void *dst, const void *task_, size_t size)
+{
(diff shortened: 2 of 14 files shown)


tools/profiling/dbpreader.c
```diff
@@ -382,6 +382,12 @@ static const dbp_event_t *dbp_iterator_next_buffer(dbp_event_iterator_t *it)
     assert( it->current_events_buffer->buffer_type == PROFILING_BUFFER_TYPE_EVENTS );
 
     next_off = it->current_events_buffer->next_buffer_file_offset;
+    if(next_off == -1) {
+        it->current_event.native = NULL;
+        it->current_events_buffer = NULL;
+        return NULL;
+    }
+
     return dbp_iterator_set_offset(it, next_off);
```
omor1, line 391: I'm not sure I see why this is needed. `dbp_iterator_set_offset` releases the current event buffer and sets `it->current_events_buffer` and `it->current_event.native` to `NULL`, then attempts to go to buffer with said offset via `refer_events_buffer`. When `offset == -1` then I think that both the `mmap` and `lseek` methods should fail and return `NULL`. Finally, `dbp_iterator_move_to_event` is called; here, if `it->current_events_buffer` is `NULL` (which it should be since `refer_events_buffer` failed) then `it->current_event.native` is set to `NULL` too. The return value is that of `dbp_iterator_current`, which is `NULL` when either of `it->current_events_buffer` or `it->current_event.native` is `NULL`.

The only way I see a codepath causing issues is if `refer_events_buffer` somehow _doesn't_ fail for `offset == -1`, in which case that would be the correct place to add such a check.

This code also doesn't release the current buffer, so you leak that memory.

therault:
This PR has been superseded by https://github.com/ICLDisco/parsec/pull/469

@omor1 to answer your question above
>Does this PR let users have task classes inject extra profiling info into each task?

Yes, this is the goal of this PR, and yes it allows to avoid cluttering the locals with information you want to display. But the target is more towards DSLs than users (although, if users want to write their own profile_info function and overwrite the DSL's, I don't see why not)

PR #430, opened 2022-09-20
Ttg/termdet dynamic PTG

therault:
This PR fixes a couple of issues with the termination detection implementation in the current master (in particular, a typo that would make the tag used for termination detection active messages be also used for data messages (put/get), resulting in catastrophic failures...), it re-enable the possibility for the user to pass custom flags to ptgpp at CMake time, and it enables dynamic termination detection above PTG.

If the user passes `--dynamic-termdet` or `-D` to parsec-ptgpp, or if they define `termdet = "dynamic"` in the PTG file, a dynamic termination detector (fourcounter currently) is selected when instantiating the taskpool, and the code does not pre-count the number of tasks. This results in atomics when scheduling tasks to count the number of discovered tasks.

Tested (in both local / pre-counted termination detection and dynamic termination detection) with pingpong/rtt and a couple of other tests.


parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -3796,16 +3809,29 @@ static void jdf_generate_internal_init(const jdf_t *jdf, const jdf_function_entr
...
             assert((f->user_defines & JDF_HAS_USER_TRIGGERED_TERMDET) ||
-                   (NULL != jdf_property_get_string(jdf->global_properties, JDF_PROP_TERMDET_DYNAMIC, NULL)));
+                   (f->user_defines & JDF_HAS_DYNAMIC_TERMDET));
             /* The startup tasks are going to count the real number of tasks as they discover them.
              * For now, we lock the idleness by creating a runtime pending action, and
              * we use sync_point to find when all the startup tasks are done. */
-            coutput("    __parsec_tp->sync_point = PARSEC_%s_NB_TASK_CLASSES\n;"
-                    "    __parsec_tp->super.super.tdm.module->taskpool_addto_nb_tasks(&__parsec_tp->super.super, 1);\n",
+            coutput("    __parsec_tp->sync_point = PARSEC_%s_NB_TASK_CLASSES;\n"
+                    "    //__parsec_tp->super.super.tdm.module->taskpool_addto_nb_tasks(&__parsec_tp->super.super, 1);\n",
```
devreal, line 3834: Why is this commented out?

devreal (commented):
I can't say much about the code generation part, just have one comment.

bosilca (approved):

PR #438, opened 2022-11-01
Fix: termination detector race condition

therault:
This is a backport of fixes made in TESSEOrg/parsec fork. There is a race condition in the termination detector that tries to free the monitor too early. The solution consists in adding an explicit 'unmonitor taskpool' function in the termination detector API and wait that the taskpool is destroyed / recycled to unmonitor the taskpool. This should help solve the issues in the taspool_wait PR too.

  - Amend cherry-pick to fit master
  - fix conflicts during rebase with master

### DIFF
--- a/parsec/class/parsec_rwlock.c
+++ b/parsec/class/parsec_rwlock.c
@@ -239,8 +239,8 @@ void parsec_atomic_rwlock_rdlock(parsec_atomic_rwlock_t *L)
         parsec_atomic_lock(&L->w);
 #if defined(PARSEC_DEBUG_PARANOID)
     assert(L->writer == 0);
+    assert(L->nbreaders < MAX_READERS_TO_CHECK);
     L->readers[L->nbreaders] = (int)pthread_self();
-    assert(L->nbreaders < MAX_READERS);
 #endif
     L->nbreaders++;
     parsec_atomic_unlock(&L->r);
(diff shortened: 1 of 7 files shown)


parsec/class/parsec_rwlock.h
```diff
@@ -82,9 +82,9 @@ typedef struct parsec_atomic_rwlock_s {
 } parsec_atomic_rwlock_t;
 
 #if defined(PARSEC_DEBUG_PARANOID)
-#define PARSEC_RWLOCK_UNLOCKED { PARSEC_ATOMIC_UNLOCKED, PARSEC_ATOMIC_UNLOCKED, 0, 0 }
+#define PARSEC_RWLOCK_UNLOCKED PARSEC_ATOMIC_UNLOCKED, PARSEC_ATOMIC_UNLOCKED, 0, 0,
```
devreal, line 85: This change seems to unintended. Didn't we just make the change to add the brackets to the `PARSEC_RWLOCK_UNLOCKED`?

therault: Fixed (and applied the same changes to the non-debug-paranoid case)

bosilca (approved):

bosilca:
Please update the commit log to identify the original commit that was cherry-picked.

devreal (approved):

PR #440, opened 2022-11-02
Fixes/tls in comm thread

therault:
Fixes an assert that triggers in debug noisier/paranoid mode after https://github.com/ICLDisco/parsec/pull/422

Also remove a warning about unused variable in scheduling.c

### DIFF
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -286,7 +286,6 @@ __parsec_schedule(parsec_execution_stream_t* es,
                   int32_t distance)
 {
     int ret;
-    parsec_task_t *task = tasks_ring;
 
 #if defined(PARSEC_DEBUG_PARANOID) || defined(PARSEC_DEBUG_NOISIER)
     {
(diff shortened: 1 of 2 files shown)


parsec/parsec.c
```diff
@@ -249,7 +249,7 @@ static void* __parsec_thread_init( __parsec_temporary_thread_initialization_t* s
     }
     gettimeofday(&tv_now, NULL);
 
-    PARSEC_TLS_SET_SPECIFIC(parsec_tls_execution_stream, es);
+    parsec_set_my_execution_stream(es);
```
bosilca, line 252: The only reason we need the `parsec_set_my_execution_stream` accessor is because parsec_tls_execution_stream is defined as static, and therefore not visible outside the parsec.c file. I think that where the symbols are visible, the uses of the TLS macros should be encouraged.

bosilca (approved):

parsec/scheduling.c
```diff
@@ -795,7 +797,7 @@ int parsec_context_wait( parsec_context_t* context )
         return PARSEC_ERR_NOT_SUPPORTED;
     }
 
-    ret = __parsec_context_wait( context->virtual_processes[0]->execution_streams[0] );
+    ret = __parsec_context_wait( parsec_my_execution_stream() );
```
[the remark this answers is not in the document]
therault: Reverted

parsec/parsec.c
therault: Reverted

abouteiller (approved):

devreal (approved):

PR #442, opened 2022-11-02
A small script to help us find which files might need copyright update.

therault:
Usage:
  bash contrib/check-copyright.sh -f '*.[hc]' -o
     lists c or h files from the CWD that have an old copyright
  bash contrib/check-copyright.sh -f '*.[hc]' -m
     lists c or h files from the CWD that have a missing copyright (or malformed one)
  bash contrib/check-copyright.sh -f '*.[hc]' -m -o -v
     lists for each c or h files if they have a copyright or not ('M'),
     and if they have a copyright, will denote them as ok (' '), or Old ('O')
     depending on the date of the last commit message for this file in git
     and the current end year in the copyright line

### DIFF
--- a/contrib/check-copyright.sh
+++ b/contrib/check-copyright.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+COPYRIGHT_LINE_RE="Copyright (c) .* The University of Tennessee"
+SCRIPTNAME="copyright-check.sh"
+
+date_last_git_mod() {
+     git log -1 --pretty="format:%cs" $1 | awk -v FS="-" '{print $1}'
+}
+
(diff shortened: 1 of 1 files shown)


contrib/check-copyright.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+
+COPYRIGHT_LINE_RE="Copyright (c) .* The University of Tennessee"
+SCRIPTNAME="copyright-check.sh"
```
bosilca, line 4: ARGV[0] ?

abouteiller (approved):

bosilca (approved):

PR #443, opened 2022-11-03
Info API

abouteiller:
This is the Info key API

At this point, I'd like to get feedback on the API (the .h) before I go and implement the ops. In particular, the proposed API may be hard to make thread safe.


parsec/utils/info.h
```diff
@@ -0,0 +1,160 @@
...
+} parsec_info_value_t;
+
+
+struct parsec_info_s {
+    parsec_hashtable_t super;
+} parsec_info_t;
+
+PARSEC_OBJ_CLASS(parsec_info_t);
+
+} parsec_info_t;
```
bosilca, line 51: typo.

bosilca:
Looks very basic, but for now it should do. The real issue is that we need to link this with the MCA params, or have to double check everywhere of we have the expected value either in one or the other (or both, with a predefined precedence).

PR #445, opened 2022-11-04
Remove dependency on argv[0]

### LINKED ISSUE #431: Remove the need for `argv[0]` being the app name

abouteiller:
Remove the dependency on ARGV[0] by getting the app name from somewhere else. 

- [x] change the examples to not copy argv0 everywhere now that it is unnecessary. 
- [x] test on OSX
- [x] test on windows WSL
- [ ] test on windows Visual Studio/native

fixes #431

### DIFF
--- a/parsec/parsec.c
+++ b/parsec/parsec.c
@@ -41,6 +41,7 @@
 #include "parsec/utils/mca_param.h"
 #include "parsec/utils/installdirs.h"
 #include "parsec/utils/cmd_line.h"
+#include "parsec/utils/process_name.h"
 #include "parsec/utils/debug.h"
 #include "parsec/utils/parsec_environ.h"
 #include "parsec/utils/mca_param_cmd_line.h"
@@ -361,8 +362,6 @@ static void parsec_vp_init( parsec_vp_t *vp,
     }
 }
 
-#define DEFAULT_APP_NAME "app_name"
-
 #define GET_INT_ARGV(CMD, ARGV, VALUE) \
 do { \
     int __nb_elems = parsec_cmd_line_get_ninsts((CMD), (ARGV)); \
@@ -399,6 +398,7 @@ parsec_context_t* parsec_init( int nb_cores, int* pargc, char** pargv[] )
     int profiling_enabled = 0;
 
     gethostname(parsec_hostname_array, sizeof(parsec_hostname_array));
+    parsec_app_name = parsec_process_name();
 
     PARSEC_PAPI_SDE_INIT();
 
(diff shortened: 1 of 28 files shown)


bosilca (approved):
Nice solutions.

tests/apps/haar_tree/main.c
```diff
@@ -171,20 +171,9 @@ int main(int argc, char *argv[])
             break;
         }
     }
-    pargv = malloc( (pargc+1) * sizeof(char*));
```
therault, line 174: Wait, we still need to find "--" on the command line to pass it to parsec_init: arguments before "--" are not understood by parsec_init and will create an error message.

tests/apps/haar_tree/main.c
```diff
@@ -171,20 +171,9 @@ int main(int argc, char *argv[])
...
-            dashdash = i;
-            pargc = 0;
-        } else if( dashdash != -1 ) {
-            pargc++;
-        }
-    } else {
-        pargv[0] = NULL;
-    }
     parsec = parsec_init(1, &pargc, &pargv);
     free(pargv);
```
therault, line 175: Also, you kept free(pargv), but pargv is not malloc anymore?

therault (changes requested):
I like the `process_name()` solution, but there is one example broken.

The other example fixes are correct, AFAICT

parsec/parsec.c
```diff
@@ -705,7 +701,7 @@ parsec_context_t* parsec_init( int nb_cores, int* pargc, char** pargv[] )
             /* Specialize the profiling filename to avoid collision with other instances */
             ret = asprintf( &cmdline_info, "%s_%d", basename(parsec_app_name), (int)getpid() );
             if (ret < 0) {
-                cmdline_info = strdup(DEFAULT_APP_NAME);
+                cmdline_info = strdup(parsec_app_name);
```
bosilca, line 704: parsec_app_name is already a malloced string. Do you need to strdup it here ?

parsec/utils/process_name.c
```diff
@@ -0,0 +1,29 @@
...
+    }
+#else
+    ret = readlink("/proc/self/exe", name, len);
+    if(-1 == ret) {
+        snprintf(name, len, "parsec-app" );
+        return strdup(name);
+    }
+    name[ret] = '\0';
+#endif
+    sname = rindex(name, '/');
```
bosilca, line 25: The path separator in parsec is `PARSEC_PATH_SEP`

bosilca (changes requested):

parsec/utils/process_name.c
abouteiller: done

tests/apps/haar_tree/main.c
abouteiller: yes look at line 168 above

bosilca (approved):
Here is a [Stackoverflow](https://stackoverflow.com/questions/4570174/how-to-get-the-process-name-in-c#:~:text=You%20can%20obtain%20the%20process,you%20can%20also%20use%20GetProcessImageFileName.) possible Windows solution

abouteiller:
I tried with WSL and it doesn't require that trick. I couldn't compile with Visual Studio (can't find MPI for some reason). Do we really care about something else than WSL at this point?

bosilca:
Let's merge it.

PR #447, opened 2022-11-04
Remove all options that are not prefixed by --mca

abouteiller:
~~--mca becomes --parsec~~, all stray options -c -V etc replaced by an mca --help becomes --parsec-help

The end goal is that we can safely pass argc/argv from the command line without worrying that we are capturing MPI or application parameters.  

- [x] looks like I broke the sched tests somehow.

### DIFF
--- a/parsec/utils/mca_param_cmd_line.c
+++ b/parsec/utils/mca_param_cmd_line.c
@@ -2,7 +2,7 @@
  * Copyright (c) 2004-2005 The Trustees of Indiana University and Indiana
  *                         University Research and Technology
  *                         Corporation.  All rights reserved.
- * Copyright (c) 2004-2005 The University of Tennessee and The University
+ * Copyright (c) 2004-2022 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2004-2005 High Performance Computing Center Stuttgart,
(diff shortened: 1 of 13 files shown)


parsec/utils/mca_param_cmd_line.c
```diff
@@ -41,30 +41,23 @@ static void add_to_env(char **params, char **values, char ***env);
...
 {
     int ret = PARSEC_SUCCESS;
 
-    ret = parsec_cmd_line_make_opt3(cmd, '\0', "mca", "mca", 2,
-                                  "Pass context-specific MCA parameters; they are considered global if --gmca is not used and only one context is specified (arg0 is the parameter name; arg1 is the parameter value)");
-    if (PARSEC_SUCCESS != ret) {
-        return ret;
-    }
-
-    ret = parsec_cmd_line_make_opt3(cmd, '\0', "gmca", "gmca", 2,
```
abouteiller, line 56: Don't see why we need gmca since we are not mpiexec. Let me know if there is a point.

bosilca:
I see the benefit, but I'm really not fan of this.

bosilca:
As discussed on 03/31/23 we decided to remove the `--parsec` addition but we keep everything else.

parsec/utils/mca_param_cmd_line.c
abouteiller: I restored the gmca functionality, it may be useful for the info params, so I'll remove it if it turns out useless then.

bosilca (approved):

PR #448, opened 2022-11-06
GPU task completion delegated to a different thread

josephjohnjj:
The GPU manager thread delegates the task completion to a compute thread. Setting mca parameter `device_cuda_delegate_task_completion = 1` will enable this functionality.

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -45,6 +45,8 @@ parsec_cuda_memory_reserve( parsec_device_cuda_module_t* gpu_device,
 static int parsec_cuda_memory_release( parsec_device_cuda_module_t* gpu_device );
 static int parsec_cuda_flush_lru( parsec_device_module_t *device );
 
+extern int parsec_cuda_delegate_task_completion; 
+
 /* look up how many FMA per cycle in single/double, per cuda MP
  * precision.
  * The following table provides updated values for future archs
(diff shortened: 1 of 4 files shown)


parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -45,6 +45,8 @@ parsec_cuda_memory_reserve( parsec_device_cuda_module_t* gpu_device,
 static int parsec_cuda_memory_release( parsec_device_cuda_module_t* gpu_device );
 static int parsec_cuda_flush_lru( parsec_device_module_t *device );
 
+extern int parsec_cuda_delegate_task_completion;
```
devreal, line 48: Can the `extern` definition be moved to a header?

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -2730,9 +2732,20 @@ parsec_cuda_kernel_scheduler( parsec_execution_stream_t *es,
         gpu_task->ec = NULL;
         goto remove_gpu_task;
     }
+    
     parsec_cuda_kernel_epilog( gpu_device, gpu_task );
-    __parsec_complete_execution( es, gpu_task->ec );
     gpu_device->super.executed_tasks++;
+
+    if( parsec_cuda_delegate_task_completion == 0 )
+        __parsec_complete_execution( es, gpu_task->ec );
```
devreal, line 2740: Please use braces for all blocks

parsec/mca/device/cuda/device_cuda_module.c
josephjohnjj: I have moved. it to parsec/mca/device/cuda/device_cuda_internal.h

devreal (approved):
LGTM but someone with more knowledge on how the device framework works should take a look too.

parsec/scheduling.c
```diff
@@ -487,6 +487,9 @@ int __parsec_task_progress( parsec_execution_stream_t* es,
         if(task->status <= PARSEC_TASK_STATUS_HOOK) {
             rc = __parsec_execute( es, task );
         }
+        else if(task->status == PARSEC_TASK_STATUS_COMPLETE) { // To deal with task completion of the GPU tasks
```
bosilca, line 490: This code is unnecessary because the task returned from the GPU will be in the state `PARSEC_TASK_STATUS_COMPLETE`, which means the test `task->status <= PARSEC_TASK_STATUS_HOOK` will fail and the execution will do in the switch with rc set to `PARSEC_HOOK_RETURN_DONE` (because of the initialization at the beginning of the function), which translates into executing `__parsec_complete_execution()`.

PR #457, opened 2022-11-11
Proper case and imported target for hwloc

abouteiller:

@therault you had changed the case for this because TTG link problems, is this ok with TTG?

### DIFF
--- a/cmake_modules/FindHWLOC.cmake
+++ b/cmake_modules/FindHWLOC.cmake
@@ -1,23 +1,23 @@
-# - Find HWLOC library
+# - Find the hwloc library
 # This module finds an installed library that implements the HWLOC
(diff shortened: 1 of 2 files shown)


bosilca:
@therault should review this, he removed the HWLOC::HWLOC for some reason (that I do not recall).

therault:
MADNESS and TTG use the name 'hwloc' for the target (without namespacing or capitalization). If we define HWLOC::HWLOC, we introduce another target hwloc, and the library is linked twice, which CMake solves, but which creates funny issues if the two findhwloc.cmake find different hwloc libraries.

I don't know what is the 'right' solution, I went with the simplest I found: use the same target name as the other two projects.

abouteiller:
I see, they are doing it wrong but it doesn't help to say that. I'll look if target alias can help with this.

PR #461, opened 2022-11-28
initial flags modification PR. Open to review.

DSMishler:
I understand this is unlikely to be immediately approved, but just wanted to put it on a PR officially before our weekly meeting.

TLDR: add a flag so that task timing can be done more accurately. The results? It turns out it does make a little bit of a difference.

### DIFF
--- a/parsec/mca/pins/task_profiler/pins_task_profiler_module.c
+++ b/parsec/mca/pins/task_profiler/pins_task_profiler_module.c
@@ -252,9 +252,10 @@ task_profiler_exec_count_end(struct parsec_execution_stream_s*   es,
 {
     if (NULL != task->taskpool->profiling_array &&
         task->task_class->task_class_id < task->taskpool->nb_task_classes)
-        PARSEC_TASK_PROF_TRACE(es->es_profile,
-                               task->taskpool->profiling_array[2*task->task_class->task_class_id+1],
-                               task);
+        PARSEC_TASK_PROF_TRACE_FLAGS(es->es_profile,
+                                     task->taskpool->profiling_array[2*task->task_class->task_class_id+1],
+                                     task,
+                                     PARSEC_PROFILING_EVENT_TIME_AT_START);
     (void)cb_data;
 }
 
--- a/parsec/mca/pins/task_profiler/pins_task_profiler_module.c
+++ b/parsec/mca/pins/task_profiler/pins_task_profiler_module.c
@@ -237,11 +237,10 @@ task_profiler_exec_count_begin(struct parsec_execution_stream_s*   es,
 {
     if (NULL != task->taskpool->profiling_array &&
         task->task_class->task_class_id < task->taskpool->nb_task_classes)
-        PARSEC_PROFILING_TRACE(es->es_profile,
-                               task->taskpool->profiling_array[2*task->task_class->task_class_id],
-                               task->task_class->key_functions->key_hash(
-                                       task->task_class->make_key( task->taskpool, task->locals ), NULL),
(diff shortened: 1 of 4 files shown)


parsec/profiling.h
```diff
@@ -85,26 +85,34 @@
...
-#define PARSEC_PROFILING_EVENT_COUNTER      (1<<2)
+#define PARSEC_PROFILING_EVENT_COUNTER          (1<<2)
 /**
  * @brief Constant to use when no handle/object ID can be associated
  *        with an event
  * @details Some functions take a handle/object ID as a parameter.
  *          When such ID is not applicable, this constant should be
  *          passed to the parameter.
  */
+#define PARSEC_PROFILING_EVENT_TIME_AT_START    (1<<3)
```
therault, line 108: The comment goes before the constant definition. Doxygen will match the wrong comment.

therault, line 108: You should also define a PARSEC_PROFILING_EVENT_TIME_AT_END as 0 and use it symmetrically in the pins profiler that you modify, just for the clarity of reading.

parsec/mca/pins/task_profiler/pins_task_profiler_module.c
```diff
@@ -252,9 +252,10 @@ task_profiler_exec_count_end(struct parsec_execution_stream_s*   es,
 {
     if (NULL != task->taskpool->profiling_array &&
         task->task_class->task_class_id < task->taskpool->nb_task_classes)
-        PARSEC_TASK_PROF_TRACE(es->es_profile,
-                               task->taskpool->profiling_array[2*task->task_class->task_class_id+1],
-                               task);
+        PARSEC_TASK_PROF_TRACE_FLAGS(es->es_profile,
```
therault, line 255: This reindentation is incorrect and should not be part of the PR

parsec/parsec_internal.h
```diff
@@ -570,9 +570,19 @@ extern int device_delegate_begin, device_delegate_end;
...
+                           (TASK)->task_class->key_functions->          \
+                           key_hash((TASK)->task_class->make_key(       \
+                              (TASK)->taskpool, (TASK)->locals ), NULL), \
+                           (TASK)->taskpool->taskpool_id,               \
+                           (void*)&(TASK)->prof_info,                   \
+                           (FLAGS)); 
 #else
 #define PARSEC_TASK_PROF_TRACE(PROFILE, KEY, TASK)
 #define PARSEC_TASK_PROF_TRACE_IF(COND, PROFILE, KEY, TASK)
+#define PARSEC_TASK_PROF_TRACE_FLAGS(PROFILE, KEY, TASK)
```
therault, line 585: Does not follow the prototype of the other case (#if defined(PARSEC_PROF_TRACE))

therault (changes requested):
See comments for list of changes

parsec/mca/pins/task_profiler/pins_task_profiler_module.c
DSMishler: I'm not sure if you meant re-indenting for `task` or reindenting everything according to the beginning of the arguments. My followup will assume you meant the latter.

therault (commented):
Changes look good, but the CI checks are failing because of a Spack update.

Omri Mor suggests to try removing `concretization: together` and adding
```
concretizer:
   unify: true
```

to the file [.github/CI/github_runner.yaml](https://github.com/ICLDisco/parsec/tree/master/.github/CI/github_runner.yaml)

Can you try this?

DSMishler:
Should be all set to go now!

PR #462, opened 2022-12-06
Topic/update spack

bosilca:


therault (approved):

bosilca:
Thanks @omor1 for the tip on how to fix this. We just need to rebase all the pending PR and we should be good to go.

PR #468, opened 2023-01-05
Comm engine register unregister

therault:
This defines a set of helper functions to register / unregister callbacks that are called when a communication engine is up or when it is teared down.

Currently, the termination detectors that use active messages do not work since the lazy initialization of the communication engine. If the user tears down a comm engine, and creates a new one, non-predefined tags are also not re-registered.

This provides a mechanism so that the different levels of software are notified that this happens, and can react by de-registering their tags and re-registering them.

It is one of the PRs that we need to consider for porting TTG over parsec master.


abouteiller (approved):

therault:
I converted it to draft to avoid merging by mistake. AFAIK, George is working on a better way to handle the issue

bosilca:
Replaced by #521

PR #450, opened 2022-11-10
prefix all parsec tests with parsec/

abouteiller:
Signed-off-by: Aurelien Bouteiller <bouteill@icl.utk.edu>

PR #478, opened 2023-01-19
Do not enable testing if PaRSEC is not the top project.

bosilca:
Alternative to #450 for disabling the tests if PaRSEC is not the top project, but it does not prepend parsec to all tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,11 +66,6 @@ set(CMAKE_NO_SYSTEM_FROM_IMPORTED True)
 # completely disregarding the versions installed by Mac Ports or Brew.
 set(CMAKE_FIND_FRAMEWORK LAST)
 
-# CTest system
-SET(DART_TESTING_TIMEOUT 120)
-enable_testing()
-include(CTest)
-
 #####
 # ccmake tunable parameters
 #####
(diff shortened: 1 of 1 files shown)


abouteiller:
I don't like this because it makes it impossible to run ctest for a parsec submodule at all. 

Also I think it does nothing because of https://cmake.org/cmake/help/latest/command/enable_testing.html
"This command is automatically invoked when the CTest module is included, except if the BUILD_TESTING option is turned off."
which is done the line below

bosilca:
Yes, I'm reading the [CTest](https://cmake.org/cmake/help/latest/module/CTest.html) documentation and there is no clean way to disable testing (aka. you should not preset the BUILD_TESTING) except maybe using presets fixtures, but that is impossible here.

bosilca:
I found other projects that have the same issue. Apparently the only way to prevent tests when build as a subproject is to only add tests if a subproject variable for testing is set (such as PARSEC_BUILD_TESTING) and only set it when the project is top-level.

bosilca:
After discussion we decided that all the solutions put forward here are not suitable for our initial goal of running only partial tests when included as a subproject. We also concluded there is no good solution, so we will leave the project as is, all tests will be run by ctest.

PR #479, opened 2023-01-27
Fix hash function generation for PTG

therault:
To auto-generate the hash function of task classes in PTG, we need to compute the range of values for a parameter, but this is not sufficient: we also need the minimum value of that parameter, so the hash function remains reversible.

In a previous commit, we used the assumption that the leftmost expression of the range is always the minimum value, but with local defitions, this is not always the case: in
  m = [im = 0 .. M] %{ return f(im); %}
if f is not monotonic, f(0) is not necessarily the min value for m.

This commit reverts the change that was 'saving' the storage of the min value. Note that the min value was computed correctly anyway during the internal init phase, so this was only saving memory.

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -1554,6 +1554,7 @@ static void jdf_generate_structure(jdf_t *jdf)
     for(f = jdf->functions; f != NULL; f = f->next) {
         if( 0 == (f->user_defines & JDF_FUNCTION_HAS_UD_MAKE_KEY) ) {
             for(pl = f->parameters; pl != NULL; pl = pl->next) {
+                coutput("  int %s_%s_min;\n", f->fname, pl->name);
                 coutput("  int %s_%s_range;\n", f->fname, pl->name);
             }
         } else {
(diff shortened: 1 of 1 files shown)


parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -3344,30 +3345,8 @@ static  void jdf_generate_deps_key_functions(const jdf_t *jdf, const jdf_functio
...
-                    have_min = 1;
-                } else {
-                    if( vl->expr->local_variables != NULL ) {
-                        char *vname;
-                        if( asprintf(&vname, "%s%s_min", JDF2C_NAMESPACE, vl->name) <= 0 ) {
-                            fprintf(stderr, "Cannot allocate internal memory for the PTG compiler\n");
-                            exit(-1);
-                        }
-                        coutput("  int %s;\n", vname);
-                        jdf_generate_range_min_without_fn(jdf, vl->expr, vname, "(&"JDF2C_NAMESPACE"_assignments)");
```
bosilca, line 3359: Is this function (`jdf_generate_range_min_without_fn`) still unecessary ?

therault: `jdf_generate_range_(min/max)_without_fn` are used by `jdf_generate_range_(min/max)`, which in turn are used by `jdf_generate_expression`. Those are useful for ranges in expressions that come from something else than a parameter, for which we don't have pre-computed min/max values...

They 'work' under the assumption that the leftmost value of the range is the minimum and the rightmost is the maximum, which is clearly an assumption that is not true anymore.

We can keep the old (unused, AFAICT) behavior, or remove them and set NULL as the minimum function and maximum function for range expression (so if we invoke dynamically, this will segfault/bus error), or we can replace them by real functions but then the only reliable way to find the minimum and maximum values are to iterate over the entire range and find the min/max... Which can be very costly.

As far as I can tell, we never dynamically invoke a range expression function.

bosilca (approved):

abouteiller (approved):

PR #481, opened 2023-02-01
remote_dep: rotate bcast topology around root

omor1:
The communication topology always prefers to send to lower-ranked nodes over higher-ranked nodes. This results in a significant communication imbalance regardless of how a computation is distributed, unless all communication is point-to-point.

This introduces a pair of functions, remote_dep_rank_to_bit and remote_dep_bit_to_rank, to translate between rank id 0 .. np-1 and the bank offset and bit index for the rank_bits array. This translation "rotates" the ranks around the root of a communication i.e. the source is always bank 0/bit 0 and there are different topologies depending on the root of a bcast. If an application's communications are otherwise balanced then this results in a better communication balance.

### DIFF
--- a/parsec/interfaces/dtd/insert_function.c
+++ b/parsec/interfaces/dtd/insert_function.c
@@ -1569,13 +1569,14 @@ dtd_release_dep_fct(parsec_execution_stream_t *es,
             if( parsec_dtd_not_sent_to_rank((parsec_dtd_task_t *)oldcontext,
                                             dep->belongs_to->flow_index, dst_rank)) {
                 struct remote_dep_output_param_s *output;
-                int _array_pos, _array_mask;
+                int _array_pos, _array_bit;
+                uint32_t _array_mask;
 
 #if !defined(PARSEC_DIST_COLLECTIVES)
                 assert(src_rank == es->virtual_process->parsec_context->my_rank);
 #endif
-                _array_pos = dst_rank / (int)(8 * sizeof(uint32_t));
-                _array_mask = 1 << (dst_rank % (8 * sizeof(uint32_t)));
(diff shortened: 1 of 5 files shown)


omor1:
I had a prior version where `remote_dep_rank_to_bit` and `remote_dep_bit_to_rank` took an explicit `nb_nodes` argument. I can restore that if that's preferred over using `parsec_remote_dep_context.max_nodes_number`.

omor1:
Most real applications are unlikely to see much improvement from this, as they aren't limited by network bandwidth or latency anyway. A test application I created based off a ping-pong that places a task blocking forward progress on node `np-1` does see significant improvement, since that task now doesn't _always_ have the worst latency for activations/data movement. The exact improvement is dependent on what bcast topology is used—star should see little-to-no improvement, while binomial should see some and chain the greatest, since chain has the highest bcast latency.

omor1:
This leaves open an option to, in the future, rotate the bcast around the rank with the highest-priority descendant instead of the root. This should ensure that the highest-priority task will always have the lowest latency for receiving the activation & data it needs. A complicating factor is that determining the highest task priority is concurrent with adding ranks to the `rank_bits` array: see `parsec_release_dep_fct`, `dtd_release_dep_fct`, and `parsec_gather_collective_pattern`.

You either need to split these operations in two and iterate descendants twice, once to figure out the priorities and the bcast pivot, and again to actually add ranks to the `rank_bits` array; or you need to rotate the `rank_bits` array afterwards.

An alternative is to store rank bits as their true rank and to start iterating over the array at the pivot instead of rank 0, but that is nontrivial to implement correctly; I found that it was much simpler to instead store ranks rotated in the bit array.

bosilca (approved):

PR #488, opened 2023-02-13
Hotfix dtd dsl and warning

therault:
Two fixes for DTD:
    
       - there was a spurious warning when registering multiple chores with
         the same DTD task class, because we changed when the task class
         is inserted (and we were trying to insert it twice)
    
       - Between the various fixes for the PARSEC_TASK_PROF_TRACE_FLAGS and
         PARSEC_TASK_PROF_TRACE macros, we stopped invoking the right tracing
         function for DSLs that do not support parameters on the locals
         array.


bosilca (approved):

DSMishler:
Afraid that the issue persists.

Also having some trouble building on xsdk since the merge, but on saturn the issue persists.

PR #489, opened 2023-02-14
Use pip to build and install the python support.

bosilca:
The old approach was deprecated, this moves us to a more modern and more clean instllation process. It did require few changes:
- use pip to build and install a temporary module such that we can use it for testing purposes.
- reinstall the module again once we are in the install path.

This PR does not fix the profiling tests, it just allows them to run correctly.


therault (approved):
Looks good to me. Tested on my setup, it works.

PR #491, opened 2023-02-15
Fix an issue that wrong name is used in hook of RECURSIVE

QingleiCao:


bosilca:
Device names should always be upper cases. @abouteiller why was this changed to lower case here ?

bosilca:
I don't think we have a valid reason to mix lower case and upper case on the device name. It should be consistent. Please check https://github.com/ICLDisco/parsec/pull/493.

PR #492, opened 2023-02-15
Make sure data_in is not NULL on GPU when accessing nb_elts

QingleiCao:
Some flows may be skipped for certain conditions. So, it needs to make sure the data_in is not NULL on GPU when accessing nb_elts.

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -6827,6 +6827,8 @@ static void jdf_generate_code_hook_gpu(const jdf_t *jdf,
             }
             coutput("  gpu_task->flow_nb_elts[%d] = 0;\n", di);
         }else{
+            coutput("  // A shortcut to check if the flow exists\n");
+            coutput("  if (gpu_task->ec->data[%d].data_in != NULL) {\n", di);
             if(size_property == NULL){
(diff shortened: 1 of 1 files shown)


bosilca:
Why would the data_in be NULL ? Scratch data or WRITE only data ?

QingleiCao:
> Why would the data_in be NULL ? Scratch data or WRITE only data ?

For instance, in TLR, only the dense kernels are on GPU now, so the flow about rank (Ar/Br/Cr) is conditional [(codes).](https://github.com/bosilca/hicma-x-dev/blob/0e2defc9425c502da8b10601ad28c7ce49fbe569/hicma_parsec/HiCMA_dpotrf_L_2flow.jdf#L265) I'm not sure what has changed since last summer, but this is what I found that causes the issue.

bosilca:
We talked about this PR on 02/28/23 and concluded that it provides a single benefit to allow data-dependent NULL input flows, at least in some particular cases. The idea is interesting but if we move to allow data-dependent PTG (aka PTG that has well defined flows that can disappear based on the data content) we need to do it in a portable and generic way across the entire generated code.

Meanwhile, where this was needed, we can replace a NULL (that signals a full rank tile) with a data_t that provide a full-rank value.

therault (approved):
The change could be improved by adding a comment at the 'if' output, stating that this is a shortcut to check if the flow exists instead of dumping the expression that conditions the existence of said flow. Otherwise it looks good to me.

bosilca (approved):

PR #499, opened 2023-02-28
Make parsec_data_t::device_copies a flexible array member

devreal:
This prevents developers from allocating parsec_data_t statically.


bosilca (approved):
We are requiring C99 (for the atomic support) so this notation for flexible array is OK.

PR #508, opened 2023-03-29
Handle the tertiary case for startup tasks.

bosilca:
Stop the code generation if we reach a tertiary where both sides depend on predecessor tasks.


therault (approved):
I remember it was fine

abouteiller (approved):

PR #509, opened 2023-04-02
Delegate GPU task completion to a co-manager

josephjohnjj:
Delegate GPU task completion to a co-manager using the MCA parameter _device_cuda_delegate_task_completion_.

1. The second CPU thread that submits the task to the GPU device is transitioned to a co-manager. 
2. The task is completed by the manager if the co manager has not yet been set. 
3. The manager pushes the task to be completed to a co-manager specific queue. 
4. The GPU task is freed by the thread (manager or co-manager) that completes it.

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_component.c
+++ b/parsec/mca/device/cuda/device_cuda_component.c
@@ -41,6 +41,7 @@ char* parsec_cuda_lib_path = NULL;
 
 static int cuda_mask, cuda_nvlink_mask;
 
+int parsec_cuda_delegate_task_completion = 0;
 
 /*
  * Instantiate the public struct with all of our public information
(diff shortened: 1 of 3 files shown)


devreal:
What is the status of this? Any reason for not taking this in? 

@josephjohnjj Could you please rebase your branch?

josephjohnjj:
@devreal There was no performance improvement when using a co-manager to _complete_ the task. @bosilca suggested that this might be due to a single task completion not generating enough child tasks to make a noticeable impact. Unlike #566,  in #509 co-manager just _completed_ the tasks and was not involved in task execution. 

I'm doubtful that rebasing the code would be helpful this at this stage. In my codebase, all task offloading to GPU occurs in `parsec_cuda_kernel_scheduler()` within `parsec/mca/device/cuda/device_cuda_module.c`.

In the current codebase, this has been moved  to `parsec_device_kernel_scheduler()` in `parsec/parsec/mca/device/cuda/device_cuda_module.c`.

If I can implement the same in the current codebase if having a co-manger will be helpful.

bosilca:
#762 provides a different approach but on today's codebase.

PR #510, opened 2023-04-03
Gpu workspace fix

therault:
See two commits:
  - issue an error when the user code requires more than the limit in number of workspace per stream, instead of checking only in debug mode
  - garbage collect workspace memory when we free the GPU memory

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -809,6 +809,7 @@ parsec_cuda_flush_lru( parsec_device_module_t *device )
     /* Free all memory on GPU */
     parsec_cuda_memory_release_list(cuda_device, &gpu_device->gpu_mem_lru);
     parsec_cuda_memory_release_list(cuda_device, &gpu_device->gpu_mem_owned_lru);
+    parsec_gpu_free_workspace(gpu_device);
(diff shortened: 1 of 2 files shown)


bosilca (approved):

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -809,6 +809,7 @@ parsec_cuda_flush_lru( parsec_device_module_t *device )
     /* Free all memory on GPU */
     parsec_cuda_memory_release_list(cuda_device, &gpu_device->gpu_mem_lru);
     parsec_cuda_memory_release_list(cuda_device, &gpu_device->gpu_mem_owned_lru);
+    parsec_gpu_free_workspace(gpu_device);
```
QingleiCao, line 812: I see this is the calling trace to here: `parsec_fini` -> `parsec_mca_device_fini` -> `device_cuda_component_close` -> `parsec_cuda_module_fini` -> `parsec_cuda_memory_release` -> `parsec_cuda_flush_lru`. So, it's not the user's responsibility to free the workspace, right? 

Also, is this workspace reusable? For instance, DPOTRFs are called multiple times. 

BTW,  is there any documentation about how to use the workspace? (Just try to understand how to use it)

PR #517, opened 2023-04-10
Gpu fix load balancing

therault:
Current load balancing of devices is broken.

Main issue: when a task uses affinity or last writer as a heuristic to select the device, it does not count the task in the load, but it does remove it from the load when the task completes. This makes the parsec_device_load array become negative for some devices, and then the CPU is wrongly selected to schedule tasks...

Simplest fix: move the `+=` instruction outside of the `if( best_device <= 1)` block.

This PR does multiple other fixes, open for comment:

  - only way to guarantee that the load will remain >= 0 is to atomically update the load... This adds two atomics (in the best case) in the critical path... An alternative approach would be to have only the GPU manager update the load for itself, but then the load would only be accounted for when the task goes from the shared list to a private list, so wrong decisions could live longer... I think the additional atomic is a better trade-off
  - I found confusing that the load is accounted for in `get_best_device` for the `+` and in `cuda_kernel_scheduler` for the `-`. I think it's easier to read to do all the accounting in `cuda_kernel_scheduler`. A user / DSL that would bypass `get_best_device` to decide where to schedule would not break the accounting if we do both end in the kernel scheduler function.
  - Currently, we never use the weight arrays (or more correctly, we only use the `parsec_device_sweight` array for all purposes). The proposed patch would allow the DSL/user to decide which weight array to use. I added a property in jdf2c to define which array to use (defaulting to single precision).

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -40,6 +40,7 @@ static const char *jdf_cfilename;
 /* Optional declarations of local functions */
 static int jdf_expr_depends_on_symbol(const char *varname, const jdf_expr_t *expr);
 static void jdf_generate_code_hooks(const jdf_t *jdf, const jdf_function_entry_t *f, const char *fname);
+static void jdf_generate_code_time_estimate(const jdf_t *jdf, const jdf_function_entry_t *f, char *name);
 static void jdf_generate_code_data_lookup(const jdf_t *jdf, const jdf_function_entry_t *f, const char *fname);
 static void jdf_generate_code_release_deps(const jdf_t *jdf, const jdf_function_entry_t *f, const char *fname);
 static void jdf_generate_code_iterate_successors_or_predecessors(const jdf_t *jdf, const jdf_function_entry_t *f,
@@ -4299,6 +4300,9 @@ static void jdf_generate_one_function( const jdf_t *jdf, jdf_function_entry_t *f
     jdf_generate_code_data_lookup(jdf, f, prefix);
     string_arena_add_string(sa, "  .prepare_output = (parsec_hook_t*)%s,\n", "NULL");
     string_arena_add_string(sa, "  .prepare_input = (parsec_hook_t*)%s,\n", prefix);
+    sprintf(prefix, "time_estimate_of_%s_%s", jdf_basename, f->fname);
+    jdf_generate_code_time_estimate(jdf, f, prefix);
+    string_arena_add_string(sa, "  .time_estimate = %s,\n", prefix);
     sprintf(prefix, "datatype_lookup_of_%s_%s", jdf_basename, f->fname);
     jdf_generate_code_datatype_lookup(jdf, f, prefix);
     string_arena_add_string(sa, "  .get_datatype = (parsec_datatype_lookup_t*)%s,\n", prefix);
@@ -6394,6 +6398,24 @@ jdf_generate_code_datatype_lookup(const jdf_t *jdf,
 }
 
 static void
+jdf_generate_code_time_estimate(const jdf_t *jdf,
+                                const jdf_function_entry_t *f,
+                                char *name) 
+{
+    jdf_def_list_t *prop = NULL;
+
+    (void)jdf;
+
+    jdf_find_property(f->properties, "time_estimate", &prop);
+    if(NULL != prop) {
+        sprintf(name, "%s", prop->expr->jdf_var);
+        return;
+    }
+
+    sprintf(name, "NULL");
+}
+
+static void
 jdf_generate_code_data_lookup(const jdf_t *jdf,
                               const jdf_function_entry_t *f,
                               const char *name)
@@ -6529,12 +6551,11 @@ static void jdf_generate_code_hook_gpu(const jdf_t *jdf,
     jdf_def_list_t *stage_out_property;
     jdf_def_list_t *size_property;
     jdf_def_list_t *desc_property;
-    jdf_def_list_t *weight_property;
     jdf_def_list_t *device_property;
+    jdf_def_list_t *prop;
     const char *dyld;
     const char *dyldtype;
     const char *device;
-    const char *weight;
     string_arena_t *sa, *sa2, *sa3;
     assignment_info_t ai;
     init_from_data_info_t ai2;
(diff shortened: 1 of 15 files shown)


parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -6753,7 +6756,23 @@ static void jdf_generate_code_hook_gpu(const jdf_t *jdf,
     } else {
         weight = "1.";
     }
-    coutput("  ratio = %s;\n", weight);
+    coutput("  __ratio = %s;\n", weight);
+
+    jdf_find_property( body->properties, "optype", &optype_property );
```
abouteiller, line 6761: Recommend renaming this as 'precision=double',

What about 'precision=double complex'? 

Also what is triple precision?

therault:
We had a discussion about this PR during the meeting. We should
  - use an integer representation for the load. int64_t. This should represent the number of nanoseconds the user estimates tasks scheduled on this GPU should take
  - the user should not provide a simple 'weight' expression, but 'weight' should be a user-defined function that computes an estimate of the duration (in nanoseconds) of the kernel on a given GPU. Which GPU should be passed as a parameter to this function (as well as the locals and other globals of the taskpool). We can provide convenience functions to get the hweight, sweight, dweight and t(ensor)weight, but the user ultimately may have to find all the information they need either dynamically or during the wrapper phase
  - use of atomics to update the load is considered costly but a lesser inconvenient than delaying the load estimate update, or than working on an incorrect load estimate.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
abouteiller: This comment will not apply when we implement plan below.

bosilca (approved):

parsec/mca/device/device.c
```diff
@@ -120,11 +128,12 @@ int parsec_get_best_device( parsec_task_t* this_task, double ratio )
     /* 0 is CPU, and 1 is recursive device */
     if( dev_index <= 1 ) {  /* This is the first time we see this data for a GPU, let's decide which GPU will work on it. */
         int best_index;
-        float weight, best_weight = parsec_device_load[0] + ratio * parsec_device_sweight[0];
+        parsec_device_module_t *dev = parsec_mca_device_get(0);
+        uint64_t duration, best_duration = time_estimate(this_task, dev);
```
abouteiller, line 132: This is wrong, this will compute the shortest execution time for the task, not the best load balancing. Case in point device->device_load is only incremented/decremented, never read to consider if the device is overloaded. In most cases all GPU device will return the same time_estimate for the same task (since they have the same gflop/s capacity). 


Correct formula should consider `device->load+time_estimate(task,device))`

parsec/mca/device/device.c
```diff
@@ -147,9 +156,10 @@ int parsec_get_best_device( parsec_task_t* this_task, double ratio )
          * other options to have some load balance */
         if( -1 != prefer_index ) {
             best_index = prefer_index;
-            /* we still prefer this device, until it is twice as loaded as the
+            /* we still prefer this device, until it is load_balance_skew as loaded as the
              * real best load balance device */
-            best_weight = load_balance_skew * (parsec_device_load[prefer_index] + ratio * parsec_device_sweight[prefer_index]);
+            dev = parsec_mca_device_get(best_index);
```
abouteiller, line 161: Same thing here, the skew should be on the resulting load balancing (that is, `skew*(device->load+time_estimate(task,device))`).

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -2552,6 +2559,8 @@ parsec_cuda_kernel_scheduler( parsec_execution_stream_t *es,
     gpu_device = (parsec_device_gpu_module_t*)parsec_mca_device_get(which_gpu);
     cuda_device = (parsec_device_cuda_module_t *)gpu_device;
 
+    parsec_atomic_fetch_add_int64(&gpu_device->super.device_load, gpu_task->load);
```
abouteiller, line 2562: We never update the load on CPU devices, which means that we never load balance. Looks like the breaking change predates this PR, but it is unclear how we are supposed to achieve load balance between CPU and GPUs now.

abouteiller (changes requested):

tests/dsl/ptg/cuda/get_best_device_check.jdf
```diff
@@ -121,6 +123,14 @@ END
 
 extern "C" %{
 
+static int64_t task_time_estimate(const parsec_task_t *task, parsec_device_module_t *dev)
```
QingleiCao, line 126: Then this is needed for every task class, right?

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -462,10 +462,16 @@ parsec_cuda_module_init( int dev_id, parsec_device_module_t** module )
...
     }
-    device->device_hweight = (float)streaming_multiprocessor * (float)hrate * (float)clockRate * 2e-3f;
-    device->device_tweight = (float)streaming_multiprocessor * (float)trate * (float)clockRate * 2e-3f;
-    device->device_sweight = (float)streaming_multiprocessor * (float)srate * (float)clockRate * 2e-3f;
-    device->device_dweight = (float)streaming_multiprocessor * (float)drate * (float)clockRate * 2e-3f;
+    /* We compute gflops based on FMA rate */
+    device->gflops_float16 = 2 * (hrate * streaming_multiprocessor) * freqHz / 1e9;
+    device->gflops_tensor32 = 2 * (trate * streaming_multiprocessor) * freqHz / 1e9;
+    device->gflops_float32 = 2 * (srate * streaming_multiprocessor) * freqHz / 1e9;
+    device->gflops_float64 = 2 * (drate * streaming_multiprocessor) * freqHz / 1e9;
```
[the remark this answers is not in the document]
abouteiller: We have decided to remove all of this and just provide the hardware architecture features. Users can then compute themselves this formula in their own time_estimate function.

parsec/mca/device/cuda/device_cuda_module.c
abouteiller: Our current thinking is that this does not cover  an important use-case anymore, so we are willing to simplify the code by dropping this accounting.

parsec/mca/device/device.c
abouteiller: problem has been resolved

parsec/mca/device/cuda/device_cuda_module.c
abouteiller: See change in device.c:132 that disables get_best_device picking CPU chores when a GPU chore is available.

PR #518, opened 2023-04-11
Can't use NULL to pass-in MPI datatypes with MPICH derivatives

abouteiller:
Using NULL would cause invalid cast errors when using MPICH

### DIFF
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2009-2023 The University of Tennessee and The University
+ *                         of Tennessee Research Foundation.  All rights
+ *                         reserved.
+ */
+
 #include "parsec/parsec_config.h"
 
 #include <mpi.h>
@@ -1676,12 +1682,14 @@ remote_dep_mpi_put_start(parsec_execution_stream_t* es,
                                    -1,
                                    &source_memory_handle, &source_memory_handle_size);
         } else {
-            /* TODO: Implement converter to pack and unpack */
-            int dtt_size;
-            parsec_type_size(dtt, &dtt_size);
+            /* TODO: Implement converter to pack and unpack
+             * register the whole region including the holes because we don't support sparse
+             * registration. */
+            ptrdiff_t extent, lb;
+            parsec_type_extent(dtt, &lb, &extent); (void)lb;
             parsec_ce.mem_register(dataptr, PARSEC_MEM_TYPE_CONTIGUOUS,
-                                   -1, NULL, // TODO JS: this interface is so broken, fix it!
-                                   dtt_size, // TODO JS: what about nbdtt? Is it ok to ignore it?!
+                                   -1, parsec_datatype_uint8_t,
+                                   nbdtt * extent,
                                    &source_memory_handle, &source_memory_handle_size);
 
         }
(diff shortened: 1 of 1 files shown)


devreal (approved):

bosilca (commented):
If the datatype is not contiguous we will either need to pack/unpack or we have the register the entire extent of the datatype. Here is a patch for that:
```c
diff --git a/parsec/remote_dep_mpi.c b/parsec/remote_dep_mpi.c
index 31d5cfd25..c0aa7cc05 100644
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -1673,11 +1673,11 @@ remote_dep_mpi_put_start(parsec_execution_stream_t* es,
                                    &source_memory_handle, &source_memory_handle_size);
         } else {
             /* TODO: Implement converter to pack and unpack */
-            int dtt_size;
-            parsec_type_size(dtt, &dtt_size);
+            ptrdiff_t extent, lb;
+            parsec_type_extent(dtt, &lb, &extent); (void)lb;
             parsec_ce.mem_register(dataptr, PARSEC_MEM_TYPE_CONTIGUOUS,
-                                   -1, NULL, // TODO JS: this interface is so broken, fix it!
-                                   dtt_size, // TODO JS: what about nbdtt? Is it ok to ignore it?!
+                                   -1, parsec_datatype_uint8_t,
+                                   nbdtt * extent,
                                    &source_memory_handle, &source_memory_handle_size);
 
         }
@@ -2018,11 +2018,11 @@ static void remote_dep_mpi_get_start(parsec_execution_stream_t* es,
                                    &receiver_memory_handle, &receiver_memory_handle_size);
         } else {
             /* TODO: Implement converter to pack and unpack */
-            int dtt_size;
-            parsec_type_size(dtt, &dtt_size);
+            ptrdiff_t extent, lb;
+            parsec_type_extent(dtt, &lb, &extent); (void)lb;
             parsec_ce.mem_register(PARSEC_DATA_COPY_GET_PTR(deps->output[k].data.data), PARSEC_MEM_TYPE_CONTIGUOUS,
-                                   -1, NULL,
-                                   dtt_size,
+                                   -1, parsec_datatype_uint8_t,
+                                   nbdtt * extent,
                                    &receiver_memory_handle, &receiver_memory_handle_size);
 
         }
```

parsec/remote_dep_mpi.c
```diff
@@ -1676,12 +1682,14 @@ remote_dep_mpi_put_start(parsec_execution_stream_t* es,
...
+            /* TODO: Implement converter to pack and unpack
+             * register the whole region including the holes because we don't support sparse
+             * registration. */
+            ptrdiff_t extent, lb;
+            parsec_type_extent(dtt, &lb, &extent); (void)lb;
             parsec_ce.mem_register(dataptr, PARSEC_MEM_TYPE_CONTIGUOUS,
-                                   -1, NULL, // TODO JS: this interface is so broken, fix it!
-                                   dtt_size, // TODO JS: what about nbdtt? Is it ok to ignore it?!
+                                   -1, parsec_datatype_uint8_t,
+                                   nbdtt * extent,
```
abouteiller, line 1692: I have implemented your solution @bosilca, but I can't decide if that should be `nbdtt * (extent-lb)`

bosilca: Definitely not `nbdtt * (extent-lb)`. Visualize each data in term of the true lb and true ub, which are the first and last byte used. If there are gaps in the middle, they are part of the true extent. Technically the span of memory you want to cover start after the initial gap (aka the first byte if the first instance of an element), and should go until the last byte (the true ub) of the last instance of the element. But this would require to move the initial pointer by the first gap (true lb of the first element), and this complexify the pointer manipulation. `nbdtt * extent` would cover more memory for all well-formed datatypes.

PR #519, opened 2023-04-11
Do not cancel persistent requests with cray-mpich, it is broken.

abouteiller:
Calling Cancel on request crashes with Cray MPICH derivative. Do not call it if we detect crayPICH.

### DIFF
--- a/parsec/parsec_mpi_funnelled.c
+++ b/parsec/parsec_mpi_funnelled.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2018 The University of Tennessee and The University
+ * Copyright (c) 2009-2023 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  */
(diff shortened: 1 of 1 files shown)


devreal (commented):
Would a better solution to cancel be to send an empty message to complete the receive? I worry that this might break if Cray MPICH barks at active requests in finalize...

abouteiller:
> Would a better solution to cancel be to send an empty message to complete the receive? I worry that this might break if Cray MPICH barks at active requests in finalize...

It would but it's more work and here we are just fixing it 'enough' so that it doesn't crash on us when it's their stuff that is broken.

parsec/parsec_mpi_funnelled.c
```diff
@@ -726,8 +726,11 @@ mpi_no_thread_tag_unregister(parsec_ce_tag_t tag)
     MPI_Status status;
 
     for(i = tag_struct->start_idx; i < tag_struct->start_idx + EACH_STATIC_REQ_RANGE; i++) {
+#if !defined(CRAY_MPICH_VERSION)
+        // MPI Cancel broken on Cray
         MPI_Cancel(&array_of_requests[i]);
         MPI_Test(&array_of_requests[i], &flag, &status);
+#endif
```
bosilca, line 733: So in their world it is legitimate to free a pending request but not to cancel it ? Is this even compliant with the MPI standard ?

abouteiller: Yes and no. Yes, it is valid to FREE a request that is pending without first cancelling it. No it is not valid to enter FINALIZE with a freed request that has no semantic garantees of being completable (i.e., all the required SEND to satisfy the FREEd RECV requests are posted before calling FINALIZE, and vice-versa). 

Open MPI would error (or at the minimum issue some warning) when during finalize we find some uncompleted/uncompletable requests. MPICH doesn't. On the other hand OMPI doesn't crash when calling CANCEL on a manifestly not-started request.

bosilca (approved):

PR #521, opened 2023-04-19
Allow for tag registration/deregistration any time.

bosilca:
We now have a higher level tag structure, and the engine works on a copy of it. Any time we want to update the tags we update the high level struct then raise a flag, and the communication engine will update its own struct at the next progress loop.


therault (changes requested):
See the unlikely race condition in the management of `parsec_ce_rebuild_array_of_requests`, and the question about if the user-level needs to make sure the tag is registered before sending messages or not. Otherwise,  LGTM.

parsec/parsec_mpi_funnelled.c
```diff
@@ -619,19 +634,24 @@ mpi_funnelled_fini(parsec_comm_engine_t *ce)
...
+ *        all necessary buffers. The registration can happen at any moment, before or after
+ *        the communication engine was started. If the communication engine is inactive,
+ *        the tags will be stored but anything related to the communication registration of
+ *        such AM will be delayed until the engine starts. If the engine is running, we
+ *        notify it that the AM infrastructure needs to be rebuilt, and it will do it at
+ *        the next progress cycle.
+ *        The requested tags should be from 0 up to MPI_FUNNELLED_MAX_TAG,
+ *        dynamic tags will start from MPI_FUNNELLED_MAX_TAG, but they are not handled here.
  */
 int
```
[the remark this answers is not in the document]
bosilca: No, this remains behind the parsec_ce.tag_register. It just does not need anything else from the communication engine (such as the communicator).

parsec/parsec_mpi_funnelled.c
```diff
@@ -640,71 +660,135 @@ mpi_no_thread_tag_register(parsec_ce_tag_t tag,
...
         return PARSEC_ERR_EXISTS;
     }
+    tag_struct->tag = tag;
+    tag_struct->msg_length = msg_length;
+    tag_struct->callback = callback;
+    tag_struct->cb_data = cb_data;
+    tag_struct->status = PARSEC_CE_TAG_STATUS_ENABLE;
 
+    /* Update the total number of requests we know about */
     size_of_total_reqs += EACH_STATIC_REQ_RANGE;
```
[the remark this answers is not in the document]
bosilca: I don't think it was ever thread-safe. As long as multiple threads are not concurently adding tags (which seem sensible), there is only one update to the size_of_total_reqs and one use.

parsec/parsec_mpi_funnelled.c
```diff
@@ -713,32 +805,26 @@ int
...
     }
+    tag_struct->status = PARSEC_CE_TAG_STATUS_DISABLE;
+    /* if the engine is active, notify it to rebuild the arrays of requests at the next cycle */
+    parsec_ce_am_design_version++;
+    parsec_atomic_unlock(&parsec_ce_am_build_lock);
 
-    free(tag_struct->buf[0]);
-    free(tag_struct->buf);
+    /* TODO: This part is not yet safe, we need to know if the engine is up or not */
+    mpi_no_thread_tag_unregister_internal(tag);
```
devreal, line 825: Is it not sufficient to defer this until the engine comes up? There won't be any AMs before and the first thing the progress thread does is to handle the internal registration.

bosilca: The real problem is that by allowing too much we are screwing ourselves. If we are using this is a very sensible way, then by the time the upper level unregister the tags the communication engine is already down, so there will be nobody to execute the tag unregister. But, I guess I can take care of this in the ce_fini.

devreal (approved):
LGTM :+1:

omor1:
Does 7615c1d only matter if the MPI doesn't support `mpi_assert_allow_overtaking`? I was under the impression that if the MPI implementation allowed overtaking it could use a tag matching table instead of a queue, so that matching cost is no longer linear in number of posted receives.

bosilca:
I am not sure I understand the question. Overtake basically removes the FIFO matching requirement on the wire, such that the receiver does not need to reorder the messages for the matching. It does not however remove the ordering of the matching on the receiver, requests should still be matched in the posted order.

omor1:
That's not my understanding of `mpi_assert_allow_overtaking`? Or what the standard says:
> If set to "true", then the implementation may assume that point-to-point communications on the given communicator do not rely on the non-overtaking rule specified in Section 3.5. In other words, the application asserts that send operations are not required to be matched at the receiver in the order in which the send operations were posted by the sender, and receive operations are not required to be matched in the order in which they were posted by the receiver.

bosilca:
It allows unordered matching of sends and receives, but the receives will be still matched in the order in which they were posted. 
            Rank 0                                  |                   Rank 1
s1: Send(tag, rank = 1, comm)       |  r1 = IRecv(tag, rank = 0, comm)
s2: Send(tag, rank = 1, comm)      |  r2 = iRecv(tag, rank = 0, comm)

The standard states that s1 and s2 can now match in any order r1 and r2 (aka either s1 -> r1 or s2 -> r2). However, from the receiver perspective r2 will only be matched once r1 was. This is not talking about the order in which they complete but the order in which they match.

omor1:
If that's the case, then it might behoove the standard to clarify—the phrasing "receive operations are not required to be matched in the order in which they were posted by the receiver" implies to me that ordering between receives is not required i.e. that there are no ordering requirements at all and that r2 can be matched and completed before r1.

therault (approved):
Tested on my side, it works.

abouteiller (approved):

parsec/remote_dep_mpi.c
```diff
@@ -2138,39 +2081,52 @@ remote_dep_mpi_get_end_cb(parsec_comm_engine_t *ce,
...
-    PARSEC_OBJ_CONSTRUCT(&dep_activates_noobj_fifo, parsec_list_t);
-    PARSEC_OBJ_CONSTRUCT(&dep_put_fifo, parsec_list_t);
-
     parsec_mpi_same_pos_items_size = context->nb_nodes + (int)DEP_LAST;
+    assert( NULL == parsec_mpi_same_pos_items );
     parsec_mpi_same_pos_items = (dep_cmd_item_t**)calloc(parsec_mpi_same_pos_items_size,
                                                         sizeof(dep_cmd_item_t*));
 
+    if(1 < context->nb_nodes) {
+        /* if nb_nodes==1, the parsec comm engine does not run with its own thread, so don't change the thread
```
devreal, line 2112: The comment describes the opposite of what the code does

parsec/scheduling.c
```diff
@@ -624,7 +631,7 @@ static int __parsec_taskpool_wait( parsec_taskpool_t* tp, parsec_execution_strea
         /* If there is a single process run and the main thread is in charge of
          * progressing the communications we need to make sure the comm engine
          * is ready for primetime. */
-        remote_dep_mpi_on(tp->context);
+        parsec_ce.enable(&parsec_ce);
```
devreal, line 634: Similar comment and call to `enable` is at https://github.com/ICLDisco/parsec/pull/521/files#diff-f3c94ef414b137da235f2a72ab4ba0b93d19b740d990ed9339d3d79881c6675bR729 but there we also `remote_dep_ce_reconfigure(parsec_context)` and `parsec_remote_dep_reconfigure(parsec_context)`. On purpose? Why?

parsec/parsec_mpi_funnelled.c
```diff
@@ -407,6 +407,49 @@ int parsec_mpi_sendrecv(parsec_comm_engine_t *ce,
...
+    }
+
+    /* The engine was never yet started, so there is nothing to clean */
+    if( NULL != ce->sync ) {
+        ce->disable(ce);
+        ce->sync(ce);
+        assert( -1 != context->comm_ctx );
+        MPI_Comm_free((MPI_Comm*)&context->comm_ctx);
+    }
+    rc = MPI_Comm_dup((MPI_Comm)opaque_comm_ctx, &comm);
```
devreal, line 437: Should we `MPI_Comm_dup_with_info` here and set the `mpi_assert_allow_overtaking` info key?

parsec/parsec_mpi_funnelled.c
```diff
@@ -603,108 +610,173 @@ mpi_funnelled_fini(parsec_comm_engine_t *ce)
...
     }
+    parsec_atomic_lock(&parsec_ce_am_build_lock);
+    tag_struct->tag = tag;
+    tag_struct->msg_length = (msg_length + 15) & ~0xF;  /* align to 16 bytes */
+    tag_struct->callback = callback;
+    tag_struct->cb_data = cb_data;
+    tag_struct->status = PARSEC_CE_TAG_STATUS_ENABLE;
 
+    /* Update the total number of requests we know about */
     size_of_total_reqs += EACH_STATIC_REQ_RANGE;
```
devreal, line 663: Shouldn't this be done in `parsec_ce_rebuild_am_requests`? I'm worried that otherwise the communication will see `size_of_total_reqs` out of sync with the request array in https://github.com/ICLDisco/parsec/pull/521/files#diff-0d6a2a76e0f5fedd623964083223b7f29041101eade63cb104701bb66750d761R921

parsec/parsec_mpi_funnelled.c
```diff
@@ -407,6 +407,49 @@ int parsec_mpi_sendrecv(parsec_comm_engine_t *ce,
...
+    MPI_Comm comm;
+    int rc;
+
+    /* We can only change the communicator if the communication engine is not active */
+    if( 1 < parsec_communication_engine_up ) {
+        parsec_warning("Cannot change PaRSEC's MPI communicator while the engine is running [ignored]");
+        return PARSEC_ERROR;
+    }
+
+    /* The engine was never yet started, so there is nothing to clean */
```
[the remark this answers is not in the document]
bosilca: The code does the cleaning only if the engine was started, so when the engine was not started there is nothing to clean.

parsec/parsec_mpi_funnelled.c
bosilca: We don't need to do that here, we never use this communicator for AM, so we don't really need to allow overtake on it. But we can create the rest of the communicators using the `dup_with_info`.

parsec/parsec_mpi_funnelled.c
bosilca: `current_size_of_total_reqs` is the current size of the arrays while `size_of_total_reqs` is the size it should be.

parsec/remote_dep_mpi.c
bosilca: not the opposite, but the other case.

parsec/scheduling.c
bosilca: singleton case, we only use MPI for local reshaping so we don't need the entire infrastructure (there will be no AM and no data transfer).

parsec/parsec_mpi_funnelled.c
devreal: The all the assert and checks for posting in the static array (https://github.com/ICLDisco/parsec/pull/521/files#diff-0d6a2a76e0f5fedd623964083223b7f29041101eade63cb104701bb66750d761R331) should use `current_size_of_total_reqs`?

bosilca: good catch !

devreal:
And there was much rejoice 🥳

PR #522, opened 2023-04-19
Update to the SDE available in PAPI-7.0.1

therault:
Changelog:
 - SDE is now an independent library from libpapi.
 - remove the SDE implementation from the parsec sources
 - only enable SDE counters if an installed libsde is found
 - add a FindSDE.cmake logic to find SDE (at the same place where PAPI can be found -- reuse PAPI_ROOT)
 - define a PAPI::SDE target to use in modern cmake
 - Include and link to the installed libsde


CMakeLists.txt
```diff
@@ -278,6 +266,25 @@ include (CheckCXXCompilerFlag)
 # check the capabilities of the system we are building for
 #
 
+# Check for PAPI::SDE
+if( PARSEC_PROF_TRACE )
+  if(CMAKE_SYSTEM_NAME MATCHES "Windows" OR
+     CMAKE_SYSTEM_NAME MATCHES "MSYS" OR
+     CMAKE_SYSTEM_NAME MATCHES "MINGW64")
```
bosilca, line 273: I don't think you need to test for these systems, just leave the find_package to do its job (and be unable to link against SDE).

bosilca (approved):

cmake_modules/FindSDE.cmake
```diff
@@ -0,0 +1,37 @@
+# - Find the SDE library (part of PAPI)
+# This module finds an installed  lirary that implements the
+# Softward Defined Events interface, part of the PAPI Project 
+# (see http://icl.cs.utk.edu/papi/).
+#
+# This module defines the PAPI::SDE target.
+##########
+
+find_path(SDE_INCLUDE_DIR sde_lib.h
+          PATHS ${PAPI_ROOT}/include ENV PAPI_INCLUDE_DIR
```
QingleiCao, line 10: I guess `PAPI_ROOT` (or `PAPI_INCLUDE_DIR` and `PAPI_LIBRARY_DIR`) needs to be passed to CMAKE, right? Because I found the module of `papi/5.7.0` on saturn does not provide `PAPI_INCLUDE_DIR` and `PAPI_LIBRARY_DIR`.

cmake_modules/FindSDE.cmake
```diff
@@ -0,0 +1,37 @@
...
+      add_library(PAPI::SDE INTERFACE IMPORTED)
+    endif()
+
+    set_property(TARGET PAPI::SDE APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${SDE_LIBRARY}")
+    set_property(TARGET PAPI::SDE PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${SDE_INCLUDE_DIR}")
+  #===============================================================================
+
+  set(SDE_LIBRARIES ${SDE_LIBRARY} )
+  set(SDE_INCLUDE_DIRS ${SDE_INCLUDE_DIR} )
+  mark_as_advanced(FORCE SDE_INCLUDE_DIR SDE_LIBRARY)
```
QingleiCao, line 35: (Not for this PR. I'm learning CMAKE.) What's the purpose of `mark_as_advanced` here?

therault:
On Wed, Apr 19, 2023 at 5:26 PM Qinglei Cao ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In cmake_modules/FindSDE.cmake
> <https://github.com/ICLDisco/parsec/pull/522#discussion_r1171860896>:
>
> > @@ -0,0 +1,37 @@
> +# - Find the SDE library (part of PAPI)
> +# This module finds an installed  lirary that implements the
> +# Softward Defined Events interface, part of the PAPI Project
> +# (see http://icl.cs.utk.edu/papi/).
> +#
> +# This module defines the PAPI::SDE target.
> +##########
> +
> +find_path(SDE_INCLUDE_DIR sde_lib.h
> +          PATHS ${PAPI_ROOT}/include ENV PAPI_INCLUDE_DIR
>
> I guess PAPI_ROOT (or PAPI_INCLUDE_DIR and PAPI_LIBRARY_DIR) needs to be
> passed to CMAKE, right? Because I found the module of papi/5.7.0 on
> saturn does not provide PAPI_INCLUDE_DIR and PAPI_LIBRARY_DIR.
>
Yes, if you don't have set PAPI_INCLUDE_DIR and PAPI_LIBRARY as cache
variables via the '-D' option, you can set them  in your environment. And
if you have not set any of those, you can pass
-DPAPI_ROOT=/path/to/papi/install to CMake, or to have a PAPI available in
the usual system paths.

I just reused what we do in FindPAPI.cmake.


> ------------------------------
>
> In cmake_modules/FindSDE.cmake
> <https://github.com/ICLDisco/parsec/pull/522#discussion_r1171867632>:
>
> > +if( SDE_FOUND )
> +    message(STATUS "SDE Library found at ${SDE_INCLUDE_DIR} ${SDE_LIBRARY}")
> +
> +    #===============================================================================
> +    # Importing PAPI SDE as a cmake target
> +    if(NOT TARGET PAPI::SDE)
> +      add_library(PAPI::SDE INTERFACE IMPORTED)
> +    endif()
> +
> +    set_property(TARGET PAPI::SDE APPEND PROPERTY INTERFACE_LINK_LIBRARIES "${SDE_LIBRARY}")
> +    set_property(TARGET PAPI::SDE PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${SDE_INCLUDE_DIR}")
> +  #===============================================================================
> +
> +  set(SDE_LIBRARIES ${SDE_LIBRARY} )
> +  set(SDE_INCLUDE_DIRS ${SDE_INCLUDE_DIR} )
> +  mark_as_advanced(FORCE SDE_INCLUDE_DIR SDE_LIBRARY)
>
> (Not for this PR. I'm learning CMAKE.) What's the purpose of
> mark_as_advanced here?
>
Again, copy/paste from FindPAPI.cmake. I assume the intention was to hide
the SDE_INCLUDE_DIR and SDE_LIBRARY variables to be manually setable by the
user, when in 'expert mode' of the cmake GUI. AFAIU from cmake manual,
`mark_as_advanced` only applies to cache variables, which SDE_LIBRARY and
SDE_INCLUDE_DIR are only if the user passes -DSDE_LIBRARY=/path/to/library
-DSDE_INCLUDE_DIR=/path/to/include/dir

cmake_modules/FindSDE.cmake
abouteiller: it doesn't show by default in ccmake

cmake_modules/FindSDE.cmake
abouteiller: PAPI_INCLUDE_DIR would be the output variable from findPAPI, I'd say we should only use PAPI_INCLUDE_DIR and not PAPI_ROOT on this secondary stage, but... 

given that SDE is a subcomponent of PAPI, this should be done during findPAPI so that we can use the COMPONENTS keyword during find_package(PAPI)

```
find_package(<PackageName> [version] [EXACT] [QUIET] [MODULE]
             [REQUIRED] [[COMPONENTS] [components...]]
             [OPTIONAL_COMPONENTS components...]
```
see https://cmake.org/cmake/help/latest/command/find_package.html?highlight=find_package

therault:
I discussed this option with the developer of SDE, and the answer was that
SDE is an independent library that is part of the PAPI project. There could
be systems where SDE is installed but not PAPI, and the parsec library
still provide the counters (read by an independently installed PAPI
library) via SDE. So, as a compromise, I treat SDE as an independent
library, that can be found via the same PAPI_ROOT.

cmake_modules/FindSDE.cmake
abouteiller: Answer below (SDE may be independent from PAPI) is sufficient to negate this proposition.

abouteiller (approved):

PR #525, opened 2023-04-20
Protect count_reshaping by PARSEC_DEBUG

QingleiCao:
Protect `count_reshaping` by PARSEC_DEBUG if PARSEC_DEBUG_HISTORY and PARSEC_DEBUG_NOISIER are ON when it's not in DEBUG mode.

### DIFF
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -378,7 +378,9 @@ remote_dep_dequeue_fini(parsec_context_t* context)
     PARSEC_OBJ_DESTRUCT(&dep_cmd_fifo);
     mpi_initialized = 0;
 
+#if defined(PARSEC_DEBUG)
     PARSEC_DEBUG_VERBOSE(10, parsec_debug_output, "Process has reshaped %zu tiles.", count_reshaping);
+#endif
 
     return 0;
 }


parsec/remote_dep_mpi.c
```diff
@@ -378,7 +378,9 @@ remote_dep_dequeue_fini(parsec_context_t* context)
     PARSEC_OBJ_DESTRUCT(&dep_cmd_fifo);
     mpi_initialized = 0;
 
+#if defined(PARSEC_DEBUG)
```
bosilca, line 381: PARSEC_DEBUG_VERBOSE only exists when PARSEC_DEBUG_NOISIER is defined, and this is only defined if we are in debug mode. This means I don't see a reason to protect it again here. What am I missing ?

QingleiCao: I see. Then do you think we need to make connections between `PARSEC_DEBUG` and  `PARSEC_DEBUG_XXX` in [CMAKE](https://github.com/ICLDisco/parsec/blob/3ad106fac8ff2e5440d574da85d3b58928b74d54/CMakeLists.txt#L183-L200) to ensure `PARSEC_DEBUG_XXX` can't be enabled when `PARSEC_DEBUG` is OFF?

bosilca: I'm mitigated here, because I do sometimes use history to debug a release build (because if I switch to debug build the error does not happen).

QingleiCao: I see. From [here](https://github.com/ICLDisco/parsec/blob/3ad106fac8ff2e5440d574da85d3b58928b74d54/parsec/utils/debug.h#L139-L148), I guess you enabled one of `PARSEC_DEBUG_HISTORY` and `PARSEC_DEBUG_NOISIER` but not both.

abouteiller: Yes, the advantage of HISTORY alone is that it doesn't cause as much system noise as real DEBUG_NOISIER and in many cases that helps the error continue to manifest with some debug info available.

abouteiller:
Invalid

PR #526, opened 2023-04-20
Profiling corrected.

josephjohnjj:
In Parsec Trace Table format V2 the information was stored incorrectly.

### DIFF
--- a/tools/profiling/dbpreader.c
+++ b/tools/profiling/dbpreader.c
@@ -780,7 +780,8 @@ int dbp_reader_last_error(const dbp_multifile_reader_t *dbp)
 
 void dbp_reader_close_files(dbp_multifile_reader_t *dbp)
 {
-    for( int i = 0; i < dbp->nb_files; i++ ) {
+	int i = 0;
+    for( i = 0; i < dbp->nb_files; i++ ) {
         dbp_reader_close_file(&dbp->files[i]);
     }
 }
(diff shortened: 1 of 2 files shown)


omor1:
`tools/profiling/dbpreader.c` seems to have been committed extraneously.
To clarify, this fixes a bug introduced by #432 that causes the PTT v2 format to be saved incorrectly if only one thread is used; the `event_infos` dictionary was being dropped.

tools/profiling/dbpreader.c
```diff
@@ -780,7 +780,8 @@ int dbp_reader_last_error(const dbp_multifile_reader_t *dbp)
 
 void dbp_reader_close_files(dbp_multifile_reader_t *dbp)
 {
-    for( int i = 0; i < dbp->nb_files; i++ ) {
```
therault, line 783: Indentation of `int i = 0;` should be aligned with the `for`.

therault (approved):
Please fix indentation before merging. Otherwise LGTM.

tools/profiling/dbpreader.c
bosilca: Do not rearrange other people's code. Remove this please.

josephjohnjj: Sorry, that was accidental.

bosilca (approved):

Issue #529, opened 2023-04-26
`profile=off` tasks are traced anyway, sometimes causing segfault


DSMishler:
## Describe the bug
Run `profile2h5` on the most recent version of dplasma and PaRSEC (as of April 25, 2023) and you will notice that it segfaults. With tracing on and with debug mode on, the issue becomes more clear. PaRSEC is tracing tasks that it was asked not to and sometimes this causes a segfault in `parsec/parsec/profiling.c:1002` (in debug mode).

### To Reproduce
compile parsec `518437965a5d788ed9d97ec7d5d9cae63001b328` and dplasma `518437965a5d788ed9d97ec7d5d9cae63001b328` with debug and tracing enabled.
`./testing_dgemm -N 4800 -nruns 2`

### Fixes
According to @therault, there are at least 3 ways to go about it

1- always generate a task type, so task_profiler can profile any task it wants. This is somehow contradictory with the 'profile = off' programmer request.
2- generate a 'NON_PROFILED' task type, and assign this task type to any task that is marked as profile=off. task_profiler can decide to ignore those or not, and we will not get the extended infos for those tasks.
3- fix task_profiler to not trace a task that is not registered.

With the current proposed fix:

To do the fix 3-, everywhere in `pins_task_profiler_module.c`, where we call `PARSEC_TASK_PROF_TRACE_FLAGS` or `PARSEC_PROFILING_TRACE` on a key that is in `task->taskpool->profiling_array`, if the value of the key in `task->taskpool->profiling_array` is <= 0, do not call the `profiling_trace` function.

omor1:
task_profiler profiling tasks with profile=off seems incorrect to me, so that should be fixed. Having a default profiling type id seems like a good idea though, to catch anything that is somehow profiled, but that we don't have any other information about.

bosilca:
```diff
diff --git a/parsec/mca/pins/pins.c b/parsec/mca/pins/pins.c
index 6e1e7a6c6..f75c6fdc6 100644
--- a/parsec/mca/pins/pins.c
+++ b/parsec/mca/pins/pins.c
@@ -12,6 +12,8 @@
 #include "parsec/constants.h"
 #include "parsec/utils/debug.h"
 #include "parsec/execution_stream.h"
+#include "parsec/parsec_internal.h"
+#include "parsec/parsec_binary_profile.h"
 
 /**
  * Mask for PINS events that are enabled by default.
[...]

PR #531, opened 2023-04-28
Profiling off fix

DSMishler:
A response to https://github.com/ICLDisco/parsec/issues/529.

Applied the patch suggested by @bosilca, but there were two more things that needed done.

1. Sometimes the `profiling_array` is not initialized. This was fixed to now have `pins.c` check first in `parsec_pins_instrument`.
2. The `profiling_array` wasn't truly being initialized to `-1`. Saying `static int myarr[200] = {-1}` only sets the first element to `-1` with the current PaRSEC compiler settings. I added code to truly initialize all tasks.

I will continue a few tests of my own and I encourage you to double-check as well. I'll let you know if I find anything else!

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -894,15 +894,6 @@ static char *dump_profiling_init(void **elem, void *arg)
                             jdf_basename, fname, 256-R, 256-G, 256-B,
                             jdf_basename, fname, jdf_basename, fname,
                             jdf_basename, fname, jdf_basename, fname);
-    // string_arena_add_string(info->sa,
-                            // "%s_profiling_array_counter = 0; /* initialize profiling array to -1 */\n"
-                            // "for(; %s_profiling_array_counter < %s_profiling_array_counter_limit; %s_profiling_array_counter++)\n"
-                            // "{\n"
-                            // "    %s_profiling_array[%s_profiling_array_counter] = -1;\n"
-                            // "}\n",
-                            // jdf_basename,
-                            // jdf_basename, jdf_basename, jdf_basename, 
-                            // jdf_basename, jdf_basename);
     string_arena_add_string(info->sa,
                             "parsec_profiling_add_dictionary_keyword(\"%s::%s\", \"fill:%02X%02X%02X\",\n"
                             "                                       sizeof(parsec_task_prof_info_t)+%d*sizeof(parsec_assignment_t),\n"
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -861,6 +861,15 @@ static char *dump_profiling_init(void **elem, void *arg)
     string_arena_t *profiling_convertor_params;
 
     if( !profile_enabled(f->properties) ) {
+        /**
+         * We have to set the `profiling_array` elements to -1
+         * then we will return NULL
+         */
+        string_arena_add_string(info->sa,
+                                "%s_profiling_array[START_KEY(%s_%s.task_class_id + PARSEC_%s_NB_TASK_CLASSES) /* %s (internal init) start key */] = -1;\n"
+                                "%s_profiling_array[END_KEY(%s_%s.task_class_id + PARSEC_%s_NB_TASK_CLASSES)   /* %s (internal init) end   key */] = -1;\n",
+                                jdf_basename, jdf_basename, fname, jdf_basename, fname,
+                                jdf_basename, jdf_basename, fname, jdf_basename, fname);
         return NULL;
     }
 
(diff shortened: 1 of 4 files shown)


DSMishler:
I'm doing some testing in this PR and I'm noticing some issues. Don't merge yet until these are resolved.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -1579,16 +1588,26 @@ static void jdf_generate_structure(jdf_t *jdf)
...
+     * but later they will be initialized
+     */
     if( need_profile )
         coutput("#if defined(PARSEC_PROF_TRACE)\n"
                 "#  if defined(PARSEC_PROF_TRACE_PTG_INTERNAL_INIT)\n"
                 "static int %s_profiling_array[4*PARSEC_%s_NB_TASK_CLASSES] = {-1}; /* 2 pairs (begin, end) per task, times two because each task class has an internal_init task */\n"
+                "static int %s_profiling_array_counter_limit = 4*PARSEC_%s_NB_TASK_CLASSES;\n"
                 "#  else /* defined(PARSEC_PROF_TRACE_PTG_INTERNAL_INIT) */\n"
                 "static int %s_profiling_array[2*PARSEC_%s_NB_TASK_CLASSES] = {-1}; /* 2 pairs (begin, end) per task */\n"
+                "static int %s_profiling_array_counter_limit = 2*PARSEC_%s_NB_TASK_CLASSES;\n"
```
therault, line 1602: We shouldn't need `%s_profiling_array_counter_limit` in this structure: AFAIU, it's always `2*nb_task_classes`.

bosilca: or 4 ... depending if we profile the internal startup tasks.

bosilca:
I have quite a few comments, because this make little sense to me.

1. the profiling_array of each taskpool is set during the taskpool constructor to point to the statically allocated write_check_profiling_array. This initialization is protected by the profile #define (which should always be consistently set across all builds from the same runtime), and by the `write_check_profiling_array[0] == -1`.
2. Because of this test it seems clear to me that we were not trying to initialize all the array but only the first element. If that element was not zero it simply means that the array should contains the correct entries in the profiling dictionary.
3. Instead of doing a loop and setting all values to -1 you should check what tasks have the `profile=off` in the constructor and only set their dictionary entries to -1. That would also remove the need for storing the number of entries in the profiling array.

therault:
@DSMishler we had a discussion about this PR during the PR reviews today.

 - Please leave the static initialization as is ( `profiling_array[...] = {-1, },` in the generated code ), this is used by other features. The -1 will have to be replaced by another value (in case the first task class is 'no-profile', but this is something orthogonal to your problem and to this PR;
 - We don't want the code to loop over all possible values and initialize to -1. jdf2c.c already loops over all entries in the profiling_array, and initializes the entries that are not 'no-profile'. You should just change that part of the code that skips over the 'no-profile' task classes to set the entry to -1 instead of leaving it untouched.

For the latter part, look for `dump_profiling_init` in jdf2c.c. You'll see that there is a loop over all task classes ('functions' in jdf2c.c), and for each task class, we call `dump_profiling_init` using the `UTIL_DUMP_LIST` macro that iterates over `functions`. `dump_profiling_init` checks if profiling is enabled for the task class passed as parameter, and if it is not, it just adds nothing to `info->sa`, otherwise it adds a call to `parsec_profiling_dictionary_add_keyword` to initialize two elements of the `profiling_array` array. You should just need to modify this function (`dump_profiling_init`) to add the initialization of the appropriate two elements of `profiling_array` to -1 instead of skipping over it.

DSMishler:
I checked to make sure the CI check passed and then squashed the commits (on my fork) so that PaRSEC won't have messy and non-compiling commits in its history for the merge.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -885,6 +894,15 @@ static char *dump_profiling_init(void **elem, void *arg)
                             jdf_basename, fname, 256-R, 256-G, 256-B,
                             jdf_basename, fname, jdf_basename, fname,
                             jdf_basename, fname, jdf_basename, fname);
+    // string_arena_add_string(info->sa,
```
therault, line 897: Do not leave a commented section like this in your code, if it's useless code, remove it before the merge.

parsec/mca/pins/pins.c
```diff
@@ -31,6 +33,19 @@ void parsec_pins_instrument(struct parsec_execution_stream_s* es,
                             parsec_task_t* task)
 {
     assert( method_flag < PARSEC_PINS_FLAG_COUNT );
+    /**
```
therault, line 36: To discuss: it's nice to put this check here, and not in each cb_func, because this removes code duplication, but at the same time, the test below changes the behavior of PINS:

 - if PARSEC_PROF_TRACE is OFF, and task != NULL, no callback is called. This could be less intrusive by protecting the entire block (lines 40 to 47) behind the #ifdef
 - if a PINS module doesn't use parsec_profiling in cb_func, but uses it to take some other statistics, it's not called anyway.

I think we should pay the cost of code duplication here, to keep the feature that some pins module can collect statistics without profiling.

DSMishler: w.r.t. line 897 in `jdf2c.c`, you're right. Oops, shouldn't have left that in. Fixed.

I'll be checking this email occasionally while off work. Just wanted to let you know I still see this stuff. I added a commit to change this, but I don't know what you mean by each callback function (`cb_func`). Where is this field set? Should we just write a wrapper that checks for the appropriate fields to be non-NULL and _then_ calls the cb_func?

therault (approved):

therault (changes requested):
There are copy/paste typos in these changes. I will create a PR from this one and fix them so we can merge faster.

DSMishler:
Closing with the merge of PR #594

PR #534, opened 2023-05-05
CMake/configure fixes 

abouteiller:
* Fixes FLEX compiler flags not being applied to FLEX generated files
* cache results for -w flag detections
* move the configure --with-python-venv log for the venv creation inside the venv itself.


cmake_modules/ParsecCompilerFlags.cmake
```diff
@@ -43,40 +43,43 @@ function(check_and_set_compiler_option)
...
-      message(TRACE "Checking flag ${parsec_cas_co_OPTION} for language ${parsec_cas_co_LANGUAGE}")
-      cmake_language(CALL "check_${parsec_cas_co_LANGUAGE}_compiler_flag" "${parsec_cas_co_OPTION}" "${parsec_cas_co_NAME}_${parsec_cas_co_LANGUAGE}" )
-      if( "${${parsec_cas_co_NAME}_${parsec_cas_co_LANGUAGE}}" )
-        set(_supported_languages "${parsec_cas_co_LANGUAGE};${_supported_languages}")
-        message(TRACE "Flag ${parsec_cas_co_OPTION} supported by ${parsec_cas_co_LANGUAGE}")
-        if(DEFINED parsec_cas_co_CONFIG)
-          list(APPEND parsec_cas_co_defined_options "$<$<CONFIG:${parsec_cas_co_CONFIG}>:$<$<COMPILE_LANGUAGE:${parsec_cas_co_LANGUAGE}>:${parsec_cas_co_OPTION}>>")
-        else()
-          list(APPEND parsec_cas_co_defined_options "$<$<COMPILE_LANGUAGE:${parsec_cas_co_LANGUAGE}>:${parsec_cas_co_OPTION}>")
+  if(NOT DEFINED ${parsec_cas_co_NAME})
```
bosilca, line 46: If I correctly read the code, we need to define all supported languages before we start testing the flags or this will not work (because `parsec_cas_co_NAME` might not have tested for all the expected `parsec_cas_co_LANGUAGE`).

abouteiller: This is correct but has no practical effect: we never call this function outside of the `ParsecCompilerFlags.cmake`  file, which is included only once at the root of the project. At this call site all compilers have been enabled correctly already. 

The main reason we have this as a function is that it makes it easier to carry changes to the dplasma file with the same functionality. It was never intended that we would use the function outside of the file itself. I just quick checked and there is no way to undefined a function in cmake, I could add a comment to that effect though.

PR #535, opened 2023-05-09
Fix bug: some scenarios would call a nullptr function in profiling

therault:
One should never call `info_fn` if `info_fn` is `NULL` (obviously), and the PARSEC_PROF_TRACE_TASK macro always pass the task as `info_data` if the higher level wants to log. Change the test to test on `info_fn` instead of `info_data`.

### DIFF
--- a/parsec/parsec_internal.h
+++ b/parsec/parsec_internal.h
@@ -590,7 +590,7 @@ extern int device_delegate_begin, device_delegate_end;
                            key_hash((TASK)->task_class->make_key(           \
                               (TASK)->taskpool, (TASK)->locals ), NULL),    \
                            (TASK)->taskpool->taskpool_id,                   \
-                           (TASK)->task_class->profile_info,                \
+                           (TRACE_INFO) ? (TASK)->task_class->profile_info : NULL, \
                            (TRACE_INFO) ? (void*)(TASK) : NULL,             \
                            (FLAGS)); 
 #else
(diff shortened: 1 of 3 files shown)


parsec/parsec_internal.h
```diff
@@ -590,7 +590,7 @@ extern int device_delegate_begin, device_delegate_end;
                            key_hash((TASK)->task_class->make_key(           \
                               (TASK)->taskpool, (TASK)->locals ), NULL),    \
                            (TASK)->taskpool->taskpool_id,                   \
-                           (TASK)->task_class->profile_info,                \
+                           (TRACE_INFO) ? (TASK)->task_class->profile_info : NULL, \
```
bosilca, line 593: why is the `task_class->profile_info` not initialized to NULL ?

therault: profile_info points to the function defined by the task class.

TRACE_INFO is a macro argument that decides if we should log the info now (typically by PINS, either at the begin or the end of the event).

With this change, we keep the previous functionality (if the task class defines a profile_info function, PINS decides if we log the info or not), and we support the case where the task class has not defined a profile_info function (DTD and PTG always define one, because there is always at least the basic info of the task, but MADNESS over parsec doesn't define one).

bosilca (approved):

abouteiller (approved):

PR #538, opened 2023-05-10
Limit the number of recv requests to ensure there is space for sends

devreal:
Large amounts of GET operations lead to enough recv requests in the request array and the request fifo to prevent any send operation to happen, causing a deadlock.

Reserve half of the dynamic request array for send requests by limiting the number of recv requests that can be put into the request array.

### DIFF
--- a/parsec/parsec_mpi_funnelled.c
+++ b/parsec/parsec_mpi_funnelled.c
@@ -100,6 +100,7 @@ static int mpi_funnelled_tag_unregister_unsafe_internal(parsec_ce_tag_t tag);
  * spots in the request array.
  */
 #define MAX_DYNAMIC_REQ_RANGE 30 /* according to current implementation */
+#define MAX_NUM_RECV_REQ_IN_ARRAY 15
 #define EACH_STATIC_REQ_RANGE 5 /* for each registered tag */
 
 typedef enum parsec_ce_tag_status_e {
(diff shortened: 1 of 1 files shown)


therault (approved):
This solves the problem identified in TTG over PaRSEC Master, at least at small scale tests.

Implementation seems alright to me, but I think we need to discuss this / iterate with @bosilca because there are important changes here.

devreal:
One thing I forgot to note: this will also prevent recv from drowning out because the code in `mpi_no_thread_push_posted_req` will prioritize recv requests if they are available and the maximum number has not yet been reached. That allows more than half of the request array to be taken by send requests if there are no recv requests available but as soon as they become available and requests complete we will start filling recv requests again, up to the limit. The maximum number of active send operations is 30 and there is a guaranteed minimum of 15 if they exist.

Limiting the number of recv requests to be tested in `MPI_Testsome` should have no negative impact on performance.

parsec/parsec_mpi_funnelled.c
```diff
@@ -100,6 +100,7 @@ static int mpi_funnelled_tag_unregister_unsafe_internal(parsec_ce_tag_t tag);
  * spots in the request array.
  */
 #define MAX_DYNAMIC_REQ_RANGE 30 /* according to current implementation */
+#define MAX_NUM_RECV_REQ_IN_ARRAY 15
 #define EACH_STATIC_REQ_RANGE 5 /* for each registered tag */
```
abouteiller, line 104: It doesn't need to be done in this PR but it would be great if these were mutable MCAs

abouteiller (approved):

abouteiller:
Performance found to be similar or better on QR, small sizes.

PR #540, opened 2023-05-11
Print -wflags found only when results differ from cache

abouteiller:
This has no real semantic changes. 

* Reduce verbosity of -wflags when results pulled from cached variables without changes (does not preclude re-testing with more enabled_languages, unlike #534)
* Do not add " " to compiler_flags when no options found
* move flex options testing to the end to ease porting to dplasma (this piece doesn't exist in the dplasma file)


cmake_modules/ParsecCompilerFlags.cmake
```diff
@@ -64,19 +64,24 @@ function(check_and_set_compiler_option)
...
+    # Something new found, output result
+    if( _supported_languages STREQUAL "" )
+      message(STATUS "Performing Test Flag ${parsec_cas_co_OPTION} supported by languages -\t[None]")
+    else( _supported_languages STREQUAL "" )
+      message(STATUS "Performing Test Flag ${parsec_cas_co_OPTION} supported by languages -\t[${_supported_languages}]")
+    endif( _supported_languages STREQUAL "" )
+  endif()
 
   if(DEFINED parsec_cas_co_COMMENT)
     set(${parsec_cas_co_NAME} "${parsec_cas_co_defined_options}" CACHE STRING "${parsec_cas_co_COMMENT}")
```
abouteiller, line 78: Discussed during meeting about this not using FORCE; we will remove the STRING type and set it to INTERNAL.

PR #542, opened 2023-05-12
Small warnings from clang14

abouteiller:


therault:
@bosilca wanted to leave these warnings, as a reminder that the API should change (since we don't use the context, this doesn't require the context). I'll let him comment if we still need to fix that.

PR #549, opened 2023-05-26
Tear-down parsec_ce after high-level communication

devreal:
The finalization order is important: `parsec_remote_dep_fini` finalizes the high-level remote dependency infrastructure (e.g., progress thread), `remote_dep_ce_fini` tears-down data structures on top of the comm engine (queues). Only then can we finalize the comm engine itself.

Also fixes some weird casts that lead to strange output.

### DIFF
--- a/parsec/parsec_comm_engine.c
+++ b/parsec/parsec_comm_engine.c
@@ -26,10 +26,9 @@ extern int remote_dep_ce_fini(parsec_context_t* context);
 int
 parsec_comm_engine_fini(parsec_comm_engine_t *comm_engine)
 {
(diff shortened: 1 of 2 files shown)


abouteiller (approved):

abouteiller:
Can you document in the PR description what problem it solves

devreal:
@abouteiller Done :+1:

bosilca:
@devreal I confirm this is the correct order. `parsec_remote_dep_fini` must be first because it is the blocking function that signal the communication engine to turn off. `remote_dep_ce_fini` can be next as it unregister the AM for activate and get_data and release the memory management constructs, so it needs the communication engine off.

bosilca (approved):

PR #556, opened 2023-06-14
Gpu pingpong test

therault:
This creates a simple test that makes data be updated on CPU and on GPUs alternatively.

As part of the test, I found out that HIP was not ported over DTD, and this ports HIP over DTD. It also provides a first test for HIP.


bosilca (dismissed):

abouteiller (dismissed):

therault:
I'm a bit lost with CI here... Another pair of eyes would help. To summarize what I observe:

 - When running in shared=OFF / profiling=ON mode, we don't detect CUDA at all (no device, no compiler)
 - When running in shared=ON / profiling=OFF mode, we always detect CUDA (device part).
     - In the master version, we asked for the slack package gcc@12.something, and that makes `check_language(CUDA)` fail, because `nvcc` cannot work with a gcc > 11.x
     - In the version proposed in this patch, we load the slack package gcc@11.3.0. Now something even more curious is happening:
         - We detect CUDAToolkit and enable the cuda device without aproblem
         - We still claim that `check_language(CUDA)` fails.
         - To investigate why, I have added some CMake messages that are output of the current failing job (https://github.com/ICLDisco/parsec/actions/runs/5294237295/jobs/9583355256?pr=556).
             - `nvcc` is where it should be based on the CUDA toolkit we have discovered
             - I can run succesfully `nvcc -c /path/to/some/cufile.cu`
             - No CMakeError.log file is generated. I display the contents of CMakeFiles/ and it doesn't seem it contains any useful information.

To conclude, I have no idea why `check_language(CUDA)` fails in this setup, and I'm now out of ideas to test...

parsec/interfaces/dtd/insert_function.c
```diff
@@ -2334,7 +2336,14 @@ static parsec_hook_return_t parsec_dtd_gpu_task_submit(parsec_execution_stream_t
...
         return parsec_cuda_kernel_scheduler(es, gpu_task, dev_index);
 #endif  /* PARSEC_HAVE_DEV_CUDA_SUPPORT */
+#if defined(PARSEC_HAVE_DEV_HIP_SUPPORT)
+    case PARSEC_DEV_HIP:
+        gpu_task->stage_in  = parsec_default_hip_stage_in;
+        gpu_task->stage_out = parsec_default_hip_stage_out;
+        return parsec_hip_kernel_scheduler(es, gpu_task, dev_index);
+#endif  /* PARSEC_HAVE_DEV_HIP_SUPPORT */
     default:
+        assert(0);
```
abouteiller, line 2346: remove

tests/dsl/dtd/dtd_test_cuda_task_insert.c
```diff
@@ -4,6 +4,7 @@
 #include "parsec/data_dist/matrix/two_dim_rectangle_cyclic.h"
 #include "parsec/interfaces/dtd/insert_function_internal.h"
 #include "tests/tests_data.h"
+#include "parsec/mca/device/cuda/device_cuda_internal.h"
```
abouteiller, line 7: why do we need this, this is internal and should not spillover into user code.

bosilca:
please rebase.

therault:
Split this PR in two: one for the tester itself and another for the CI/runners

abouteiller:
There has not been a lot of progress on this, and I think we can autogenerate (hipify) the hip testers, lets postpone to 4.1

bosilca:
Partially replaced by #775.

PR #557, opened 2023-06-16
Correct the logic for passing over CPU/RECURSIVE devices

abouteiller:
Correct the logic for passing over CPU/RECURSIVE devices when the recursive device is not built-in

We use the dev->type instead of the dev_index to identify the devices.

### DIFF
--- a/parsec/interfaces/dtd/insert_function.c
+++ b/parsec/interfaces/dtd/insert_function.c
@@ -2304,7 +2304,7 @@ static parsec_hook_return_t parsec_dtd_gpu_task_submit(parsec_execution_stream_t
 
     dev_index = parsec_get_best_device(this_task, &load);
     assert(dev_index >= 0 && dev_index < parsec_mca_device_enabled());
-    if (parsec_mca_device_get(dev_index)->type <= PARSEC_DEV_RECURSIVE) {
+    if (!parsec_mca_device_is_gpu(dev_index)) {
         return PARSEC_HOOK_RETURN_NEXT; /* Fall back */
     }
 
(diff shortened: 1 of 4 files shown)


parsec/interfaces/dtd/insert_function.c
```diff
@@ -2303,8 +2303,8 @@ static parsec_hook_return_t parsec_dtd_gpu_task_submit(parsec_execution_stream_t
     parsec_dtd_task_class_t *dtd_tc = (parsec_dtd_task_class_t*)this_task->task_class;
 
     dev_index = parsec_get_best_device(this_task, &load);
-    assert(dev_index >= 0);
-    if (dev_index < 2) {
+    assert(dev_index >= 0 && dev_index < parsec_mca_device_enabled());
+    if (parsec_mca_device_get(dev_index)->type <= PARSEC_DEV_RECURSIVE) {
```
devreal, line 2307: Comparing with `PARSEC_DEV_RECURSIVE` is still not great, IMHO. For one, it's not intuitive as it relies on a specific ordering of device identifier. Second, it's error prone (e.g., if the order changes) and verbose (which is probably the reason for the short-cut `< 2`).

Could we have masks to check for device classes? Or a function `parsec_mca_device_is_accel(dev_index)`?

abouteiller:
* [ ] fix bug that makes us miscount gpu_copy->readers

abouteiller:
bug is actually due to #554

bosilca:
Not sure I understand how the bug can be due to a yet-to-be merged PR ?

abouteiller:
wrong number, it's the async one.

abouteiller:
this needs reviews

PR #558, opened 2023-06-22
Add capability of saving GPU statistics and printing diff vs saved stats

abouteiller:
1. new API parsec_devices_save_statistics fills up an array of statistics for later use
2. new API parsec_devices_print_statistics prints the diff between start_stats (as obtained in 1.) and the current internal stats
3. new API parsec_devices_free_statistics deallocate the array obtained from save_statistics. 

This is to support customizable reporting of stats, see https://github.com/ICLDisco/dplasma/pull/89

### DIFF
--- a/parsec/mca/device/device.c
+++ b/parsec/mca/device/device.c
@@ -330,29 +330,46 @@ void parsec_compute_best_unit( uint64_t length, float* updated_value, char** bes
     return;
 }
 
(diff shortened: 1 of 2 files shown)


devreal:
Is this a public API? Are we removing an old public API, breaking existing codes? Can both coexist?

abouteiller:
The new 3 APIs in the .h are public, the old 2 APIs were private (and still exist, called from parsec_finalize)

devreal (approved):

PR #561, opened 2023-06-28
Profiling off fix - 2

therault:
This PR fixes some issues with PR https://github.com/ICLDisco/parsec/pull/531
It supersedes PR https://github.com/ICLDisco/parsec/pull/531 which should be closed once this is merged.

The author of PR https://github.com/ICLDisco/parsec/pull/531 is in internship and may not have time to update his PR, so I took this route. I kept @DSMishler commits in this PR to maintain ownership as possible.

The issue with the previous PR is described in the commit message, returning NULL from the dumping function (as I adviced to do) made the changes empty in the generated code and hid the other problems.

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -894,15 +894,6 @@ static char *dump_profiling_init(void **elem, void *arg)
                             jdf_basename, fname, 256-R, 256-G, 256-B,
                             jdf_basename, fname, jdf_basename, fname,
                             jdf_basename, fname, jdf_basename, fname);
-    // string_arena_add_string(info->sa,
-                            // "%s_profiling_array_counter = 0; /* initialize profiling array to -1 */\n"
-                            // "for(; %s_profiling_array_counter < %s_profiling_array_counter_limit; %s_profiling_array_counter++)\n"
-                            // "{\n"
-                            // "    %s_profiling_array[%s_profiling_array_counter] = -1;\n"
-                            // "}\n",
-                            // jdf_basename,
-                            // jdf_basename, jdf_basename, jdf_basename, 
-                            // jdf_basename, jdf_basename);
     string_arena_add_string(info->sa,
                             "parsec_profiling_add_dictionary_keyword(\"%s::%s\", \"fill:%02X%02X%02X\",\n"
                             "                                       sizeof(parsec_task_prof_info_t)+%d*sizeof(parsec_assignment_t),\n"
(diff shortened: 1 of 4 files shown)


tests/profiling/check-async.py
```diff
@@ -12,12 +12,11 @@
     FULL_RESCHED = t.event_types['FULL_RESCHED']
     ASYNC = t.event_types['async::ASYNC']
     RESCHED = t.event_types['async::RESCHED']
-    STARTUP = t.event_types['async::STARTUP']
 except AttributeError:
         print("HDF5 file {} does not contains the event_types attribute. The file might be empty, corrupted or it was incorrectly generated".format(filename), file=sys.stderr)
         sys.exit(2)
 except KeyError:
-    print("One of keys FULL_ASYNC, FULL_RESCHED, ASYNC, RESCHED, or STARTUP is not defined in the trace",
```
bosilca, line 20: I think we should check that we have no STARTUP task in the profile.

parsec/mca/pins/pins.c
```diff
@@ -31,6 +33,19 @@ void parsec_pins_instrument(struct parsec_execution_stream_s* es,
                             parsec_task_t* task)
 {
     assert( method_flag < PARSEC_PINS_FLAG_COUNT );
+    #if defined(PARSEC_PROF_TRACE)
```
bosilca, line 36: It would be cheaper to move this check directly in the PARSEC_PINS macro (in parsec/mca/pins/pins.h) and save one function call for non profile-able events. Here we can keep an assert to make sure we are doing the right thing.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -4708,12 +4727,17 @@ static void jdf_generate_constructor( const jdf_t* jdf )
     pi.idx = 0;
     JDF_COUNT_LIST_ENTRIES(jdf->functions, jdf_function_entry_t, next, pi.maxidx);
     {
-        char* prof = UTIL_DUMP_LIST(sa1, jdf->functions, next,
-                                    dump_profiling_init, &pi, "", "    ", "\n", "\n");
+        int need_profile = 0;
```
bosilca, line 4730: If I understand this code correctly it is meant to save us 2 integers per task class, but it makes the code more complicated to read. Not sure this is worth the effort.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -860,12 +860,27 @@ static char *dump_profiling_init(void **elem, void *arg)
     int nb_locals;
     string_arena_t *profiling_convertor_params;
 
+    string_arena_init(info->sa);
     if( !profile_enabled(f->properties) ) {
-        return NULL;
+        /**
+         * We have to set the `profiling_array` elements to -1
```
bosilca, line 866: I don't understand this comment and it does not seem to match the code.

abouteiller (approved):

bosilca:
The correct solution to this issue is proposed in #594.

PR #563, opened 2023-06-29
Fix the management of GPU copies.

bosilca:
This patch fixes few issues:
- consistent return codes (based on HOOK_RETURN + positive values) for all the callbacks in the device code.
- avoid overwriting the task data_in. Instead keep them in a temporary copy in the gpu_task. Removes many bugs in the DSLs.
- Correctly handle the refcount and readers of the copies in the D2D case.

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -1261,7 +1261,6 @@ parsec_gpu_data_stage_in( parsec_device_cuda_module_t* cuda_device,
     parsec_gpu_data_copy_t* gpu_elem = task_data->data_out;
     uint32_t nb_elts = gpu_task->flow_nb_elts[flow->flow_index];
     int transfer_from = -1;
-    int undo_readers_inc_if_no_transfer = 0;
 
     if( gpu_task->task_type == PARSEC_GPU_TASK_TYPE_PREFETCH ) {
         PARSEC_DEBUG_VERBOSE(3, parsec_gpu_output_stream,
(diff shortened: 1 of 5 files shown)


devreal (approved):

QingleiCao:
We do not have node hours on Summit and Frontier, so I tested this PR using HiCMA on ICL:guyot, which has 8 A100 GPUs. It seems there is a deadlock if using the current master, but this PR solves this issue. Regarding performance, I tried one case, i.e., N=163840, NB=2048, nb_gpus=8, and double precision. The performance is a little lower compared to the previous result, i.e., 95 Tflop/s now and 112 Tflop/s before. I will do more experiments. BTW, the same problem, DPLASMA shows: `W@00000 /home/qcao3/hicma-x-dev-icl/build-a100/dplasma/src/dpotrf_U.c:6915 cublasDpotrf_v2 profiler not initialized: call cudaProfilerInitialize()
W@00000 /home/qcao3/hicma-x-dev-icl/dplasma/parsec/parsec/mca/device/cuda/device_cuda_module.c:1196 cudaMemcpyAsync an illegal memory access was encountered
W@00000 /home/qcao3/hicma-x-dev-icl/dplasma/parsec/parsec/mca/device/cuda/device_cuda_module.c:1465 gpu_task->stage_in
W@00000 <<0x7f4668080000 on CUDA device 1>> -> <<0x7ee7dc000000 on CUDA device 6>> [33554432, D2D]
W@00000 Critical issue related to the GPU discovered. Giving up`

PR #566, opened 2023-08-09
co_manager shortcuting the scheduler

BrieucNicolas:
Continuation of #509 adding a shortcut when discovering tasks in the dtd. The co-manager executes ready tasks discovered during completion immediatly instead of scheduling them

### DIFF
--- a/parsec/interfaces/dtd/insert_function.c
+++ b/parsec/interfaces/dtd/insert_function.c
@@ -40,6 +40,10 @@
 
 #if defined(PARSEC_HAVE_DEV_CUDA_SUPPORT)
 #include "parsec/mca/device/cuda/device_cuda.h"
+#include "parsec/sys/tls.h"
+
+extern PARSEC_TLS_DECLARE(co_manager_tls);
+
 #endif  /* defined(PARSEC_HAVE_DEV_CUDA_SUPPORT) */
 
 #include "parsec/mca/mca_repository.h"
@@ -224,8 +228,8 @@ static int parsec_dtd_taskpool_leave_wait(parsec_taskpool_t* tp, void*_)
     parsec_termdet_open_module(tp, "local");
     tp->tdm.module->monitor_taskpool(tp, parsec_taskpool_termination_detected);
     tp->tdm.module->taskpool_set_nb_tasks(tp, 0);
-    tp->tdm.module->taskpool_set_runtime_actions(tp, 0); 
-    
+    tp->tdm.module->taskpool_set_runtime_actions(tp, 0);
+
     /* We are re-attached to the context */
     parsec_atomic_fetch_inc_int32(&tp->context->active_taskpools);
     return PARSEC_SUCCESS;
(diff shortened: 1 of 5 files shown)


parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -2579,12 +2596,15 @@ parsec_cuda_kernel_scheduler( parsec_execution_stream_t *es,
...
+     *             or
+     *             There is a manager, but it is waiting for the co_manager to release
+     *             new tasks, so we just append our task and don't become manager
      *   - rc > 0: there is a manager, and at the exit of the while, this thread has
      *             committed new work that the manager will need to do, but the work is
      *             not in the queue yet.
      */
     while(1) {
-        rc = gpu_device->mutex;
+        rc = rc1 = gpu_device->mutex;
```
devreal, line 2607: I don't like `rc1` for its lack of information. I suggest using a boolean variable like `was_first_co_manager` (you get the point :)) that is set to `true` if the CAS below succeeds and `rc == 1`.

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -2781,4 +2877,91 @@ parsec_cuda_kernel_scheduler( parsec_execution_stream_t *es,
...
+
+        gpu_task = (parsec_gpu_task_t*)parsec_fifo_pop( &(gpu_device->to_complete) );
+        if( gpu_task != NULL)
+        {
+            task = gpu_task->ec;
+            __parsec_complete_execution( es, task );
+            PARSEC_DEBUG_VERBOSE(4, parsec_gpu_output_stream,"GPU[%s]: gpu_task %p completed by co-manager %d at %s:%d", gpu_device->super.name,
+                                 gpu_task, es->th_id, __FILE__, __LINE__);
+            parsec_atomic_fetch_dec_int32( &(gpu_device->complete_mutex) );
+            parsec_list_push_back(gpu_tasks_to_free, (parsec_list_item_t*)gpu_task);
```
devreal, line 2939: Why do we have to accumulate task objects here? Can they not be free'd directly?

devreal (changes requested):
I am not familiar with #509 so I might be missing something in my review.

parsec/mca/device/cuda/device_cuda_module.c
BrieucNicolas: Yes, thank you for pointing that out. i found a fix for this, it will be applied

parsec/interfaces/dtd/insert_function.c
```diff
@@ -40,6 +40,10 @@
 
 #if defined(PARSEC_HAVE_DEV_CUDA_SUPPORT)
 #include "parsec/mca/device/cuda/device_cuda.h"
+#include "parsec/sys/tls.h"
+
+extern PARSEC_TLS_DECLARE(co_manager_tls);
```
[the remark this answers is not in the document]
devreal: This is moot. I looked at an earlier version of PaRSEC, the `static` qualifier was removed in https://github.com/ICLDisco/parsec/commit/bf264d9d85d552b1cb1aa5c074c6f2a9f3587c2a (but the accompanying comment wasn't updated).

bosilca:
We have a simpler and less intrusive approach in #762. It lacks the DTD support for executing the ready tasks returned by the completion immediately, but if we really want that capability we should reimplement it instead of reviving this PR.

PR #569, opened 2023-08-25
Topic/add evaluate keyword

bosilca:
Allow the user to defined validators for specific BODY chores. The chores will then be disabled per task instance (aka for each task) until one of the chores accepts to run and the BODY itself completes successfully.

### DIFF
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -151,7 +151,7 @@ int __parsec_execute( parsec_execution_stream_t* es,
     }
 
     PARSEC_PINS(es, EXEC_BEGIN, task);
-    /* Try all the incarnations until one agree to execute. */
+    /* Try all the incarnations until one accepts to execute. */
     do {
         if( NULL != (eval = tc->incarnations[chore_id].evaluate) ) {
             rc = eval(task);
@@ -163,7 +163,6 @@ int __parsec_execute( parsec_execution_stream_t* es,
                                          tmp, tc->incarnations[chore_id].type,
                                          chore_id);
 #endif
-                    break;
                 }
                 goto next_chore;
             }
@@ -185,13 +184,9 @@ int __parsec_execute( parsec_execution_stream_t* es,
             if( PARSEC_HOOK_RETURN_ASYNC != rc ) {
                 /* Let's assume everything goes just fine */
                 task->status = PARSEC_TASK_STATUS_COMPLETE;
-                parsec_device_module_t *dev = NULL;
-                if(PARSEC_DEV_CPU == tc->incarnations[chore_id].type) {
-                    dev = parsec_mca_device_get(0);
-                    parsec_atomic_fetch_inc_int64((int64_t*)&dev->executed_tasks);
-                }
-                else if(PARSEC_DEV_RECURSIVE == tc->incarnations[chore_id].type) {
-                    dev = parsec_mca_device_get(1);
+                if(PARSEC_DEV_RECURSIVE >= tc->incarnations[chore_id].type) {
+                    /* accelerators count their own executed tasks */
+                    parsec_device_module_t *dev = parsec_mca_device_get(chore_id);
                     parsec_atomic_fetch_inc_int64((int64_t*)&dev->executed_tasks);
                 }
             }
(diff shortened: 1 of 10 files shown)


parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -1765,8 +1765,8 @@ parsec_gpu_callback_complete_push(parsec_device_gpu_module_t   *gpu_device,
         assert( flow );
         assert( flow->flow_index == i );
         if(PARSEC_FLOW_ACCESS_NONE == (PARSEC_FLOW_ACCESS_MASK & flow->flow_flags)) continue;
-        if(PARSEC_DATA_STATUS_UNDER_TRANSFER == task->data[i].data_out->data_transfer_status ) {
-            assert(task->data[i].data_out->push_task == task );  /* only the task who did this PUSH can modify the status */
+        if( PARSEC_DATA_STATUS_UNDER_TRANSFER == task->data[i].data_out->data_transfer_status ) {
+            /* only the task who did the PUSH can modify the status */
```
devreal, line 1769: Is this comment still relevant? The associated assert has been removed and the following doesn't seem to be related.

devreal (approved):
LGTM although I cannot say much about the JDF part. One minor comment.

parsec/mca/device/cuda/device_cuda_module.c
bosilca: the comment is still valid but we do not have a direct way to know who was the task that initiated the transfer. However, as the transfers in each direction are sequentialized for each GPU, the first task that completes its transfer and has any of the data marked as "under transfer" must be the task that initiated it and thus the only one allowed to change it.

tests/dsl/ptg/user-defined-functions/udf.jdf
```diff
@@ -65,7 +79,7 @@ UD_STARTUP1(m, n) [ startup_fn = ud_startup1 ]
 
 READ X <- A(m, n)
 
-BODY
+BODY [evaluate = %{ return PARSEC_HOOK_RETURN_DONE; %}]
```
therault, line 82: Did you intend to use `always_here` here? Or you want to use the inline (in which case we don't need to define `always_here` above)?

therault (commented):
Just a couple of nitpicking, LGTM overall.

parsec/scheduling.c
```diff
@@ -185,13 +184,8 @@ int __parsec_execute( parsec_execution_stream_t* es,
...
                 /* Let's assume everything goes just fine */
                 task->status = PARSEC_TASK_STATUS_COMPLETE;
-                parsec_device_module_t *dev = NULL;
-                if(PARSEC_DEV_CPU == tc->incarnations[chore_id].type) {
-                    dev = parsec_mca_device_get(0);
-                    parsec_atomic_fetch_inc_int64((int64_t*)&dev->executed_tasks);
-                }
-                else if(PARSEC_DEV_RECURSIVE == tc->incarnations[chore_id].type) {
-                    dev = parsec_mca_device_get(1);
+                if(PARSEC_DEV_RECURSIVE >= tc->incarnations[chore_id].type) {
```
abouteiller, line 187: We have a macro to find this in a 'portable' way: `parsec_mca_device_is_gpu`
The removal of the counting for the CPUs will break again the report for CPU tasks executed?

tests/dsl/ptg/user-defined-functions/udf.jdf
```diff
@@ -79,14 +93,32 @@ UD_HASH_STRUCT(m, n) [ hash_struct = ud_hash_struct ]
 
 READ X <- A(m, n)
 
-BODY
+BODY [ evaluate = never_here
```
abouteiller, line 96: I like the example in general but we do not have an example of 'prefer_here' or 'here_if_load_lower_than', that would be also a common case that would be good to showcase.

tests/dsl/ptg/user-defined-functions/udf.jdf
bosilca: I want to test all forms for defining this `evaluate` capability so I will keep the inline form here and use the `always_here` later for the UD_HASH_STRUCT CPU task.

parsec/scheduling.c
bosilca: The incarnations array is packed, I cannot use the `chore_id` for figure out the device.

tests/dsl/ptg/user-defined-functions/udf.jdf
bosilca: This test checks the `evaluate` support in the infrastructure, they are not meant as an showcase for users. That being said, I don't think I would know how to do anything related to the load without access to internal information.

PR #570, opened 2023-08-28
Refactor GPU device to increase code factorization between the devices.

therault:
- Expose 8 functions that are device hardware-specific in the gpu_module_s: set_device, memcpy_async, event_query, event_record, memory_info, memory_allocate, memory_free, find_incarnation
 - Make 90% of the code in device_cuda_module.c use these functions and remain hardware-oblivious
 - Move all hardware-oblivious code in device_gpu.c
 - Keep cuda_module_init and cuda_module_fini, as well as other functions in the module API that were 90% hardware-specific in device_cuda_module.c


parsec/mca/device/device_gpu.c
```diff
@@ -354,3 +354,2268 @@ void dump_GPU_state(parsec_device_gpu_module_t* gpu_device)
...
+ * function progresses a single task, which is then returned as the
+ * out_task parameter.
+ *
+ * Beware: this function does not generate errors by itself, instead
+ * it propagates upward the return code of the progress function.
+ * However, by convention the error code follows the parsec_hook_return_e
+ * enum.
+ */
+static inline int
+progress_stream( parsec_device_gpu_module_t* gpu_device,
```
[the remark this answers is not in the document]
therault: This is a static inline function, it's not visible outside of this file. I kept the changes to a minimum, but let me know if you think the name is really a problem

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -6888,9 +6888,8 @@ static void jdf_generate_code_hook_gpu(const jdf_t *jdf,
     string_arena_free(info.sa);
 
     coutput("\n"
-            "  return parsec_%s_kernel_scheduler( es, gpu_task, dev_index );\n"
-            "}\n\n",
-            dev_lower);
+            "  return parsec_gpu_kernel_scheduler( es, gpu_task, dev_index );\n"
```
bosilca, line 6891: how is this working when multiple devices are present ? I think we need to name the scheduler appropriately, to allow support for multiple devices in same time (even if we do not currently have the cmake-configury nor the code generation capability for)

parsec/mca/device/cuda/device_cuda_component.c
```diff
@@ -141,7 +141,7 @@ static int device_cuda_component_query(mca_base_module_t **module, int *priority
                 PARSEC_CUDA_CHECK_ERROR( "(parsec_device_cuda_component_query) cuCtxEnablePeerAccess ", cudastatus,
                                          {continue;} );
                 source_gpu->super.peer_access_mask = (int16_t)(source_gpu->super.peer_access_mask | (int16_t)(1 <<
-                        target_gpu->cuda_index));
+                        target_gpu->super.super.device_index));
```
bosilca, line 144: This is a major change ! You repurpose the device index bitmask from a device naming scheme (`cuda_index` being the CUDA naming of the device) to a parsec naming scheme (`device_index` being the position in the parsec's array of devices).

However, I don't see any changes on the use of this `peer_access_mask` which point to something weird in this patch.

parsec/mca/device/cuda/device_cuda_component.c
```diff
@@ -192,9 +192,9 @@ static int device_cuda_component_register(void)
     (void)parsec_mca_param_reg_int_name("device_cuda", "max_streams",
                                         "Maximum number of Streams to use for the GPU engine; 2 streams are used for communication between host and device, so the minimum is 3",
                                         false, false, PARSEC_GPU_MAX_STREAMS, &parsec_cuda_max_streams);
-    (void)parsec_mca_param_reg_int_name("device_cuda", "sort_pending_tasks",
```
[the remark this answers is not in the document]
bosilca: do this make sense to exist per devide or be generic and apply to all devices equally ?

parsec/mca/device/device_gpu.c
```diff
@@ -14,6 +14,9 @@
 #include "parsec/execution_stream.h"
 #include "parsec/utils/argv.h"
 #include "parsec/parsec_internal.h"
+#include "parsec/scheduling.h"
+
+#define PARSEC_CUDA_DATA_COPY_ATOMIC_SENTINEL 1024
```
bosilca, line 19: This define moved in the `mca/device/device_gpu.c` but kept it's original name that includes  `CUDA` ?

parsec/mca/device/device_gpu.c
```diff
@@ -76,7 +80,7 @@ parsec_gpu_check_space_needed(parsec_device_gpu_module_t *gpu_device,
 void parsec_gpu_init_profiling(void)
 {
     if(parsec_gpu_profiling_initiated == 0) {
-        parsec_profiling_add_dictionary_keyword("cuda", "fill:#66ff66",
```
bosilca, line 79: Each accelerator should generate their own events, instead of generating generic-named events.

abouteiller: `device->name` will contain a string of the form "cuda(2)" where 2 is the dev_index.

parsec/mca/device/device_gpu.c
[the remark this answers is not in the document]
abouteiller: The rest of them are also static and have the full name, so for consistency, and also to help with gdb-ing.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
abouteiller: One way to revolve this would be to keep this generic function as a switchyard that would `switch(devices[dev_index]->type` and call the appropriate function based on this. That would keep the complexity in the generated code at a minimum.

parsec/mca/device/cuda/device_cuda_component.c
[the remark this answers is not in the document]
abouteiller: The sort function itself has moved to the generic device_gpu code, so it makes sense that the control for it would be at the same framework level.

For the more general question, should it apply per device type, no serious machine is having mixed device types, so its a low-importance feature. Should it be per-device (not by type, by devindex), maybe, but that would be a large refactor without proofs that it serves a purpose atm.

parsec/mca/device/device_gpu.c
bosilca: The device name is translated into the stream naming, but these events being unique they will be unique across all profiling streams, making them matcheable across execution domains. That make sense when we are talking about threads, but for devices it might be a stretch.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
bosilca: We could make the code more complicated instead of just adding the well-defined device type into the function name.  In fact, I think I would prefer a solution where the scheduler function is part of the device, so we will directly call `devices[dev_index]->scheduler(es, gpu_task, devices[dev_index])`

parsec/mca/device/cuda/device_cuda_component.c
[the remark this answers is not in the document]
bosilca: There is a runtime with serious funding, and application integration perspectives, that runs on exactly this "benefit". Why restricting the capability of parsec when it can easily be enriched.

Moreover, the sort function can be common, it is just the strategy that is device dependent.

parsec/mca/device/cuda/device_cuda_component.c
therault: I did change the initialization of `peer_access_mask` in `device_cuda_component.c:143`: we use `target_gpu->super.super.device_index` instead of `target_gpu->cuda_index`, so I believe the access and initialization are consistent.

In parsec master, `peer_access_mask` is 
 - initialized to 0 in `parsec_cuda_module_init`  in `device_cuda_module.c`
 - then computed in `device_cuda_component_query` in `device_cuda_component.c` after all modules have been loaded
 - then read in `parsec_cuda_data_stage_in` in `device_cuda_module.c`

In this PR, `parsec_cuda_data_stage_in` is moved in `device_gpu.c`, and I have changed for both the access and initialization what we use as index.



The user-facing MCA parameter  `device_cuda_nvlink_mask` is still using the CUDA index to exclude pairs of devices that could rely on NVLINK (so things have not changed here), and we only display `peer_access_mask` in debugging functions (the mask show parsec-device indices instead of cuda-indices, so it's not a 1-to-1 with the MCA parameter).



If this is not acceptable, to make `parsec_gpu_data_stage_in` device-type independent and keep it in `device_gpu.c`, I can extend the API of the devices to provide a function `bool device_to_device_direct(parsec_device_module_t *src, parsec_device_module_t *dst)` and each device will fill it up. But since `src` and `dst` might be of different types, the first test will be to return `false` if `src->type != dst->type`.

parsec/mca/device/cuda/device_cuda_component.c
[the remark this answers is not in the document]
therault: See commit b098633e7 if it implements the preferred approach?

parsec/mca/device/device_gpu.c
therault: Forgot to rename. See b098633e7

parsec/mca/device/device_gpu.c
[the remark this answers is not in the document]
therault: See 9cdf7db71

parsec/mca/device/device_gpu.h
```diff
@@ -120,8 +120,106 @@ struct parsec_gpu_task_s {
...
+ * @param fname the function name to look for
+ * @return address of the symbol that implements this function
+ */
+typedef void* (*parsec_device_gpu_find_incarnation_fn_t)(parsec_device_gpu_module_t* gpu_device, const char* fname);
+
 struct parsec_device_gpu_module_s {
     parsec_device_module_t     super;
+
+    /* This set of base functions is used by the GPU devices to implement their Device Management Functions */
+    parsec_device_gpu_set_device_fn_t       gpu_set_device;
```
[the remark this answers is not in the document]
therault: See 7b1a05b17 and 9cdf7db71

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
therault: Renamed in `parsec_device_kernel_scheduler` in 9cdf7db71.

When multiple accelerator devices are present, `dev_index` uniquely identifies a single device, and the code in `parsec_device_kernel_scheduler` is device-agnostic (the code uses the functions defined in the module to schedule tasklets, data movement and allocation).

Recursive and CPU devices do not call this function, as before.

bosilca: This approach limits the parsec runtime to a single enabled device per binary instance.

therault: Maybe the description of the PR was not clear . As part of the factorization effort, each accelerator device defines 8 functions (set_device, memcpy_async, event_query, event_record, memory_info, memory_allocate, memory_free, find_incarnation), and `parsec_device_kernel_scheduler` (and the other functions in `device_gpu.c`) use those 8 device-specific functions to call device-specific operations within a device-agnostic code.

So, `parsec_device_kernel_scheduler` in `device_gpu.c` is the generic part of the existing code that does not depend on the device type, and when we do device-specific things (e.g.  adding a CUDA event in a CUDA stream, or appending a Level Zero fence to a Level Zero command queue), we call one of the module-specific functions (gpu_module->event_record for that example).

This relies on the user / DSL putting the right thing in the `gpu_task_t` passed to `parsec_device_kernel_scheduler`: the `submit`, `stage_in`, and `stage_out` need to target the right type of device that corresponds to the `dev_index` in the call.

I believe this code should allow to have a HIP, Level Zero and CUDA device functionning simultaneously, but I may be missing something?

bosilca: it is clear. However, the only reason you need to define a `set_device` function is to be able to create a unified parsec_device_kernel_scheduler and therefore define a unique behavior across all devices. If instead you add another member to the device structure `_scheduler`, then you can in addition to having the unified behavior you describe, have specialized behavior per device (hint recursive and CPU).

bosilca (approved):

PR #572, opened 2023-09-06
HAVE_PEER_ACCESS is always present in all relevant versions of CUDA or

abouteiller:
Issue: ROCM would not use peer-access operators because we don't have a cmake test for the rocm-specific version of the have_peer_access function.

However, HAVE_PEER_ACCESS is always present in all relevant versions of CUDA or ROCM, so no need to cmake detect it, just require minimum cuda and rocm versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -696,18 +696,10 @@ int main(int argc, char *argv[]) {
         message(STATUS "Using PGI internal CUDA SDK in ${CUDAToolkit_ROOT}")
       endif(_path_pgcuda)
     endif (_match_pgi)
-    find_package(CUDAToolkit)
+    find_package(CUDAToolkit 4)
     set(PARSEC_HAVE_CUDA ${CUDAToolkit_FOUND} CACHE BOOL "True if PaRSEC provide support for CUDA")
     if (CUDAToolkit_FOUND)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -725,7 +717,7 @@ int main(int argc, char *argv[]) {
     # This is kinda ugly but the PATH and HINTS don't get transmitted to sub-dependents
     set(CMAKE_SYSTEM_PREFIX_PATH_save ${CMAKE_SYSTEM_PREFIX_PATH})
     list(APPEND CMAKE_SYSTEM_PREFIX_PATH /opt/rocm)
```
bosilca, line 719: I don't think this explicit path should be here.

devreal (approved):
LGTM

CMakeLists.txt
abouteiller: Removing it would break the detection of the system install rocm (as installed from the RPMs distributed by AMD).

bosilca: Why cant you use PATHS or HINTS ? According to the cmake [documentation](https://cmake.org/cmake/help/latest/command/find_package.html) that's the preferred way.

abouteiller: Cmake 'feature/bug': unlike ROCM_ROOT, they do not get propagated to the internal find_packages that are done within FindRocm and that results in an inconsistent/incomplete detection of internal features.

bosilca (approved):

PR #573, opened 2023-09-06
Mishler pins tasks appear

DSMishler:
Following up on the discussion on Monday, these are the changes I was inquired about pushing. I would request someone take a minute or two and review these because I am personally not sure what everything does here and wondering if all of my changes to the compiler are necessary - they might mess up the software in other places.

### DIFF
--- a/parsec/mca/pins/pins.c
+++ b/parsec/mca/pins/pins.c
@@ -38,11 +38,13 @@ void parsec_pins_instrument(struct parsec_execution_stream_s* es,
      * profiling array only generated if at least one task requires profiling,
      * so must check it.
      */
+#ifdef PARSEC_PROF_TRACE
     if((NULL != task) && 
        (NULL != task->taskpool->profiling_array) &&
        (-1 == task->taskpool->profiling_array[START_KEY(task->task_class->task_class_id)])) {
         return;
     }
+#endif
 
 
     parsec_pins_next_callback_t* cb_event = &es->pins_events_cb[method_flag];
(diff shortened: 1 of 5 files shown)


parsec/mca/pins/pins.c
```diff
@@ -32,6 +34,18 @@ void parsec_pins_instrument(struct parsec_execution_stream_s* es,
                             parsec_task_t* task)
 {
     assert( method_flag < PARSEC_PINS_FLAG_COUNT );
+    /**
```
therault, line 37: This ties PINS with the profiling system in a new way: if profiling is enabled at configure time, only tasks with a profiling key will instrument that PINS event, if profiling is disabled at configure time, all tasks will instrument that PINS event.

Initially, PINS could be independent from the profiling system. Some PINS instruments (e.g. task_profiler) would do things only if profiling was enabled, but others (e.g. `print_steals` or `iterator_checkers`) would do other kind of printing/checking that does not depend on the profiling.

PRO of this approach: we don't even record the callback function, so overheads are 0 for tasks that are no-profile

CON of this approach: we lose the feature of doing other things than profiling with PINS.

I would advocate fixing the issue (error when no-profile tasks are used) without doing this, unless we also decide to drop the other PINS modules?

therault (commented):
@bosilca please look at my comment: should we fix the issue in another way (i.e. probably checking if `-1 == task->taskpool->profiling_array[START_KEY(task->task_class->task_class_id)]` in each of the callbacks of `task_profiler`? Or should we disable PINS event for all no-profile tasks, even if the PINS module is not profiling per-se?

bosilca:
The correct solution to this issue is proposed in #594.

PR #575, opened 2023-09-30
Prevent race condition in accelerator copies management

bosilca:
Prevent the owner device from repurposing a GPU copy while another GPU is planning to use it as a candidate for a d2d transfer.

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -27,6 +27,8 @@
 #include <cuda.h>
 #include <cuda_runtime_api.h>
 
+#define PARSEC_CUDA_DATA_COPY_ATOMIC_SENTINEL 1024
+
 static int parsec_cuda_data_advise(parsec_device_module_t *dev, parsec_data_t *data, int advice);
 /**
  * According to
@@ -874,8 +876,8 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
     parsec_gpu_data_copy_t* temp_loc[MAX_PARAM_COUNT], *gpu_elem, *lru_gpu_elem;
     parsec_data_t* master, *oldmaster;
     const parsec_flow_t *flow;
-    int i, j, data_avail_epoch = 0;
-    parsec_gpu_data_copy_t *gpu_mem_lru_cycling;
+    int i, j, data_avail_epoch = 0, copy_readers_update = 0;
+    parsec_gpu_data_copy_t *gpu_mem_lru_cycling = NULL;
     parsec_device_gpu_module_t *gpu_device = &cuda_device->super;
 
 #if defined(PARSEC_DEBUG_NOISIER)
@@ -888,7 +890,6 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
      * corresponding data on the GPU available.
      */
     for( i = 0; i < this_task->task_class->nb_flows; i++ ) {
-        gpu_mem_lru_cycling = NULL;
         flow = gpu_task->flow[i];
         assert( flow && (flow->flow_index == i) );
 
@@ -902,7 +903,6 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
         if (this_task->data[i].data_in == NULL)
             continue;
 
-	/* BEWARE: here we are using the CPU copy as an input */
         master   = this_task->data[i].data_in->original;
         parsec_atomic_lock(&master->lock);
         gpu_elem = PARSEC_DATA_GET_COPY(master, gpu_device->super.device_index);
@@ -916,10 +916,10 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
                                  flow->name, i, gpu_elem,
                                  gpu_elem->data_transfer_status == PARSEC_DATA_STATUS_UNDER_TRANSFER ? " [in transfer]" : "");
             if ( gpu_elem->data_transfer_status == PARSEC_DATA_STATUS_UNDER_TRANSFER ) {
-	      /* We might want to do something special if the data is under transfer, but in the current
-	       * version we don't need to because an event is always generated for the push_in of each
-	       * task on the unique push_in stream.
-	       */
+                /* We might want to do something special if the data is under transfer, but in the current
+                 * version we don't need to because an event is always generated for the push_in of each
+                 * task on the unique push_in stream.
+                 */
             }
             parsec_atomic_unlock(&master->lock);
             continue;
@@ -932,27 +932,29 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
                              gpu_device->super.name, task_name,
                              gpu_elem, gpu_task->flow_nb_elts[i], gpu_elem->super.super.obj_reference_count, master);
         gpu_elem->flags = PARSEC_DATA_FLAG_PARSEC_OWNED | PARSEC_DATA_FLAG_PARSEC_MANAGED;
-      malloc_data:
+    malloc_data:
         assert(0 != (gpu_elem->flags & PARSEC_DATA_FLAG_PARSEC_OWNED) );
         gpu_elem->device_private = zone_malloc(gpu_device->memory, gpu_task->flow_nb_elts[i]);
         if( NULL == gpu_elem->device_private ) {
 #endif
 
-          find_another_data:
+        find_another_data:
+            copy_readers_update = 1;
+            temp_loc[i] = NULL;
             /* Look for a data_copy to free */
             lru_gpu_elem = (parsec_gpu_data_copy_t*)parsec_list_pop_front(&gpu_device->gpu_mem_lru);
             if( NULL == lru_gpu_elem ) {
                 /* We can't find enough room on the GPU. Insert the tiles in the begining of
-                 * the LRU (in order to be reused asap) and return without scheduling the task.
+                 * the LRU (in order to be reused asap) and return with error.
                  */
-	    release_temp_and_return:
+            release_temp_and_return:
 #if defined(PARSEC_DEBUG_NOISIER)
                 PARSEC_DEBUG_VERBOSE(2, parsec_gpu_output_stream,
                                      "GPU[%s]:%s:\tRequest space on GPU failed for flow %s index %d/%d for task %s",
                                      gpu_device->super.name, task_name,
                                      flow->name, i, this_task->task_class->nb_flows, task_name );
 #endif  /* defined(PARSEC_DEBUG_NOISIER) */
-                for( j = 0; j < i; j++ ) {
+                for( j = 0; j <= i; j++ ) {
                     /* This flow could be a control flow */
                     if( NULL == temp_loc[j] ) continue;
                     /* This flow could be non-parsec-owned, in which case we can't reclaim it */
@@ -970,13 +972,20 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
                 parsec_atomic_unlock(&master->lock);
                 return PARSEC_HOOK_RETURN_AGAIN;
             }
-
             PARSEC_LIST_ITEM_SINGLETON(lru_gpu_elem);
             PARSEC_DEBUG_VERBOSE(20, parsec_gpu_output_stream,
                                  "GPU[%s]:%s: Evaluate LRU-retrieved CUDA copy %p [ref_count %d] original %p",
                                  gpu_device->super.name, task_name,
                                  lru_gpu_elem, lru_gpu_elem->super.super.obj_reference_count,
                                  lru_gpu_elem->original);
+
+            if( gpu_mem_lru_cycling == lru_gpu_elem ) {
+                PARSEC_DEBUG_VERBOSE(2, parsec_gpu_output_stream,
+                                     "GPU[%s]: Cycle detected on allocating memory for %s",
+                                     gpu_device->super.name, task_name);
+                temp_loc[i] = lru_gpu_elem;  /* save it such that it gets pushed back into the LRU */
+                goto release_temp_and_return;
+            }
 
             /* If there are pending readers, let the gpu_elem loose. This is a weak coordination
              * protocol between here and the parsec_gpu_data_stage_in, where the readers don't necessarily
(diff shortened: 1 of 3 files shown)


parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -1385,7 +1396,15 @@ parsec_gpu_data_stage_in( parsec_device_cuda_module_t* cuda_device,
...
-            parsec_atomic_fetch_inc_int32( &candidate->readers );
+            int readers = parsec_atomic_fetch_inc_int32(&candidate->readers);
+            if( readers < 0 ) {
+                PARSEC_DEBUG_VERBOSE(10, parsec_gpu_output_stream,
+                                     "GPU[%s]:\tCandidate %p [ref_count %d] on CUDA device %d is being repurposed by owner device. Looking for another candidate",
+                                     gpu_device->super.name, candidate, candidate->super.super.obj_reference_count, target->cuda_index);
+                /* We are trying to use a candidate that is repurposed by the owner device. Let's find another one */
+                parsec_atomic_fetch_add_int32(&candidate->readers, -1);
+                continue;
+            }
```
devreal, line 1407: After the increment we also need to check that the candidate has not been detached?

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -1131,9 +1137,15 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
...
+
+        /* Do not push it back into the LRU for now to prevent others from discovering
+         * this copy and trying to acquire it. If we fail to find all the copies we need
+         * we will push it back in the release_temp_and_return, otherwise they will become
+         * available once properly updated.
+         */
         gpu_elem->coherency_state = PARSEC_DATA_COHERENCY_INVALID;
-        gpu_elem->version = 0;
+        gpu_elem->version = UINT_MAX;  /* scrap value for now */
+        gpu_elem->readers = 0;         /* safe to set to no access */
```
devreal, line 1148: I'm not sure this is safe: there might be someone looking at the copy seeing negative readers and decrementing by one. I think this should be an atomic increment by `INT_MAX` so that eventually `readers` reaches 0 in all cases.

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -1052,19 +1046,31 @@ parsec_gpu_data_reserve_device_space( parsec_device_cuda_module_t* cuda_device,
...
+                                         "GPU[%s]:%s: Push back LRU-retrieved CUDA copy %p [readers %d, ref_count %d] original %p : Concurrent accesses",
+                                         gpu_device->super.name, task_name,
+                                         lru_gpu_elem, lru_gpu_elem->readers, lru_gpu_elem->super.super.obj_reference_count,
+                                         lru_gpu_elem->original);
                     parsec_atomic_unlock( &oldmaster->lock );
                     goto find_another_data;
                 }
+                /* Check if this copy is the last dangling reference to the oldmaster. This is safe to do as we own one of the data refcounts. */
                 int do_unlock = oldmaster->super.obj_reference_count != 1;
                 parsec_data_copy_detach(oldmaster, lru_gpu_elem, gpu_device->super.device_index);
```
devreal, line 1072: This needs to be guaranteed sequenced after the CAS so we need a write memory barrier before the detach.

devreal (changes requested):

parsec/mca/device/cuda/device_cuda_module.c
therault: Also, `parsec_data_copy_detach()` doesn't update `data->owner_device`, so this detach can lead to a situation where the `data->device_copies[data->owner_device]` is `NULL`. Is that a behavior we want to support?

parsec/mca/device/cuda/device_cuda_module.c
therault: Or an atomic increment by `-INT_MIN` ? (`INT_MIN` is probably not equal to `INT_MAX`)

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -1295,7 +1306,7 @@ parsec_gpu_data_stage_in( parsec_device_cuda_module_t* cuda_device,
     transfer_from = parsec_data_start_transfer_ownership_to_copy(original, gpu_device->super.device_index, (uint8_t)type);
 
     if( PARSEC_FLOW_ACCESS_WRITE & type && gpu_task->task_type != PARSEC_GPU_TASK_TYPE_PREFETCH ) {
-        gpu_elem->version++;  /* on to the next version */
+        gpu_elem->version = candidate->version + 1;  /* on to the next version */
```
therault, line 1309: I agree we need to do this here, but I think we also need to fix at line 1487 (before the patch) / 1506 (after the patch): we do `gpu_elem->version = candidate->version;` when the transfer (if a transfer is needed) is done, and the version of `candidate` is one less than what we want, and this will overwrite the effect of this action here.

This doesn't typically fail with GEMM or POTRF as benchmarks / tests because they only send the result back to the host memory once, and then we discard / release the GPU copy, so the fact that the version is wrong doesn't trigger a problem, but if you remove the release_memory from the test this should create a problem.

parsec/mca/device/cuda/device_cuda_module.c
bosilca: At this point all the protection that could be offered was offered by the atomic update protocols. Now, I know this code remains unsafe for extreme cases, but the only way to fix it is to add a lock. Adding yet another atomic operation will not help.

If another thread is trying to use this copy as a candidate for d2d, and that thread is delayed, it might see the readers set to 0 and assume everything is fine and then move ahead and use the copy. Hopefully that will not happen because the version will be incorrect and the coherency is invalid.

parsec/mca/device/cuda/device_cuda_module.c
bosilca: No, there is no need for write memory barrier. There is a single owner of the copy, the thread that will do the CAS and the detach, and the same that will eventually free the copy. Every other threads is limited to updates of the reader field, if they want to manipulate the copy more, they do it via a message to the owner thread.

I don't see why changing the `owner_device` is needed here, we only change the data_t this copy is associated with, the `owner_device` will remain the same. The detach will clear the `origin` so `data->device_copies[data->owner_device]` will segfault because `data` will be NULL.

therault: In `parsec_data_copy_detach` we do `    data->device_copies[device] = copy->older;` (where `data` is `copy->original` before we set `copy->original` to `NULL`. If `copy->older` is NULL, then that `data_t` has a `device_copies[owner_device]` set to NULL?

bosilca: That is a bigger problem overall. 
1. `NULL == copy->older` we have the owner_device pointing to a NULL data. But, we don't know what is legitimately the good version of the remaining copies, so in the generic code we are unable to redirect the `owner_device` to the correct version.
2. `NULL != copy->older` we are setting the `owner_device` to point to an older version of the data that might exists on the device. Not clear why that version would be the correct one.

therault:
I get some asserts triggering with dplasma using this branch for parsec.

dplasma: master (749c912911ed970ef292d7b1d49e6312dd156485)
parsec: latest commit of this branch (4a4fad091dc80a4ec05a2458ffb88454cc702073)

Compile mode: CMAKE_BUILD_TYPE=Debug, PARSEC_DEBUG_NOISIER=ON and PARSEC_DEBUG_PARANOID=ON

Machine: xsdk (with CUDA enabled)

test: `./tests/testing_dgemm -g 1 -N 1024 -K 1024 -M 1024 -t 1024 -x 0 --nruns 1`

Symptom:
testing_dgemm: /home/herault/dplasma/parsec/parsec/data.c:418: parsec_data_start_transfer_ownership_to_copy: Assertion `(!transfer_required) || (data->device_copies[valid_copy]->version >= copy->version)' failed.

Context: trying to stage_in flow 0 of GEMM(0, 0, 0): Flows 0, 1 and 2 of GEMM(0, 0, 0) have been allocated on the device in `parsec_gpu_data_reserve_device_space` and the version of the new copies set to `UINT_MAX` at line 1147. Then the code reaches line 1306 in stage_in, calls `parsec_data_start_transfer_ownership_to_copy` and the assert fails, because the device's copy has version UINT_MAX, which is never smaller than the CPU copy version. Or at least, that's what I understand (and yes, I reckon I don't understand the GPU code... :) ).

parsec/mca/device/cuda/device_cuda_module.c
devreal: That's the path if `readers < 0` (the case where we don't use the copy). But what about the alternative path (anything below this block)? We could encounter a copy with `readers == 1` that is detached from the data we're looking for, no?

bosilca: I don't think this scenario is possible anymore. Either the copy is not repurposed in which case the readers is `> 0`, or is currently repurposed by the owner in which case the atomicity guarantees that either readers is positive and the current thread can use it, is negative and the owner will get it from there.

However, as my comment in another of the comments states, if the current thread is significantly delayed at considering the copy, the owner could have completed the transfer and the copy is now associated with a different data_t. Without a lock I don't know how to avoid that.

devreal: I think it is sufficient to check for the data after the increment (properly sequenced with a read memory barrier). There can be three cases:

1) `readers >= 0` and the right data: good, can use
2) `readers < 0`: bad, cannot use (another thread is in the process of detaching the copy)
3) `readers >= 0` and a different data: bad, cannot use (another thread has detached the copy and reset readers to 0)

Number 3 would be covered by checking for the data_t after the increment if the readers looked right, i.e., after this block. That is also why [the detach](https://github.com/ICLDisco/parsec/pull/575/files#diff-a42087f8b140e7fc55fd74480d49a4334d21ff2810f51ca57e0d1955534286e8R1072) needs to be properly sequenced with a write memory barrier (so that the detach cannot occur before the atomic operation on `readers`).

parsec/mca/device/cuda/device_cuda_module.c
therault: My problem is I don't understand (neither in the code before the PR or inside the PR) what is supposed to increase the version of the GPU copy to what I believe it should be.

I did a run (or the PR, at commit fc82786a2c791ec65c2bc3f0b01d1086229b3fbf) with M=1 tile, N=1 tile, K=2 tiles, a single GPU under gdb.

I break on `parsec_cuda_kernel_scheduler` to see all kernels passed to the GPU, and on `parsec_gpu_data_reserve_device_space`. When I'm in kernel scheduler, I print the task name. When I'm in reserve_device_space, I go until the GPU copy for C is allocated (line 927), then I install a watchpoint on any change of this copy's version.

Here is what I see:
  - we start GEMM(0, 0, 0), we allocate the GPU copy for C and set the version to 0 (inside the allocation)
  - version goes to UINT_MAX at line 1149 according to the patch
  - it is then set to 2 at line 1311, because it's a write (C) and we take candidate (CPU copy)'s version + 1. transfer_from is 0, everything is logic
  - next step, it is reset to 1 at line 1512 after we have started the copy. You seem to expect that, OK, why not.
  - then it remains unmodified until we start scheduling GEMM(0, 0, 1) on the GPU. But definitely here, its version should be 2. Shouldn't it? The CPU version is 2 (has been updated by the DSL at the completion of the task), so it appears like if the GPU copy is too old (which it isn't).
  - then we reach the call to `parsec_data_start_transfer_ownership_to_copy` at line 1308:
     - original->owner_device is set to 2, and the GPU copy is in the state SHARED.
     - there is no other copy with the state OWNED, and all other copies are in the state SHARED, so we never check if another copy has a version higher than the GPU version.
     - that explains why we don't consider a new transfer is necessary, and we return -1
  - then we set the GPU copy version to 3 directly, by line 1311 (candidate->version+1)
  - we don't schedule the transfer, so we don't reset the version to candidate->version

It works, but I'm surprised that this is the expected behavior. I would have expected the GPU copy to get the version 2 when the transfer is complete, and the transfer_ownership to return -1 because the GPU copy has the highest version. At the time transfer_ownership returns -1, it looks to me it returns -1 for the wrong reason.

therault:
> I get some asserts triggering with dplasma using this branch for parsec.
> 
> dplasma: master (749c912911ed970ef292d7b1d49e6312dd156485) parsec: latest commit of this branch ([4a4fad0](https://github.com/ICLDisco/parsec/commit/4a4fad091dc80a4ec05a2458ffb88454cc702073))
> 
> Compile mode: CMAKE_BUILD_TYPE=Debug, PARSEC_DEBUG_NOISIER=ON and PARSEC_DEBUG_PARANOID=ON
> 
> Machine: xsdk (with CUDA enabled)
> 
> test: `./tests/testing_dgemm -g 1 -N 1024 -K 1024 -M 1024 -t 1024 -x 0 --nruns 1`
> 
> Symptom: testing_dgemm: /home/herault/dplasma/parsec/parsec/data.c:418: parsec_data_start_transfer_ownership_to_copy: Assertion `(!transfer_required) || (data->device_copies[valid_copy]->version >= copy->version)' failed.
> 
> Context: trying to stage_in flow 0 of GEMM(0, 0, 0): Flows 0, 1 and 2 of GEMM(0, 0, 0) have been allocated on the device in `parsec_gpu_data_reserve_device_space` and the version of the new copies set to `UINT_MAX` at line 1147. Then the code reaches line 1306 in stage_in, calls `parsec_data_start_transfer_ownership_to_copy` and the assert fails, because the device's copy has version UINT_MAX, which is never smaller than the CPU copy version. Or at least, that's what I understand (and yes, I reckon I don't understand the GPU code... :) ).

this is solved

parsec/mca/device/cuda/device_cuda_module.c
bosilca: This is not what I expect to see here. But, that is another problem for another PR, here we are just fixing the race condition between accelerator copies management.

parsec/mca/device/cuda/device_cuda_module.c
devreal: This assignment conflicts with the roll back a potential reader would do here: https://github.com/ICLDisco/parsec/pull/575/files#diff-a42087f8b140e7fc55fd74480d49a4334d21ff2810f51ca57e0d1955534286e8R1415

The sequence would be: Thread A sets `readers` to `INT_MIN`; Thread B increments `readers` and finds it negative; Thread A detaches the copy and sets `readers` to 0; Thread B does the rollback by atomically decrementing `readers`. The result would be `readers == -1`. I don't think that should be valid. 

Instead of assigning 0 we should atomically add `-INT_MIN` which would guarantee `readers >= 0`.

devreal (changes requested):

therault (approved):
LGTM

PR #577, opened 2023-10-09
Fix/rocm detect and unknown device warning

abouteiller:
1. fixes detection of Rocm not working
2. add AMD device detection
3. add cuda 8.5--9 device detection
4. removes non-conditional warnings about unknown device
5. roll-in unknown device warnings to the relevant places (debug when computing default-time-estimate, and the SHOW_CAPS option)

### DIFF
--- a/parsec/mca/device/device.h
+++ b/parsec/mca/device/device.h
@@ -143,6 +143,7 @@ struct parsec_device_module_s {
     int64_t   gflops_fp32;  /**< Number of single precision operations per nanosecond (or gflops/s) */
     int64_t   gflops_fp64;  /**< Number of double precision operations per nanosecond (or gflops/s) */
     int64_t   gflops_tf32;  /**< Number of tensor operations per nanosecond (or gflops/s) */
+    uint8_t   gflops_guess; /**< True if the device is not 'known' which entails that the 'gflops' rates have been populated with fallback (arbitrary) values. */
     int64_t   time_estimate_default; /**< An estimate of the time to execute on that device a task that would take 1ns using the aggregate power of all devices. This is the default time_estimate if none is user-set. */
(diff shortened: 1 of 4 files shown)


devreal (approved):
LGTM

parsec/mca/device/device.h
```diff
@@ -143,6 +143,7 @@ struct parsec_device_module_s {
     int64_t   gflops_fp32;  /**< Number of single precision operations per nanosecond (or gflops/s) */
     int64_t   gflops_fp64;  /**< Number of double precision operations per nanosecond (or gflops/s) */
     int64_t   gflops_tf32;  /**< Number of tensor operations per nanosecond (or gflops/s) */
+    uint8_t   gflops_guess; /**< True if the device is not 'known' which entails that the 'gflops' rates have been populated with fallback (arbitrary) values. */
```
bosilca, line 146: Move this at the end of the structure to allow the compiler to pack it more efficiently.

devreal: This is a good point to move it to: https://github.com/ICLDisco/parsec/pull/577/files#diff-3a7133ac2f9430b1fc1cde3edef7f3b7b6d060e86ec881967ae51f585a781c21R142

bosilca: The link does not work for me, but I am referring to the end of the structure where the 3 other uint8_t fields are located.

abouteiller:
detection by device name does not work on Frontier (name is empty there). It's a bug that has been reported since Rocm 4 so we need to find another way.

bosilca:
There was no need to reorder the entire struct, just move the `gflops_guess` at the correct place. Also, one of the commits is not signed and as a result the PR cannot be merged.

PR #578, opened 2023-10-09
Pick a stable spack branch

bosilca:
Add RUNNER_SPACK_BRANCH to allow an updated spack branch to be used instead of the curretn default (e4s-23.08).


devreal (approved):
LGTM

PR #581, opened 2023-10-15
Move MAX_DEP_IN_COUNT, MAX_DEP_OUT_COUNT, MAX_LOCAL_COUNT, and MAX_PA…

QingleiCao:
Move MAX_DEP_IN_COUNT, MAX_DEP_OUT_COUNT, MAX_LOCAL_COUNT, and MAX_PARAM_COUNT to CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,6 +242,28 @@ option(PARSEC_PROFILING_USE_HELPER_THREAD
   "Use a Helper Thread to create the profile files" ON)
 mark_as_advanced(PARSEC_PROFILING_USE_HELPER_THREAD)
 
+# A set of constants defining the capabilities of the underlying runtime
+if(MAX_DEP_IN_COUNT)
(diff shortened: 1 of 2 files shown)


QingleiCao:
Not sure there is a better way.

therault (changes requested):
I think this approach is too fragile: structures that use these constants for array members are exposed to the users (via `parsec_internal.h`). If a user compiles any of his files with a `MAX__LOCAL_COUNT` or `MAX_PARAM_COUNT` different from what the parsec library was compiled with, the structures will not be aligned, debugging will be hard, etc...

It might not be important for the highest-level users who use parsec via dplasma, but for users like TTG that develop DSLs on top of parsec, this will be very error prone.


To address Issue 579, I would prefer moving these definitions from `parsec_config_bottom.h` into `parsec_config.h.in` and have CMake define them, but at least the generated `parsec_options.h` shipped with the compiled parsec library would expose the correct values only.

bosilca:
#583 presents a more flexible solution to this problem.

PR #582, opened 2023-10-16
Add an option to skip HWLOC compat run

### LINKED ISSUE #580: Issue of MPI_AND_HWLOC_COMPATIBLE on some systems
On some systems, this check cannot be done on the login node. Here are the outputs on Summit:
```
...and run output:
HWLOC headers version: 20500.
Calling `hwloc_get_api_version()`...
HWLOC library version: 20500.
Calling `MPI_Init()`...
Error: OMPI_COMM_WORLD_RANK is not set in environment like it should be.
Error checking ibm license.
```

bosilca:
By default we want to check that the selected MPI and HWLOC are compatible, but in some cases we cannot run MPI applications on the compile node. Add an option to allow the user to skip this check, and instead assume that it was successful.

Fixes #580

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,7 +120,6 @@ mark_as_advanced(BUILD_TOOLS)
 option(BUILD_PARSEC
        "Compile the PaRSEC framework" ON)
 mark_as_advanced(BUILD_PARSEC)
-
 ### Misc options
 option(BUILD_SHARED_LIBS
     "Build shared libraries" ON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -631,7 +633,7 @@ if( BUILD_PARSEC )
     list(APPEND EXTRA_LIBS ${HWLOC_LIBRARIES})
     list(APPEND EXTRA_INCLUDES ${HWLOC_INCLUDE_DIRS})
     set (PARSEC_PKG_REQUIRE "hwloc")
-    if(MPI_C_FOUND AND NOT CMAKE_CROSSCOMPILING)
+    if(MPI_C_FOUND AND NOT CMAKE_CROSSCOMPILING AND MPI_HWLOC_COMPAT_CHECK)
```
therault, line 636: I'm not sure if this reads `MPI_C_FOUND AND (NOT CMAKE_CROSSCOMPILING) AND MPI_HWLOC_COMPAT_CHECK` or `MPI_C_FOUND AND NOT (CMAKE_CROSSCOMPILING AND MPI_HWLOC_COMPAT_CHECK)`. I guess CMake does the right thing, but this is CMake... I also guess that could change any time with a policy change ;)

Could we use explicit parentheses here? Or would that be too heavy?

therault (approved):
I'm suggesting a more explicit parenthesing of the test, but if you're sure CMake will never do the wrong thing, I'm fine with this (but I am not confident in CMake to keep doing the right thing).

PR #584, opened 2023-10-23
[python] do not build python support unless pandas is available

evaleev:
although only cython is needed, python module is unusable without pandas, so makes sense to consider pandas a hard prerequisite and skip building python module unless pandas is available


bosilca:
I thought we were installing it automatically in our env. Let me check on this before approving.

bosilca (approved):

PR #590, opened 2023-11-03
Increment version of RW copies on reset

devreal:
This is needed for TTG because otherwise the rw device copy version will remain the same as the host, which may cause the copy to be selected. This may cause issues with tasks that return `NEXT` as multiple copies will have incremented versions. A solution is to disallow returning `NEXT` from the task hook after #569 has been brought in.


abouteiller:
Needed after #648

PR #594, opened 2023-11-10
Skip profiling for task classes without profiling information

DSMishler:
This PR fixes the issues related to profiling of tasks without profiling information. We decided to allow PINS modules to generate interesting information even for non-profilable tasks. Thus, we had to make sure that PINS module generating profile information would skip tasks marked without profiling (property `profile=off` in PTG as an example).

Added -2 as another value for the profiling array. `profiling_array[0] == -2` signifies that nothing has been initialized yet.

Replacing both #561 and #573.

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -861,11 +861,22 @@ static char *dump_profiling_init(void **elem, void *arg)
     int nb_locals;
     string_arena_t *profiling_convertor_params;
 
-    if( !profile_enabled(f->properties) ) {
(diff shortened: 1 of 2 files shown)


therault (approved):
LGTM

devreal (changes requested):
@DSMishler I can't judge the fix, but: could you please change the commit message to something more descriptive? And if it is two independent changes maybe split them? Also, can we give `-2` a name to avoid magic numbers?

bosilca (approved):

bosilca:
@DSMishler I merged the PR that fixes the profiling tests. Please rebase this PR to current HEAD and wait for the CI to complete a test before merging.

PR #595, opened 2023-11-14
CMAKE: Bring back checks for atomic CAS

devreal:
Testing for atomic_is_lock_free() may not be enough as it may not require `-latomic` that is otherwise needed for 128bit atomics.

Reported by @DSMishler.

### DIFF
--- a/cmake_modules/CheckAtomicIntrinsic.cmake
+++ b/cmake_modules/CheckAtomicIntrinsic.cmake
@@ -53,7 +53,9 @@ if( SUPPORT_C11 AND PARSEC_ATOMIC_USE_C11_ATOMICS )
     #include <stdint.h>
     #include <stdatomic.h>
     int main(void) {
-        int32_t where = 0;
+        int32_t where = 0, expected = 0;
+        if( !atomic_compare_exchange_strong( (_Atomic __int32_t*)&where, &expected, 1 ) )
+            return -1;
         return (atomic_is_lock_free(&where) ? 0 : 1);
     }
     " PARSEC_ATOMIC_USE_C11_32)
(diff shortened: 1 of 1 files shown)


cmake_modules/CheckAtomicIntrinsic.cmake
```diff
@@ -53,7 +53,9 @@ if( SUPPORT_C11 AND PARSEC_ATOMIC_USE_C11_ATOMICS )
     #include <stdint.h>
     #include <stdatomic.h>
     int main(void) {
-        int32_t where = 0;
+        int32_t where = 0, expected = 0;
+        if( !atomic_compare_exchange_strong( (_Atomic __int32_t*)&where, &expected, 1 ) )
+            return -1;
```
therault, line 58: The original code returns 0 or 1. This patch makes it return 0 (success), 1 or -1. I guess 1 and -1 can be conflated to failure in both cases. POSIX specifies only that the 8 lower bits are meaningful (https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/V2_chap02.html#tag_15_13), so -1 is not clear.

May I suggest we use EXIT_SUCCESS and EXIT_FAILURE for all return/exit values, since those are POSIX compliant?

therault (changes requested):

bosilca:
In what case this is necessary ? On my tests when using the libatomic is necessary, the associated intrinsics are automatically made available, and `atomic_is_lock_free` return the expected outcome.

devreal:
I can only guess here based on @DSMishler issue: `atomic_is_lock_free` is a macro that may not need to call functions in `libatomic` even when CAS on 128bit integer would require `libatomic` to be present. So we never detect that we need to link `-latomic`.

Either way, these checks don't hurt and may help if a weird compiler may need `-latomic` for all atomic operations, but not for `atomic_is_lock_free`.

cmake_modules/CheckAtomicIntrinsic.cmake
devreal: I thought about harmonizing them but then I realized that it might be helpful to have different return values for the two error cases. That way we can distinguish (in the CMake log) whether the operation is not lock-free or whether we failed to execute it in the first place...

abouteiller: return 1 and 2 then, that's probably more clear than 1 and 254

therault (approved):

abouteiller (approved):

PR #604, opened 2023-12-04
Make sure the w2r task has a stageout set

devreal:
AFAIU, the w2r uses the stage-out mechanism to move data is meant to be evicted from the device. It should have a stage-out function set to do that.


therault (approved):

devreal:
Many moons ago there was this commit that removed the stage_out: https://github.com/ICLDisco/parsec/commit/e8a383fa9f5760d5743dc13778b4922b7a443b34 

Commit says: 

```
 Make sure we don't call stage_in/stage_out on w2r tasks, they don't have one
```

I'm confused.

bosilca:
That commit was incorrect. It might have started from a good intention, make sure that w2r tasks do not have a stage_in, but it went haywire on the stage_out.

PR #605, opened 2023-12-04
Consistently use size_t for nb_elts in data and flows

devreal:
`size_t` should be used to avoid overflows, since `nb_elts` can represent a number of bytes.

### DIFF
--- a/parsec/mca/device/device_gpu.c
+++ b/parsec/mca/device/device_gpu.c
@@ -89,7 +89,13 @@ void parsec_device_init_profiling(void)
                                                 PARSEC_PROFILE_DATA_COLLECTION_INFO_CONVERTOR,
                                                 &parsec_gpu_prefetch_key_start, &parsec_gpu_prefetch_key_end);
         parsec_profiling_add_dictionary_keyword("gpu_mem_alloc", "fill:#FF66FF",
-                                                sizeof(int64_t), "size{uint64_t}",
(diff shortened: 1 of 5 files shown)


bosilca (approved):

parsec/mca/device/device_gpu.c
```diff
@@ -89,7 +89,11 @@ void parsec_device_init_profiling(void)
                                                 PARSEC_PROFILE_DATA_COLLECTION_INFO_CONVERTOR,
                                                 &parsec_gpu_prefetch_key_start, &parsec_gpu_prefetch_key_end);
         parsec_profiling_add_dictionary_keyword("gpu_mem_alloc", "fill:#FF66FF",
-                                                sizeof(int64_t), "size{int64_t}",
+#if (PARSEC_SIZEOF_SIZE_T == 4)
+                                                sizeof(uint32_t), "size{uint32_t}",
+#else
```
therault, line 94: Suggested change:

```
#elif (PARSEC_SIZEOF__SIZE_T == 8)
     sizeof(uint64_t), "size{uint64_t}",
#else
#error "Unsupported case: sizeof(size_t) is neither 8 nor 4"
#end
```

therault (approved):
I'm nit-picking... If you don't want to implement the suggested change, just merge.

parsec/mca/device/device_gpu.c
devreal: Fair point, added the fallback.

PR #606, opened 2023-12-04
Relative path, symbolic links, and python examples

therault:
file(GLOB ...) executes in the source directory, and PYTHON_TOOLS_BIN_DIR is in the build directory.

As a consequence, when doing
  file(GLOB pyfiles RELATIVE ${PYTHON_TOOLS_BIN_DIR} examples/*)
pyfiles hold values like
  ../../../../../../parsec-src/tools/profiling/python/examples/demo_ptt.py

This works, but if the VPATH is pointed by a symbolic link, at make install time, ../../../../.. doesn't point to the directory hosting the parsec source.

I believe the intent was to have
  examples/demo_ptt.py
in pyfiles anyway. To do so, we need to glob it relative to the CMAKE_CURRENT_SOURCE_DIR (and do the install from the CMAKE_CURRENT_SOURCE_DIR).


tools/profiling/python/CMakeLists.txt
```diff
@@ -4,8 +4,7 @@ if( NOT BUILD_SHARED_LIBS )
   return()
 endif( NOT BUILD_SHARED_LIBS )
 
-FIND_PACKAGE(Python COMPONENTS Interpreter Development
```
bosilca, line 7: Not necessary.

bosilca (commented):
Left over from the 2to3 conversion. When using 2to3 all python files were converted to python3, the generated files, which were the one to be installed, were in the build directory. How did this survived for so long in the CI when we regularly test the profiling=on case?

therault:
It passed the CI, because it works as it is, if there is no symlink in the
VPATH: `file(glob ... RELATIVE ${OTHER_DIR} ... )` will find where
OTHER_DIR is relative to CWD, and write
`../../../../relative/path/to/OTHERDIR/...` in each file name.  But then,
if there is a symlink somewhere in that path, the install fails.

abouteiller:
On line 40 we set python_tools_bin_dir to cmake_build_directory, do we still need that intermediary variable?

abouteiller (approved):

PR #607, opened 2023-12-07
Hotfix: corner case in parsec_dtd_task_snprintf

therault:
When there are no parameters to a task (as for some sample_tasks() in dtd_test_task_generation), using `GET_HEAD_OF_PARAM_LIST(dtd_ask)` does not return NULL, but there are no parameters and we should not cycle over them.

This bug, signaled by @abouteiller  started to manifest after using the DSL specific task_snprintf for all task_snprintfs in PR #603.

I'm reading tc->nb_flows in a separate variable at the beginning of the code, so the compiler should be smart enough to store that in a register or realize that if nb_flows is not 0 the first time it does the loop, it will never become 0 and it can only test current_param != NULL


bosilca:
Replaced by  #609.

PR #608, opened 2023-12-08
Fix the parsec_future_t.

bosilca:
Our initial impementation used an old trick (functions prototypes without parameters), with the parameters correctly instantiated by the caller. Unfortunately, this is not supported by newer versions of the Cstandard, and compilers are becoming more and more picky. We have to move away from it.

Thus, we now use variadic functions for parsec futures. A quite invasive PR, but without any logical impact.


bosilca:
Needs #609 in order to be able to run all the tests. Once #609 is merged this can be pushed out of draft mode.

devreal (approved):
LGTM

therault (approved):

PR #612, opened 2024-01-11
Make sure the passed flag to check_type_size is not polluted

QingleiCao:


QingleiCao:
As discussed during the meeting, this is not the correct solution. It needs to be fixed in ExaGeoStat as `CMAKE_REQUIRED_FLAGS` is a global parameter.

PR #613, opened 2024-01-18
HOTFIX: make the default number of devices be all the devices seen by…

therault:
… cudaGetDeviceCount/hipGetDeviceCount/zeDeviceGet

Current code is testing if parsec_device_cuda_enabled_index < 0 (or similar counter for LZ), but parsec_device_cuda_enabled_index is the value returned by parsec_mca_param_reg_int_name() which is always positive, as it returns where that parameter is located in the parameter array.

The code wanted to check if testing if parsec_device_cuda_enabled < 0, so that if the user passes -1 or if the user omits to define the MCA parameter (default value is -1), then parsec takes the number of devices discovered by CUDA/HIP/LZ, as is claimed in the documentation of the MCA parameter.

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_component.c
+++ b/parsec/mca/device/cuda/device_cuda_component.c
@@ -239,11 +239,11 @@ static int device_cuda_component_open(void)
     }
 
     if( ndevices > parsec_device_cuda_enabled ) {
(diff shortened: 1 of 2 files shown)


devreal (approved):
LGTM

abouteiller (approved):

bosilca (approved):

PR #615, opened 2024-01-22
GPU: lazy memory allocation

therault:
PR #613 made all CI tests initialize the GPU if there is a GPU available. When running in oversubscribe mode, this can lead to falsely failing tests, that fail not because of a software issue, but because of a deployment issue (multiple processes trying to allocate 90% of the GPU memory at the same time).

In general, since we don't know if the GPU will be used or not, we should not preemptively allocate all the memory on it. This PR makes memory allocation lazy: it is delayed until we do try to use some GPU memory.

The drawback is that the first GPU task will also pay the cost of a large cuda_malloc / zmalloc etc...

### DIFF
--- a/parsec/mca/device/device_gpu.h
+++ b/parsec/mca/device/device_gpu.h
@@ -231,6 +231,11 @@ struct parsec_device_gpu_module_s {
                                                    *   is increased every time a new data is made available, so
                                                    *   that we know which tasks can be evaluated for submission.
                                                    */
+
+    int                         memory_percentage; /**< What % of the memory available on the device we want to use*/
+    int                         number_blocks;     /**< In case memory_percentage is not set, how many blocks we want to allocate on the device */
+    size_t                      eltsize;           /**< And what size in byte are these blocks */ 
+
     parsec_list_t              gpu_mem_lru;   /* Read-only blocks, and fresh blocks */
     parsec_list_t              gpu_mem_owned_lru;  /* Dirty blocks */
     parsec_fifo_t              pending;
@@ -334,10 +339,7 @@ typedef struct {
 
 #endif  /* defined(PROFILING) */
 
-int parsec_device_memory_reserve( parsec_device_gpu_module_t* gpu_device,
-                                      int           memory_percentage,
-                                      int           number_blocks,
-                                      size_t        eltsize );
+int parsec_device_memory_reserve( parsec_device_gpu_module_t* gpu_device );
 int parsec_device_attach( parsec_device_module_t* device, parsec_context_t* context );
 int parsec_device_detach( parsec_device_module_t* device, parsec_context_t* context );
 int parsec_device_taskpool_register(parsec_device_module_t* device, parsec_taskpool_t* tp);
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -186,8 +186,11 @@ int __parsec_execute( parsec_execution_stream_t* es,
                 task->status = PARSEC_TASK_STATUS_COMPLETE;
                 if(PARSEC_DEV_RECURSIVE >= tc->incarnations[chore_id].type) {
                     /* accelerators count their own executed tasks */
+#if 0
+                    /* This is wrong: chore_id is the index in incarnations, but it's not the device id */
                     parsec_device_module_t *dev = parsec_mca_device_get(chore_id);
                     parsec_atomic_fetch_inc_int64((int64_t*)&dev->executed_tasks);
+#endif
                 }
             }
             /* Record EXEC_END event only for incarnation that succeeds */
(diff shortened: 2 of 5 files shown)


parsec/mca/device/device_gpu.h
```diff
@@ -231,6 +231,11 @@ struct parsec_device_gpu_module_s {
                                                    *   is increased every time a new data is made available, so
                                                    *   that we know which tasks can be evaluated for submission.
                                                    */
+
+    int                         memory_percentage; /**< What % of the memory available on the device we want to use*/
+    int                         number_blocks;     /**< In case memory_percentage is not set, how many blocks we want to allocate on the device */
+    size_t                      eltsize;           /**< And what size in byte are these blocks */
```
devreal, line 237: Indentation seems off by 1

```suggestion
    int                        memory_percentage; /**< What % of the memory available on the device we want to use*/
    int                        number_blocks;     /**< In case memory_percentage is not set, how many blocks we want to allocate on the device */
    size_t                     eltsize;           /**< And what size in byte are these blocks */ 
```

parsec/scheduling.c
```diff
@@ -186,8 +186,11 @@ int __parsec_execute( parsec_execution_stream_t* es,
                 task->status = PARSEC_TASK_STATUS_COMPLETE;
                 if(PARSEC_DEV_RECURSIVE >= tc->incarnations[chore_id].type) {
                     /* accelerators count their own executed tasks */
+#if 0
+                    /* This is wrong: chore_id is the index in incarnations, but it's not the device id */
                     parsec_device_module_t *dev = parsec_mca_device_get(chore_id);
                     parsec_atomic_fetch_inc_int64((int64_t*)&dev->executed_tasks);
+#endif
```
devreal, line 193: Why is that disabled?

therault: Because in `__parsec_execute`  (scheduling.c:138-141) we do
```
   /* Find first bit in chore_mask that is not 0 */
    for(chore_id = 0; NULL != tc->incarnations[chore_id].hook; chore_id++)
        if( 0 != (task->chore_mask & (1<<chore_id)) )
            break;
```

The way I understand this, this finds the first TYPE of incarnation that we want to execute. If I have X CPUs, Y NVIDIA cards and Z Intel cards, incarnations can hold 3 entries, in any order, not X+Y+Z entries.
Then, once we have chosen the type, the evaluate can decide to skip the type, and then the hook can call `get_best_device()` to chose which device between the Y NVIDIA cards that are available.
BUT
Later in the file, at line 189, we do
```
parsec_device_module_t *dev = parsec_mca_device_get(chore_id);
parsec_atomic_fetch_inc_int64((int64_t*)&dev->executed_tasks);
```
I think this is erroneous, and I think  we don't have the device id at this time, it's lost within hook  (which calls `parsec_get_best_device().`

devreal: This counts only CPU and RECURSIVE, not GPUs (it's `>=`, not `<=`). The GPU accounting is done separately in `device_gpu.c`

therault: Right. This should be fixed in PR #616 . If you approve PR #616 and we merge it, I'll rebase and remove that part.

parsec/mca/device/device_gpu.h
abouteiller: see #616

parsec/scheduling.c
abouteiller: see #616 that fixes the underlying problem

bosilca (commented):
I can see how this solves the problem if we are not using the devices, but I don't see how it does if we are using them in an oversubscribed scenario. Second, it leads to non deterministic behaviors, as depending on what process starts allocating device memory first and when, the memory will be unevenly allocated among processes. 

A much cleaner solution would be to keep delaying the memory allocation until the parsec context is up. During the initialization we detect oversubscription, and if we expose that value (aka the number of processes per node), we can dynamically change the percentage of memory allowed for each process.

parsec/scheduling.c
abouteiller: I merged 616, this will need rebasing

abouteiller:
I have slated this for 4.0 but we are getting really close to our self-imposed deadline. Given that the CI fails randomly without this, I'd like to see it in.

devreal:
This needs to go through, blocks other PRs.

bosilca:
As described in my priuor comment, this PR does not provide a correct solution.

devreal:
We have two options: hold up everything because CI breaks or merge this to at least be able to run CI. Device memory allocation has been greedy for a long time and so far no one has cared about it.

bosilca:
The choice is between a deterministic bad behavior or a non-deterministic behavior. In two months we will forget about this non-determinism and we will start complaining that our accelerator tests fails randomly.

Moreover, we do have another solution: add the MCA `device_cuda_memory_use` to the tests (specialized by the accelerator).

abouteiller:
> Moreover, we do have another solution: add the MCA `device_cuda_memory_use` to the tests (specialized by the accelerator).

lets see if that works #629

abouteiller:
As shown in #629 there is something more nefarious going on, we try to allocate way more memory that is possibly available whatever limits we set.

bosilca:
This is very stale, if there is a need for this capability it should be reimplemented not rebased.

PR #616, opened 2024-01-22
bugfix: in dtd sometimes the cpu incarnation and gpu incarnations are

abouteiller:
not in order and that caused all sort of havoc. Also we should not try to obtain an invalid iid when querying per-stream-infos (again from dtd when gpus are disabled by hand)

### DIFF
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -168,16 +168,15 @@ int __parsec_execute( parsec_execution_stream_t* es,
             }
         }
 
-        parsec_hook_t *hook = tc->incarnations[chore_id].hook;
-        if( NULL == hook ) goto next_chore;
 #if defined(PARSEC_DEBUG)
         parsec_debug_verbose(5, parsec_debug_output, "Thread %d of VP %d Execute %s[%d] chore %d",
                              es->th_id, es->virtual_process->vp_id,
                              tmp, tc->incarnations[chore_id].type,
                              chore_id);
 #endif
+        parsec_hook_t *hook = tc->incarnations[chore_id].hook;
+        assert( NULL != hook );
         rc = hook( es, task );
-
 #if defined(PARSEC_PROF_TRACE)
         task->prof_info.task_return_code = rc;
 #endif
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -168,15 +168,16 @@ int __parsec_execute( parsec_execution_stream_t* es,
             }
         }
 
+        parsec_hook_t *hook = tc->incarnations[chore_id].hook;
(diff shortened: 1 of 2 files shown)


parsec/scheduling.c
```diff
@@ -185,8 +186,9 @@ int __parsec_execute( parsec_execution_stream_t* es,
                 /* Let's assume everything goes just fine */
                 task->status = PARSEC_TASK_STATUS_COMPLETE;
                 if(PARSEC_DEV_RECURSIVE >= tc->incarnations[chore_id].type) {
-                    /* accelerators count their own executed tasks */
-                    parsec_device_module_t *dev = parsec_mca_device_get(chore_id);
+                    /* accelerators count their own executed tasks so this is a
+                     * type DEV_RECURSIVE or DEV_CPU, they must be dev 0 and 1 */
+                    parsec_device_module_t *dev = parsec_mca_device_get(tc->incarnations[chore_id].type - PARSEC_DEV_CPU);
```
therault, line 191: That's not correct either... dev is a possible device for this chore, but not necessarily the device that we execute the task on.

If hook is a chore of type device, it will call `parsec_get_best_device`, and that's what will chose which device will actually execute the task...

@bosilca at some point you were working on a PR to move the get_best_device call before the hook. Do you still have this PR ? Can this solve this issue?

abouteiller: this executes only if this is a CPU or RECURSIVE device, and we have only one of each.

therault (approved):

parsec/scheduling.c
```diff
@@ -168,15 +168,16 @@ int __parsec_execute( parsec_execution_stream_t* es,
             }
         }
 
+        parsec_hook_t *hook = tc->incarnations[chore_id].hook;
```
bosilca, line 171: the incarnations are supposed to be packed. How can we get a NULL hook here ?

abouteiller: the incarnation exists because it has been created from a DTD task class, but the hook is not initialized because there is no device corresponding to the incarnation

parsec/scheduling.c
```diff
@@ -526,7 +528,10 @@ int __parsec_task_progress( parsec_execution_stream_t* es,
         case PARSEC_HOOK_RETURN_NEXT:    /* Try next variant [if any] */
```
abouteiller, line 528: why is next an error here?

parsec/scheduling.c
abouteiller: I cannot reproduce the error case I saw before, and I don't understand why it would have happened in the first place, so I just put an assert instead if it ever shows up again.

PR #621, opened 2024-01-26
Add a context-level query capability.

bosilca:
Augment the PaRSEC context API with query capabilities, allowing users to find the number of processes, cores and other information from a PaRSEC context.

### DIFF
--- a/parsec/parsec.c
+++ b/parsec/parsec.c
@@ -2759,3 +2759,50 @@ void parsec_set_my_execution_stream(parsec_execution_stream_t *es)
 {
     PARSEC_TLS_SET_SPECIFIC(parsec_tls_execution_stream, es);
 }
+
+
+/**
+ * Query the number of devices of the requested type available in the
+ * PaRSEC context.
+ */
+int parsec_context_query(parsec_context_t *context, parsec_context_query_cmd_t cmd, ...)
+{
+    parsec_device_module_t* dev;
+    va_list args;
+    va_start(args, cmd);
+
+    switch(cmd) {
+        case PARSEC_CONTEXT_QUERY_NODES:
+            switch (parsec_communication_engine_up) {
+                case 0: return 0;  /* context not ready for distributed runs, and lacking datatype chandling capabilities */
+                case 1: return 1;  /* single node runs, but the context has datatype management capabilties */
+                case 2: return PARSEC_ERR_NOT_FOUND; /* we are in a distributed run, but the MPI engine is not yet ready, so the nb_nodes might not be accurate */
+                case 3: return context->nb_nodes;
+            }
+            return PARSEC_ERROR;
+
+        case PARSEC_CONTEXT_QUERY_RANK:
+            return context->my_rank;
+
+        case PARSEC_CONTEXT_QUERY_DEVICES:
+            int device_type = va_arg(args, int), count = 0;
+            for( uint32_t i = 0; i < parsec_nb_devices; i++ ) {
+                dev = parsec_mca_device_get(i);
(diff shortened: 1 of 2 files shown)


parsec/parsec.c
```diff
@@ -2759,3 +2759,50 @@ void parsec_set_my_execution_stream(parsec_execution_stream_t *es)
...
+
+/**
+ * Query the number of devices of the requested type available in the
+ * PaRSEC context.
+ */
+int parsec_context_query(parsec_context_t *context, parsec_context_query_cmd_t cmd, ...)
+{
+    parsec_device_module_t* dev;
+    va_list args;
+    va_start(args, cmd);
```
devreal, line 2772: Missing call to `va_end`

parsec/parsec.c
```diff
@@ -2759,3 +2759,50 @@ void parsec_set_my_execution_stream(parsec_execution_stream_t *es)
 {
     PARSEC_TLS_SET_SPECIFIC(parsec_tls_execution_stream, es);
 }
+
+
+/**
+ * Query the number of devices of the requested type available in the
+ * PaRSEC context.
+ */
```
devreal, line 2767: It's not just deviced :)

```suggestion
/**
 * Query the numerical properties of the given
 * PaRSEC context.
 */
```

parsec/runtime.h
```diff
@@ -286,6 +286,33 @@ int parsec_context_add_taskpool( parsec_context_t* context, parsec_taskpool_t* t
...
+ * Query PaRSEC context capabilities.
+ */
+
+typedef enum parsec_context_query_cmd_e {
+    PARSEC_CONTEXT_QUERY_NODES,
+    PARSEC_CONTEXT_QUERY_RANK,
+    PARSEC_CONTEXT_QUERY_DEVICES,
+    PARSEC_CONTEXT_QUERY_CORES,
+    PARSEC_CONTEXT_QUERY_ACTIVE_TASKPOOLS
+} parsec_context_query_cmd_t;
```
devreal, line 299: Some documentation for vararg parameters (`PARSEC_CONTEXT_QUERY_DEVICES`) would be good.

devreal (changes requested):
I like this, nice and concise. Just a few comments.

parsec/parsec.c
bosilca: There are too many return points to place an va_end. Plus everything related to the va_list is on the stack, so returning should clean up everything.

bosilca:
Added documentation for the vararg, and the different commands.

devreal (approved):

parsec/parsec.c
```diff
@@ -2759,3 +2759,51 @@ void parsec_set_my_execution_stream(parsec_execution_stream_t *es)
...
+    parsec_device_module_t* dev;
+    va_list args;
+    va_start(args, cmd);
+
+    switch(cmd) {
+        case PARSEC_CONTEXT_QUERY_NODES:
+            switch (parsec_communication_engine_up) {
+                case 0: return 0;  /* context not ready for distributed runs, and lacking datatype chandling capabilities */
+                case 1: return 1;  /* single node runs, but the context has datatype management capabilties */
+                case 2: return -1; /* we are in a distributed run, but the MPI engine is not yet ready, so the nb_nodes might not be accurate */
```
abouteiller, line 2778: PARSEC_ERROR and -1 are the same numerical value. 

Maybe we could return PARSEC_ERR_NOT_FOUND in that case?

abouteiller (changes requested):

abouteiller (approved):

PR #622, opened 2024-01-26
Hotfix: profiling initialized from the wrong thread in comm engine

therault:
After we introduced lazy initialization of the comm engine, we left the profiling initialization for the remote_dep comm engine be called from the main parsec thread. That initialization must be called from the communication thread, this patch moves it to the actual comm thread. (bug found and fixed by @devreal).

### DIFF
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -190,6 +190,71 @@ static int remote_dep_nothread_memcpy(parsec_execution_stream_t* es,
 
 int remote_dep_ce_reconfigure(parsec_context_t* context);
 
+#ifdef PARSEC_PROF_TRACE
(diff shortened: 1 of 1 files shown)


bosilca (commented):
What is the real change here ? This patch moves a bunch of code around for no obvious reason.

devreal:
Thanks @therault!

@bosilca The real change is that `remote_dep_mpi_profiling_init` should be called from within `remote_dep_dequeue_main` (the comm thread's main function) instead of `remote_dep_ce_init` (called by the main thread to spin up the communication backend). The rest is moving code around for correct ordering.

bosilca:
Can we just pre-decleare the function instead of moving all the code around ?

abouteiller:
We will use #626 instead

PR #626, opened 2024-01-31
Move comm profiling initialization into comm thread

devreal:
The comm thread must call `remote_dep_mpi_profiling_init` to make sure it's thread-local variable is properly set.

Supersedes https://github.com/ICLDisco/parsec/pull/622

### DIFF
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -190,6 +190,12 @@ static int remote_dep_nothread_memcpy(parsec_execution_stream_t* es,
 
 int remote_dep_ce_reconfigure(parsec_context_t* context);
 
+#ifdef PARSEC_PROF_TRACE
+static void remote_dep_mpi_profiling_init(void);
+#else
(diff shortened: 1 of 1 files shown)


therault (approved):

abouteiller (approved):

abouteiller:
the CI errors can be ignored @therault are you satisfied with the changes? If so please merge

bosilca (commented):
`remote_dep_mpi_profiling_init` function does two things: registers the dictionary entries for the communication thread and allocate the first bucket for the profiling stream. As far as I recall the dictionary must be created early in the process, this is not something that one can be part of the lazy initialization. 

The correct approach would be to split the current `remote_dep_mpi_profiling_init` function in two, one for registration (that must be called where `remote_dep_mpi_profiling_init` is called today), and one doing memory allocation (basically the last 2 lines of`remote_dep_mpi_profiling_init`), that should be called from the communication thread.

devreal:
@bosilca please recheck

bosilca (approved):

PR #629, opened 2024-01-31
Prevent CI from running OOM when oversubscribing GPUs

abouteiller:


bosilca (approved):

abouteiller:
`Allocating 79855878144 bytes of memory on the GPU device failed` 
that's 79GB, wat?

abouteiller:
There is really not a reason not to do this in the CI, without decreasing the value of not initializing cuda memory if we won't use it as an independent effort.

bosilca:
`not not of the not not`. What was that supposed to mean ?

PR #631, opened 2024-02-01
Cleanup/cosmetics

abouteiller:
A PR for all things cosmetic like wrong output alignment, units, etc.

### DIFF
--- a/parsec/mca/device/device.c
+++ b/parsec/mca/device/device.c
@@ -26,6 +26,7 @@
 #if defined(PARSEC_HAVE_STRING_H)
 #include <string.h>
 #endif  /* defined(PARSEC_HAVE_STRING_H) */
+#include <math.h>
 #if defined(__WINDOWS__)
 #include <windows.h>
 #endif  /* defined(__WINDOWS__) */
@@ -375,7 +376,6 @@ void parsec_devices_print_statistics(parsec_context_t *parsec_context, uint64_t 
     float best_required_in, best_required_out;
     char *data_in_unit, *data_out_unit, *d2d_unit;
     char *required_in_unit, *required_out_unit;
-    char percent1[64], percent2[64], percent3[64];
     parsec_device_module_t *device;
     uint32_t i;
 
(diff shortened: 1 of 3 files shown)


parsec/mca/device/device.c
```diff
@@ -375,7 +375,6 @@ void parsec_devices_print_statistics(parsec_context_t *parsec_context, uint64_t
     float best_required_in, best_required_out;
     char *data_in_unit, *data_out_unit, *d2d_unit;
     char *required_in_unit, *required_out_unit;
-    char percent1[64], percent2[64], percent3[64];
```
bosilca, line 378: What's wrong with printing `nan` when there are nothing to be printed ?

parsec/mca/device/device.c
```diff
@@ -443,23 +440,11 @@ void parsec_devices_print_statistics(parsec_context_t *parsec_context, uint64_t
     parsec_compute_best_unit( total_data_out,     &best_data_out,     &data_out_unit     );
     parsec_compute_best_unit( total_d2d,          &best_d2d,          &d2d_unit          );
 
-    if( 0 == total_required_in ) {
```
bosilca, line 446: What if `*_required_*` is 0 (aka the device was not used) ?

parsec/mca/device/device.c
abouteiller: That also answer the next question, dividing by 0 will naturally produce a nan that will be printed without all that complication.

parsec/mca/device/device.c
abouteiller: see above, produces a nan value that we will print.

parsec/mca/device/device.c
bosilca: or raise a signal that we don't deal with. I would rather handle the `nan` internally than rely on an external mechanism that we do not control.

abouteiller: You are correct the changed code is valid only if the implementation is IEEE754. `If the second operand is zero, the behavior is undefined, except that if the IEEE floating-point arithmetic is supported, and the floating-point division is taking place, then`. 

That same defect is present in the code per-device computation right above,  I will update it to match.

bosilca (approved):

Issue #630, opened 2024-02-01
GPU memory allocation far exceeds requested size


abouteiller:
## Describe the bug

Rarely, the GPU memory allocator will try to obtain a very large amount of memory, far in excess of what is physically available or requested by the mca params. 

### To Reproduce

This is not particular to allreduce:mp, it just happened on this on this time 
```
104/111 Test #104: dsl/dtd/allreduce:mp ............................................***Failed   13.36 sec
W@00000 /tmp/parsec/parsec/parsec/mca/device/cuda/device_cuda_module.c:428 cudaStreamCreate out of memory
[line repeats multiple times]

W@00001 /tmp/parsec/parsec/parsec/mca/device/cuda/device_cuda_module.c:348 cudaMallocout of memory
W@00001 GPU[cuda(6)] Allocating 79855878144 bytes of memory on the GPU device failed


W@00003 /tmp/parsec/parsec/parsec/mca/device/cuda/device_cuda_module.c:428 cudaStreamCreate out of memory
W@00000 /tmp/parsec/parsec/parsec/mca/device/cuda/device_cuda_module.c:428 cudaStreamCreate out of memory
Root: 0; value=780

My rank: 1, bcast recv data: 780
My rank: 2, bcast recv data: 780
My rank: 3, bcast recv data: 780
[c53340475b50:333063] *** Process received signal ***
[c53340475b50:333063] Signal: Segmentation fault (11)
[c53340475b50:333063] Signal code: Address not mapped (1)
[c53340475b50:333063] Failing at address: 0x10
[c53340475b50:333063] [ 0] /lib64/libc.so.6(+0x54df0)[0x7f929adf1df0]
[c53340475b50:333063] [ 1] /tmp/parsec/parsec/build/Release/shared_ON/profile_OFF/parsec/libparsec.so.4(+0x29d40)[0x7f929b69fd40]
[c53340475b50:333063] [ 2] /tmp/parsec/parsec/build/Release/shared_ON/profile_OFF/parsec/libparsec.so.4(parsec_cuda_module_fini+0x149)[0x7f929b6d6aa9]
[c53340475b50:333063] [ 3] /tmp/parsec/parsec/build/Release/shared_ON/profile_OFF/parsec/libparsec.so.4(+0x5f297)[0x7f929b6d5297]
[...]

PR #633, opened 2024-02-07
Alternative solution to the CI problem with GPUs

therault:
Testing shows that the CI has issues forwarding the environment variables to ctest.

Moreover, if one tests directly by calling 'ctest' on our own machine, the CI-level environment variables does not fix the problem.

In this PR, we set the `PARSEC_MCA_cuda_device_enabled` environment variable to 0 directly from CMake in `parsec_addtest_cmd`. 

   - This does the same thing as the change in the CI files was trying to do
   - specific tests can still overwrite this by either
         - setting the environment variable to 1 after adding the test with `parsec_addtest_cmd`
         - passing the MCA parameter to the command line
         - setting the number of GPUs on the command line (e.g. in dplasma)
  
This PR should replace PR#615 and close Issue #630

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -94,12 +94,6 @@ jobs:
       # run: ctest -C $BUILD_TYPE
       run: |
         source ${{github.workspace}}/.github/CI/spack_setup.sh
-        # enable devices only in tests that explicitely require them
-        PARSEC_MCA_device_cuda_enabled=0
-        PARSEC_MCA_device_hip_enabled=0
-        # restrict memory use for oversubscribed runners
-        PARSEC_MCA_device_cuda_memory_use=10
-        PARSEC_MCA_device_hip_memory_use=10
         ctest --output-on-failure
 
     - name: Save Artifact
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -66,6 +66,10 @@ function(parsec_addtest_cmd target)
   endif()
   add_test(${ARGV})
   set_tests_properties(${target} PROPERTIES DEPENDS parsec_build_tests_test)
+  # Disable GPU accelerators by default for all tests
+  set_property(TEST ${target} PROPERTY ENVIRONMENT "PARSEC_MCA_device_cuda_enabled=0")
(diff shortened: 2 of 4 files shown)


therault:
It seems ctest, as it is run from the CI, doesn't forward the environment variables passed to it by the shell in CI.

For example, we wrote
```
PARSEC_MCA_device_cuda_enabled=0 ctest ...
```
in `.github/workflows/build_cmake.yml`, but when we test if `PARSEC_MCA_device_cuda_enabled` is defined and to what in a test, we don't find it in the environment. See 
https://github.com/ICLDisco/parsec/actions/runs/7821760842/job/21339489092?pr=633
for an example (the test name was `cuda-disabled-by-default-in-environment`).

I tested the following:
  - if PARSEC_MCA_device_cuda_enabled is found in the environment of processes launched by ctest in CI. Result: no.
  - if we define some variable in the mpiexec command line, does the child process sees that variable. Result: yes.

The reliable way seems to use `set_property(TEST test-name PROPERTY ENVIRONMENT "PARSEC_MCA_device_cuda_enabled=0")` directly in CMake files, when we define the tests.

I will code this in `parsec_addtest`

bosilca:
If you code this in `add_test` we will never test the accelerators. This should be part of the CI  ...

therault:
My understanding of MCA is that environment variables are the weakest
selector. When we set the MCA parameter via command line, or when we set
the MCA parameter via the API, this will supersede that environment
variable.

Thus, as written in the comment of the PR, this should not prevent to run
any GPU test, and contrary to the approach where we set environment
variables only in the CI case, that reduces the chance of CI behaving
differently from interactive ctest runs.

.github/workflows/build_cmake.yml
```diff
@@ -94,12 +94,6 @@ jobs:
       # run: ctest -C $BUILD_TYPE
       run: |
         source ${{github.workspace}}/.github/CI/spack_setup.sh
-        # enable devices only in tests that explicitely require them
-        PARSEC_MCA_device_cuda_enabled=0
-        PARSEC_MCA_device_hip_enabled=0
-        # restrict memory use for oversubscribed runners
-        PARSEC_MCA_device_cuda_memory_use=10
```
abouteiller, line 101: this is also needed to be replicated

abouteiller (changes requested):

abouteiller (approved):

Issue #641, opened 2024-03-07
stress:gpu crashes when compiled with CUDA but run without a device


abouteiller:
## Describe the bug
The stress:gpu compiled with CUDA support may crash when run on a system without a CUDA GPU. See

https://github.com/ICLDisco/parsec/actions/runs/8190649752/job/22398165838?pr=515

Note that there appears to be some interaction with Intel ZE libraries that is unexpected here. 

### To Reproduce

See https://github.com/ICLDisco/parsec/actions/runs/8190649752/job/22398165838?pr=515

abouteiller:
There are 2 different failure modicum: 

1. stress tries to allocate 33GB of memory, which may or may not be possible, especially on low-end cuda devices, or as host memory. 
2. stage tries to run hook = tc->incarnations[chore_id], but disabling the cuda device nullified the hook for the only valid chore_id of 0 (the incarnations array does not have a CPU implementation).

therault:
I pushed some commit in PR #642 to handle the lack of device more gracefully, both in the test and the runtime system.

However, the test still fails in the PI, since there is no way to fix that at the runtime/test level. The issue is with the CI logic here: serotonin is a machine that has both NVIDIA and ROCM installed at the software stack level, but only ROCM device. The cmake logic defaults to NVIDIA in that case, and we should pass ROCM.

I don't think we should complicate the CMake logic of the tests more here: they fail because we asked them to run on  NVIDIA and there is no NVIDIA cards. 

Geri is working on preparing a PR on the CI to fix this issue at the CI logic level.

bosilca:
the CI instances are tagged with their devices. As an example serotonin is tagger with `gpu_amd` while guyot is tagged with `gpu_nvidia`. The CI should use these tags to drive the correct set of testing.

abouteiller:
There is a third problem: 

The JDF of the stress tester is not symmetrical

https://github.com/ICLDisco/parsec/blob/1ababbe248064c5f3deaab2f9b04e56b556a3f02/tests/runtime/cuda/stress.jdf#L106
https://github.com/ICLDisco/parsec/blob/1ababbe248064c5f3deaab2f9b04e56b556a3f02/tests/runtime/cuda/stress.jdf#L128

we believe that the READ B should be from READ_A(m, r) without the (m+r)%mt randomization

PR #642, opened 2024-03-08
Fix the lack of direct GPU to GPU communications in multi-device runs.

therault:
In https://github.com/ICLDisco/parsec/pull/570, we moved from using cuda_index to using device_index in the 'nvlink' mask that decides if we can directly communicate to another GPU. However, this bitmask was initialized at query time, before devices get assigned a device_index. As a consequence, the bitmask was wrong and no direct device to device communication was happening.

In this PR, we add a step, after all devices have been registered, to complete this initialization.

An alternative is to come back to using cuda_index-based bitmasks, but then the decision if two GPUs can directly communicate is device-type specific and needs to be moved from device_gpu.c to device_cuda_module.c, which means we add another function call to device_cuda_module.c inside the stage_in() function of device_gpu.c

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_component.c
+++ b/parsec/mca/device/cuda/device_cuda_component.c
@@ -35,12 +35,12 @@ static int device_cuda_component_query(mca_base_module_2_0_0_t **module, int *pr
 static int device_cuda_component_register(void);
 
 /* mca params */
-int parsec_device_cuda_enabled_index, parsec_device_cuda_enabled;
+int parsec_device_cuda_enabled_index, parsec_device_cuda_enabled, parsec_cuda_nvlink_mask;
 int parsec_cuda_max_streams = PARSEC_GPU_MAX_STREAMS;
 int parsec_cuda_memory_block_size, parsec_cuda_memory_percentage, parsec_cuda_memory_number_of_blocks;
 char* parsec_cuda_lib_path = NULL;
 
-static int cuda_mask, cuda_nvlink_mask;
+static int cuda_mask;
 static int parsec_cuda_sort_pending;
 
 #if defined(PARSEC_PROF_TRACE)
@@ -122,37 +122,6 @@ static int device_cuda_component_query(mca_base_module_t **module, int *priority
         parsec_device_cuda_component.modules[j] = NULL;
     }
 
-    parsec_device_cuda_module_t *source_gpu, *target_gpu;
-    cudaError_t cudastatus;
-
-    for( i = 0; i < parsec_device_cuda_enabled && NULL != (source_gpu = (parsec_device_cuda_module_t*)parsec_device_cuda_component.modules[i]); i++ ) {
-        int canAccessPeer;
-        source_gpu->super.peer_access_mask = 0;
-
-        if( ! ( (1<<i) & cuda_nvlink_mask ) )
-            continue; /* The user disabled NVLINK for that GPU */
-
-        cudastatus = cudaSetDevice( source_gpu->cuda_index );
-        PARSEC_CUDA_CHECK_ERROR( "(parsec_device_cuda_component_query) cudaSetDevice", cudastatus,
-                                 {continue;} );
-
-        for( j = 0; NULL != (target_gpu = (parsec_device_cuda_module_t*)parsec_device_cuda_component.modules[j]); j++ ) {
-            if( i == j ) continue;
-
-            /* Communication mask */
-            cudastatus = cudaDeviceCanAccessPeer( &canAccessPeer, source_gpu->cuda_index, target_gpu->cuda_index );
-            PARSEC_CUDA_CHECK_ERROR( "(parsec_device_cuda_component_query) cudaDeviceCanAccessPeer", cudastatus,
-                                     {continue;} );
-            if( 1 == canAccessPeer ) {
-                cudastatus = cudaDeviceEnablePeerAccess( target_gpu->cuda_index, 0 );
-                PARSEC_CUDA_CHECK_ERROR( "(parsec_device_cuda_component_query) cuCtxEnablePeerAccess", cudastatus,
-                                         {continue;} );
-                source_gpu->super.peer_access_mask = (int16_t)(source_gpu->super.peer_access_mask | (int16_t)(1 <<
-                        target_gpu->super.super.device_index));
-            }
-        }
-    }
-
     parsec_device_enable_debug();
 
     /* module type should be: const mca_base_module_t ** */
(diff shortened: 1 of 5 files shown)


bosilca:
Why did we moved from using cuda_index to using the device_index ? We cannot use d2d on devices of different types, and we check for that in `parsec_default_gpu_stage_in` and `parsec_default_gpu_stage_out`, so using the device_index in the mask makes little sense. Going back and rebuilding this mask based on cuda_index seems like a simpler and most resilient solution (we could change the device_index without having to rebuild all the info).

therault:
device_gpu.c doesn't know the cuda_index (or hip_index or
level_zero_index), that's the reason why we moved that mask to global
device_index based.

I don't think we want to replicate the full parsec_device_gpu_stage_in()
function in all device implementations, so if we want to go back to
cuda_index based bitmask, we can replace the test

`if( gpu_device->peer_access_mask & (1 <<
candidate_dev->super.device_index) )`
(and similar tests in device_gpu.c)
by a call to the device-specific implementation, something like
`if( gpu_device->peer_access(candidate_dev)`

I'm fine with that, but I thought the global device_index mask would save
these function calls that cannot be inlined in stage_in, which is taking a
lot of time already.

abouteiller:
Passes all ctests and works with large size POTRF -g 8 with nvlink active on Leconte.

abouteiller (approved):

parsec/mca/device/device_gpu.c
```diff
@@ -1510,11 +1524,14 @@ parsec_device_data_stage_in( parsec_device_gpu_module_t* gpu_device,
 
     /* update the data version in GPU immediately, and mark the data under transfer */
     assert((gpu_elem->version != candidate->version) || (gpu_elem->data_transfer_status == PARSEC_DATA_STATUS_NOT_TRANSFER));
-    gpu_elem->version = candidate->version;
+    if( PARSEC_FLOW_ACCESS_WRITE & type && gpu_task->task_type != PARSEC_GPU_TASK_TYPE_PREFETCH )
+        gpu_elem->version = candidate->version + 1;
+    else
+        gpu_elem->version = candidate->version;
```
devreal, line 1530: Can you please document why we set the version the way we set it? This is delicate and not obvious.

parsec/mca/device/device_gpu.c
```diff
@@ -1510,11 +1524,14 @@ parsec_device_data_stage_in( parsec_device_gpu_module_t* gpu_device,
 
     /* update the data version in GPU immediately, and mark the data under transfer */
     assert((gpu_elem->version != candidate->version) || (gpu_elem->data_transfer_status == PARSEC_DATA_STATUS_NOT_TRANSFER));
-    gpu_elem->version = candidate->version;
+    if( PARSEC_FLOW_ACCESS_WRITE & type && gpu_task->task_type != PARSEC_GPU_TASK_TYPE_PREFETCH )
```
devreal, line 1527: For readability

```suggestion
    if( (PARSEC_FLOW_ACCESS_WRITE & type) && (gpu_task->task_type != PARSEC_GPU_TASK_TYPE_PREFETCH) )
```

parsec/mca/device/device_gpu.c
```diff
@@ -1700,7 +1717,7 @@ parsec_device_callback_complete_push(parsec_device_gpu_module_t   *gpu_device,
                     /* Nobody is at the door to handle that event on the source of that data...
                      * we do the command directly */
                     parsec_atomic_lock( &source->original->lock );
-                    int readers = parsec_atomic_fetch_sub_int32(&source->readers, 1);
+                    int readers = parsec_atomic_fetch_sub_int32(&source->readers, 1) - 1;
```
devreal, line 1720: Curious: does that mean we were leaking items here because we never pushed them back to the LRU?

therault: In the (somewhat rare) case when there was no GPU-manager for the source GPU, yes, we would leak some memory allocated via the `zone_malloc()`. The checks at termination would then complain, which is how we found the leak, and why we added the `zone_debug()` call at the completion checks in debug mode.

bosilca:
I can understand how the need to add debugging messages can be justified as part of this PR, but there are other things (such as changing the coherency state or moving data copy version manipulation code across function calls) that do not fit into the description of this PR.

Also, there are several instances where the indentation of the new code is incorrect, 8 commits (half of them not signed)  for a seemingly minor issue.

parsec/mca/device/device_gpu.c
```diff
@@ -1339,6 +1331,10 @@ parsec_device_data_stage_in( parsec_device_gpu_module_t* gpu_device,
 
         parsec_data_end_transfer_ownership_to_copy(original, gpu_device->super.device_index, (uint8_t)type);
 
+        if( (PARSEC_FLOW_ACCESS_WRITE & type) && (gpu_task->task_type != PARSEC_GPU_TASK_TYPE_PREFETCH) ) {
+            gpu_elem->version = candidate->version + 1;
+		}
```
devreal, line 1336: indentation

parsec/scheduling.c
```diff
@@ -175,7 +175,9 @@ int __parsec_execute( parsec_execution_stream_t* es,
                              chore_id);
 #endif
         parsec_hook_t *hook = tc->incarnations[chore_id].hook;
-        assert( NULL != hook );
+        if( NULL == hook ) {
+			goto next_chore; /* Gracefuly manage the case when NO chores are available for this task */
+		}
```
devreal, line 180: indentation

tests/runtime/cuda/stage_main.c
```diff
@@ -23,6 +23,7 @@ int main(int argc, char *argv[])
     int NB;
     int P = 1;
     int ret = 0;
+	int nb_gpus = 0;
```
devreal, line 26: indentation

tests/runtime/cuda/stage_main.c
```diff
@@ -41,6 +42,19 @@ int main(int argc, char *argv[])
...
+			nb_gpus++;
+		}
+	}
+	if(nb_gpus == 0) {
+		fprintf(stderr, "This test can only run if at least one GPU device is present\n");
+		parsec_fini(&parsec);
+		MPI_Finalize();
+		return EXIT_FAILURE;
+	}
+
```
devreal, line 57: indentation

abouteiller:
> I can understand how the need to add debugging messages can be justified as part of this PR, but there are other things (such as changing the coherency state or moving data copy version manipulation code across function calls) that do not fit into the description of this PR.
> 
> Also, there are several instances where the indentation of the new code is incorrect, 8 commits (half of them not signed) for a seemingly minor issue.

The issue is not minor at all. Turning back on D2D management unearthed a whole bunch of bugs that would produce wrong results (in particular in TRSM where we would exercise CPU->GPU1,2,3->CPU data motions). While I agree that the description of the PR does not match the real scope of what is achieved here, the changes are not random and serve the greater case of making D2D work at a basic level.

parsec/scheduling.c
abouteiller: I had made that same change on a prior pr and got shot down for it. Let me review this why it wasn't acceptable last time around.

abouteiller:
I will be extracting the fix for #641 in a separate PR, @therault that means I will be rewriting history in your branch to achieve this sorry.

abouteiller (approved):

parsec/scheduling.c
abouteiller: Found back the discussion https://github.com/ICLDisco/parsec/pull/616#discussion_r1462462587, I could not reproduce a scenario where that would happen, and added the assert so we would find that error easily when it shows up again, which is now.

PR #644, opened 2024-03-12
Consolidated error handling when GPU only tests execute on CPU systems

abouteiller:
Fixes issue #641 

* [ ] capture ci tag and compile for Cuda or HIP based on this tag, not on availability of the software stack. 
* [x] decide if we keep the check in stage_main about counting GPUs. We now will detect (later) erroneous behavior and produce a `parsec_fatal` so maybe that is already clear enough?
* [x] cleanup error cases for when hook==NULL, cases have been split between when the dyld function was not populated (goto next_chore) and other cases (this is a bug, assert/fail). 
* [x] stress tester is not symmetrical https://github.com/ICLDisco/parsec/issues/641#issuecomment-2040819481

### DIFF
--- a/tests/runtime/cuda/stage_main.c
+++ b/tests/runtime/cuda/stage_main.c
@@ -23,6 +23,7 @@ int main(int argc, char *argv[])
     int NB;
     int P = 1;
     int ret = 0;
+    int nb_gpus = 0;
 
 #if defined(DISTRIBUTED)
     {
@@ -39,6 +40,19 @@ int main(int argc, char *argv[])
         /* Failed to correctly initialize. In a correct scenario report*/
          /* upstream, but in this particular case bail out.*/
         exit(-1);
+    }
+
+    for(int id = 0; id < (int)parsec_nb_devices; id++) {
+        parsec_device_module_t *device = parsec_mca_device_get(id);
+        if( PARSEC_DEV_IS_GPU(device->type) ) {
+            nb_gpus++;
+        }
+    }
+    if(nb_gpus == 0) {
+        fprintf(stderr, "This test can only run if at least one GPU device is present\n");
+        parsec_fini(&parsec);
+        MPI_Finalize();
+        return EXIT_FAILURE;
     }
 
     assert(size == 1);
--- a/tests/runtime/cuda/stage_main.c
+++ b/tests/runtime/cuda/stage_main.c
@@ -14,7 +14,6 @@ int main(int argc, char *argv[])
 {
     parsec_context_t *parsec = NULL;
     parsec_taskpool_t *tp;
-    int i;
     int size = 1;
     int rank = 0;
     int M;
@@ -22,8 +21,7 @@ int main(int argc, char *argv[])
     int MB;
     int NB;
     int P = 1;
-    int ret = 0;
-    int nb_gpus = 0;
+    int ret = EXIT_SUCCESS;
 
 #if defined(DISTRIBUTED)
     {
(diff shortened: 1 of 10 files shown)


parsec/scheduling.c
```diff
@@ -175,7 +175,9 @@ int __parsec_execute( parsec_execution_stream_t* es,
                              chore_id);
 #endif
         parsec_hook_t *hook = tc->incarnations[chore_id].hook;
-        assert( NULL != hook );
+        if( NULL == hook ) {
+            goto next_chore; /* Gracefuly manage the case when NO chores are available for this task */
```
bosilca, line 179: I think this is coming up regularly, but as far as I recall the code it is not correct. The chores are packed, no NULL chores can exists except at the end. Second, this loop is only executed for valid chores (aka not NULL hook). So, the only way to get into this is if no eval function accepted the work, in which case the correct outcome is to bail out of the loop (not go to the next chore).

abouteiller: Yes I remembered this discussion and checked the history https://github.com/ICLDisco/parsec/pull/642#discussion_r1522166794 Summary of last time: I dropped it because I couldn't reproduce it anymore in the dtd case I was looking at.

Now it is back. What happens here is that we have only 1 valid chore (the JDF declares only a GPU chore), but we have runtime disabled it (set hook=NULL by force) because we didn't find a device of that type during the task pool creation. Not sure there is anything related to evaluate, looking into more details.

bosilca: ok, that makes sense. But, at the end, it is an incorrect application, as we do not have support for all tasks available. If we stop this at a single process, it will be difficult to properly report and act on this. Thus, I think this might be something that needs to be checked upfront, before the first task is even executed, and all processes must act uniformly.

abouteiller: The chore is NULL because it is a dyld function, and when we tried to populate the dyld functions,  the device was not active, so we didn't search for it at all. The proposed behavior in that case seems reasonable to me. Discuss.

tests/runtime/cuda/stage_main.c
```diff
@@ -41,6 +42,19 @@ int main(int argc, char *argv[])
         exit(-1);
     }
 
+    for(int id = 0; id < (int)parsec_nb_devices; id++) {
```
abouteiller, line 45: Don't think this is necessary if we have appropriate error handling in parsec?

parsec/scheduling.c
bosilca: The device was not active when we tried to dlsym but it is now ? If now the device is active, that's bad, the devices should have been freezed before. If the device is still inactive then this is an incorrect application and we should have detected it before. The problem is that at this point there is no easy way to bail out, we have to abort the entire app.

abouteiller: The app is correct (it has only cuda incarnations, but that is not a bug), but we have detected (late) that the app cannot run on this system (no cuda devices, before and after freeze). 

Ideally we should not call `parsec_fatal` when we capture HOOK_RETURN_ERROR (whatever the reason), but instead bubble up the error condition all the way to `taskpool_wait` to return as an rc code, but that's work for another day. When we do that, the error cases discovered here will be handled correctly. 

Last thing left is to fix the CI to use the correct tag. Thomas wanted to wait for Geri's rework of the CI, but it may be pretty easy to do now.

bosilca: As I said in my previous message, we cannot escalate this across a distributed run, so from user perspective it might not be the best solution. The user should have tested upfront if it can start a GPU-only run on the setup.

abouteiller: this all went away with the preselection logic in #648

tests/runtime/cuda/stage_main.c
abouteiller: we could replace this with parsec_query

abouteiller:
Ok so current approach is that the tests will not fail, just put out a warning in the output. 
That is not ideal but we have to choose between two bad options due to how ctest works: either we 1. cause spurious failures when running ctest when we don't have the hardware to run the test, or 2. we cause spurious success  when doing so, there is no way to report that the test actually did not run. 

This happens in CI, but also happens when running ctest manually (e.g., running with `srun -wxyz ctest` where xyz doesn't have a CUDA gpu). 

@therault has been working with Geri to integrate the workflow tags into the ctest logic, but that is currently stalled so lets move with this for now.

bosilca:
Why not returning a specific error code from the tests to indicate that the test requires hardware not available. We could then use expect to identify tests that cannot be run because the CI does not have the specific hardware (aka. the tag).

abouteiller:
In the end, ctest will report the test as failed of passed, even if the test itself has a ternary output, ctest can only report binary results. So there will be loss of information, we just have to decide how we want it rounded up (passed) or down (failed, current behavior).

abouteiller:
https://cmake.org/cmake/help/v3.19/prop_test/SKIP_RETURN_CODE.html

therault (approved):

Issue #645, opened 2024-03-13
Overflow when creating data if tile size is large


QingleiCao:
[Here](https://github.com/ICLDisco/parsec/blob/4f76b6d2fa622b5e7ba96cd09a3333415acba220/parsec/data_dist/matrix/matrix.c#L38), matrix->bsiz and parsec_datadist_getsizeoftype are integers, which may lead to overflow when the tile size is large. This case is encountered in a case where the tile size is L * (L^2/2) (L = 720) and the datatype is double complex. 

Do we need to use `size_t` to protect here or `bsiz` in `parsec_tiled_matrix_t`?

QingleiCao:
https://github.com/ICLDisco/parsec/pull/646

PR #646, opened 2024-03-15
Fix overflow when calling parsec_data_create

QingleiCao:
#645

### DIFF
--- a/parsec/data_dist/matrix/matrix.c
+++ b/parsec/data_dist/matrix/matrix.c
@@ -94,7 +94,7 @@ void parsec_tiled_matrix_init( parsec_tiled_matrix_t *tdesc,
     tdesc->tileld   = (storage == PARSEC_MATRIX_TILE) ? mb : lm;
     tdesc->mb       = mb;
     tdesc->nb       = nb;
-    tdesc->bsiz     = mb * nb;
+    tdesc->bsiz     = (size_t)mb * nb;
 
     /* Large matrix parameters */
     tdesc->lm = lm;
@@ -304,7 +304,8 @@ int parsec_tiled_matrix_data_read(parsec_tiled_matrix_t *tdesc, char *filename)
     parsec_data_t* data;
     FILE *tmpf;
     char *buf;
-    int i, j, k, ret;
+    int i, j, k;
+    size_t ret;
     uint32_t myrank = tdesc->super.myrank;
     int eltsize =  parsec_datadist_getsizeoftype( tdesc->mtype );
 
@@ -322,7 +323,7 @@ int parsec_tiled_matrix_data_read(parsec_tiled_matrix_t *tdesc, char *filename)
                     buf = parsec_data_get_ptr(data, 0);
                     ret = fread(buf, eltsize, tdesc->bsiz, tmpf );
                     if ( ret !=  tdesc->bsiz ) {
-                        parsec_warning("The read on tile(%d, %d) read %d elements instead of %d",
+                        parsec_warning("The read on tile(%d, %d) read %zu elements instead of %zu",
                                 i, j, ret, tdesc->bsiz);
                         fclose(tmpf);
                         return PARSEC_ERR_TRUNCATE;
@@ -337,8 +338,8 @@ int parsec_tiled_matrix_data_read(parsec_tiled_matrix_t *tdesc, char *filename)
                     buf = parsec_data_get_ptr(data, 0);
                     for (k=0; k < tdesc->nb; k++) {
                         ret = fread(buf, eltsize, tdesc->mb, tmpf );
-                        if ( ret !=  tdesc->mb ) {
-                            parsec_warning("The read on tile(%d, %d) read %d elements instead of %d",
+                        if ( ret !=  (size_t)tdesc->mb ) {
+                            parsec_warning("The read on tile(%d, %d) read %zu elements instead of %d",
                                     i, j, ret, tdesc->mb);
                             fclose(tmpf);
                             return PARSEC_ERR_TRUNCATE;
--- a/parsec/data_dist/matrix/matrix.h
+++ b/parsec/data_dist/matrix/matrix.h
@@ -104,7 +104,7 @@ typedef struct parsec_tiled_matrix_s {
     int tileld;         /**< leading dimension of each tile (Should be a function depending on the row) */
     int mb;             /**< number of rows in a tile */
     int nb;             /**< number of columns in a tile */
-    int bsiz;           /**< size in elements including padding of a tile - derived parameter */
+    size_t bsiz;        /**< size in elements including padding of a tile - derived parameter */
     int lm;             /**< number of rows of the entire matrix */
     int ln;             /**< number of columns of the entire matrix */
     int lmt;            /**< number of tile rows of the entire matrix - derived parameter */


devreal (changes requested):
I still think that the right approach is to make `bsiz` `size_t`. 

Also, the build failure is legit (stemming from the added assert):

```
/tmp/parsec/parsec/parsec/data_dist/matrix/matrix.c: In function 'parsec_tiled_matrix_init':
/tmp/parsec/parsec/parsec/data_dist/matrix/matrix.c:98:31: error: 'INT_MAX' undeclared (first use in this function)
   98 |     assert((size_t)mb * nb <= INT_MAX);
      |                               ^~~~~~~
/tmp/parsec/parsec/parsec/data_dist/matrix/matrix.c:19:1: note: 'INT_MAX' is defined in header '<limits.h>'; did you forget to '#include <limits.h>'?
   18 | #include "parsec/data_dist/matrix/two_dim_tabular.h"
  +++ |+#include <limits.h>
```

QingleiCao:
> I still think that the right approach is to make `bsiz` `size_t`.
> 
> Also, the build failure is legit (stemming from the added assert):
> 
> ```
> /tmp/parsec/parsec/parsec/data_dist/matrix/matrix.c: In function 'parsec_tiled_matrix_init':
> /tmp/parsec/parsec/parsec/data_dist/matrix/matrix.c:98:31: error: 'INT_MAX' undeclared (first use in this function)
>    98 |     assert((size_t)mb * nb <= INT_MAX);
>       |                               ^~~~~~~
> /tmp/parsec/parsec/parsec/data_dist/matrix/matrix.c:19:1: note: 'INT_MAX' is defined in header '<limits.h>'; did you forget to '#include <limits.h>'?
>    18 | #include "parsec/data_dist/matrix/two_dim_tabular.h"
>   +++ |+#include <limits.h>
> ```

I'm not sure whether there is potential side-effect as @abouteiller pointed out.

bosilca:
INT_MAX is not a size_t so even with the correct header the compiler will generate a warning and the comparaison will be done in the lesser precisions. You really intend to create data that is over 2GB ?

devreal:
> I'm not sure whether there is potential side-effect as @abouteiller pointed out.

Unless we pass a pointer to `bsiz` somewhere that expects an `int*` there should be no side-effect. Compiling with `-Wall` should help finding any place where that happens, or simply a `grep`.

> INT_MAX is not a size_t so even with the correct header the compiler will generate a warning and the comparison will be done in the lesser precisions.

The `int` of `INT_MAX` will be promoted to `size_t` as per the C integer promotion rules (64bit integer has higher rank than 32bit integer). That assert is legit, albeit not very helpful unless applications run with a debug build of PaRSEC.

devreal:
A quick grep showed no place where we pass a pointer to `bsiz`. There are places where we cast `bsiz` to `size_t` already:

```
parsec/data_dist/matrix/two_dim_rectangle_cyclic.c:            pos *= (size_t)dc->super.bsiz;
parsec/data_dist/matrix/two_dim_rectangle_cyclic.c:            pos *= (size_t)dc->super.bsiz;
```

and in `matrix.c` we store the result of `fread`/`fwrite` in an `int` and compare against `bsiz`. Those functions return `size_t` though, so that could be adjusted as well.

bosilca:
Rule 6.3.1.1 applies here:
> Otherwise, if the operand that has unsigned integer type has rank greater or equal to the rank of the type of the other operand, then the operand with signed integer type is converted to the type of the operand with unsigned integer type.

The conversion will always work here because we are dealing with a constant with a positive value.

QingleiCao:
> You really intend to create data that is over 2GB ?

Yeah, each timeslot includes L^2/2 ZGEMV. the size of each vector is L. Those L^2/2 number of vectors are stored in a file, so I'm loading them together. L can be 720 or 1440. If L = 720, then 720^2/2 * 720 * sizeof(complex double) = 2,985,984,000

QingleiCao:
> A quick grep showed no place where we pass a pointer to `bsiz`. There are places where we cast `bsiz` to `size_t` already:
> 
> ```
> parsec/data_dist/matrix/two_dim_rectangle_cyclic.c:            pos *= (size_t)dc->super.bsiz;
> parsec/data_dist/matrix/two_dim_rectangle_cyclic.c:            pos *= (size_t)dc->super.bsiz;
> ```
> 
> and in `matrix.c` we store the result of `fread`/`fwrite` in an `int` and compare against `bsiz`. Those functions return `size_t` though, so that could be adjusted as well.



> A quick grep showed no place where we pass a pointer to `bsiz`. There are places where we cast `bsiz` to `size_t` already:
> 
> ```
> parsec/data_dist/matrix/two_dim_rectangle_cyclic.c:            pos *= (size_t)dc->super.bsiz;
> parsec/data_dist/matrix/two_dim_rectangle_cyclic.c:            pos *= (size_t)dc->super.bsiz;
> ```
> 
> and in `matrix.c` we store the result of `fread`/`fwrite` in an `int` and compare against `bsiz`. Those functions return `size_t` though, so that could be adjusted as well.

Yes, others directly related to `bsiz` are already protected by `size_t`.

QingleiCao:
Any suggestions? Changing `bsiz` to `size_t` or this is fine as mb * nb > INT_MAX is rare?

devreal:
> Rule 6.3.1.1 applies here:
>
>> Otherwise, if the operand that has unsigned integer type has rank greater or equal to the rank of the type of the other operand, then the operand with signed integer type is converted to the type of the operand with unsigned integer type.
>
> The conversion will always work here because we are dealing with a constant with a positive value.

Let's run through this text: `INT_MAX` is a signed integer. `size_t` is an unsigned integer with rank higher than `int` (32 vs 64b on 64bit archs). 

> if the operand that has unsigned integer type has rank greater or equal to the rank of the type of the other operand

So yes, this rule applies: `size_t` is the unsigned interger type that has rank greater than `int`, which is the signed integer type with lower rank.

 > then the operand with signed integer type is converted to the type of the operand with unsigned integer type.

This requires that `INT_MAX` (int) is promoted to the type of the operand with the unsigned type (`size_t`). 

Now, if we were to compare `size_t` against `INT_MIN` we would be in trouble because promoting a negative number to an unsigned type of higher rank would create a larger number, because 2's complement. But that's not happening here and it would be a dubious thing to do in the first place.

> Any suggestions? Changing bsiz to size_t or this is fine as mb * nb > INT_MAX is rare?

Things that encode counts should be `size_t` to avoid future overflows. The check we're talking about then becomes irrelevant and we can all go home happy :) there is of course the issue of sending such a tile via MPI, which can be solved via the big count API that hopefully we an use soon. My understand though is that in @QingleiCao's case the data remains local.

bosilca:
@devreal I'm confused by your comment, because it started as a stark contradiction to my statement in the end it basically states exactly the same thing. 😕 

Anyway, this PR looks good as it is.

bosilca (approved):

devreal:
@bosilca You said:

> INT_MAX is not a size_t so even with the correct header the compiler will generate a warning and the comparison will be done in the lesser precisions.

I don't think that is what the standard says and I hope I did not convey that in my comment :)

bosilca:
If the right-hand side was of a different signedness and not a constant, where the compiler can verify that the conversion is safe, then the compiler should have warned that the conversion is unsafe.

abouteiller:
> there is of course the issue of sending such a tile via MPI, which can be solved via the big count API that hopefully we an use soon.

Indeed, that was my first reaction reading all this. The drawback is that we have to use the large count variants for everything, not just the types.

abouteiller:
> > You really intend to create data that is over 2GB ?
> 
> Yeah, each timeslot includes L^2/2 ZGEMV. the size of each vector is L. Those L^2/2 number of vectors are stored in a file, so I'm loading them together. L can be 720 or 1440. If L = 720, then 720^2/2 * 720 * sizeof(complex double) = 2,985,984,000

I'm not sure how the patch fixes your issue (at best it delays it to when you want to load a matrix that is 8x larger?), bsiz is still in int and will still be truncated for some (apparently, since less than 1 order of magnitude to the one you use now) reasonable sizes. What is our reasoning for not promoting bsiz to a size_t overall?

QingleiCao:
I will change `bsiz` to size_t. I think we also need to change the matrix size as well. I spent quite a while debugging an issue because the matrix is larger that INT_MAX.

parsec/data_dist/matrix/matrix.c
```diff
@@ -337,7 +338,7 @@ int parsec_tiled_matrix_data_read(parsec_tiled_matrix_t *tdesc, char *filename)
                     buf = parsec_data_get_ptr(data, 0);
                     for (k=0; k < tdesc->nb; k++) {
                         ret = fread(buf, eltsize, tdesc->mb, tmpf );
-                        if ( ret !=  tdesc->mb ) {
+                        if ( ret !=  (size_t)tdesc->mb ) {
```
devreal, line 341: If you change the type of `ret` you should also change the format specifier below to `%zu`

QingleiCao:
The matrix size is already cast in the high-level distribution API, e.g., parsec_matrix_sym_block_cyclic_init, so this PR is only about `bsiz`.

parsec/data_dist/matrix/matrix.c
QingleiCao: Fixed

devreal: Should `tdesc->mb` also be changed to `size_t`?

QingleiCao: `mb` should not be that large, right?

bosilca (approved):

devreal (approved):

abouteiller (approved):

PR #648, opened 2024-04-01
Compute CPU and GPU versions without lying during kernel epilog (enable TTG/PTG versioning to coexist)

abouteiller:
new: rework get_best_device to lock-in the device used (it still use the evaluate functions), this will let us consolidate version management for GPU and CPU for all DSLs

This incidentally make potrf_dtd -g 2  work (while it was suspicious before)

New: parsec_select_best_device
Removed: parsec_get_best_device

* [x] stress and stage crash when no GPU available at runtime (defer to #644)
* [x] testing_best_device computes wrong values under some cases (on b00 notably)
* [x] re-enable the PTG device=123 body decorator
* [x] compute CPU versions upfront to enable TTG
* [ ] optional: stop lying about data_out and data_in being the CPU copy all the time when it most of the time is a GPU copy that we are just passing down directly, there are complexities with updating the right copy->readers (especially in dtd)

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -3937,6 +3937,7 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
     jdf_def_list_t* type_property;
     jdf_def_list_t* dyld_property;
     jdf_def_list_t* evaluate_property = NULL;
+    jdf_def_list_t* device_property = NULL;
 
     (void)jdf;
     string_arena_add_string(sa, "static const __parsec_chore_t __%s_chores[] ={\n", base_name);
@@ -3969,8 +3970,17 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
                 }
                 string_arena_add_string(sa, "      .dyld     = \"%s\",\n", dyld_property->expr->jdf_var);
             }
-            string_arena_add_string(sa, "      .evaluate = (parsec_evaluate_function_t*)%s,\n",
-                                    (NULL == evaluate_property) ? "NULL" : evaluate_property->expr->jdf_c_code.fname);
+            jdf_find_property(body->properties, "device", &device_property);
+            if( evaluate_property && device_property ) {
+                fprintf(stdout, "Warning: Property %s defined at line %d is overruled by %s defined at line %d.\n", device_property->name, JDF_OBJECT_LINENO(device_property), evaluate_property->name, JDF_OBJECT_LINENO(evaluate_property));
+                device_property = NULL;
+            }
+            if( NULL != device_property )
+                string_arena_add_string(sa, "      .evaluate = (parsec_evaluate_function_t*)eval_hook_of_%s_%s_select_device,\n",
+                                        base_name, dev_upper);
+            else
+                string_arena_add_string(sa, "      .evaluate = (parsec_evaluate_function_t*)%s,\n",
+                                        (NULL == evaluate_property) ? "NULL" : evaluate_property->expr->jdf_c_code.fname);
             string_arena_add_string(sa, "      .hook     = (parsec_hook_t*)hook_of_%s_%s },\n", base_name, dev_upper);
             string_arena_add_string(sa, "#endif  /* defined(PARSEC_HAVE_DEV_%s_SUPPORT) */\n", dev_upper);
             free(dev_upper);
@@ -6577,7 +6587,7 @@ static void jdf_generate_code_hook_gpu(const jdf_t *jdf,
     jdf_def_list_t *prop;
     const char *dyld;
     const char *dyldtype;
-    const char *device;
+    const char *device_expr;
     string_arena_t *sa, *sa2, *sa3;
     assignment_info_t ai;
     init_from_data_info_t ai2;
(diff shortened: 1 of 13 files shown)


devreal (approved):
This works with TTG and DPLASMA on Frontier. Thanks @abouteiller

bosilca (commented):
forgot to submit my review.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -3157,7 +3157,7 @@ static void jdf_generate_startup_tasks(const jdf_t *jdf, const jdf_function_entr
             "%s    vpid = (vpid + 1) %% context->nb_vp;  /* spread the initial joy */\n"
             "%s  }\n"
             "%s  new_task = (%s*)parsec_thread_mempool_allocate( context->virtual_processes[vpid]->execution_streams[0]->context_mempool );\n"
-            "%s  new_task->status = PARSEC_TASK_STATUS_NONE;\n",
+            "%s  PARSEC_OBJ_CONSTRUCT(new_task, parsec_task_t); /* construct called only when new, force-construct it again */\n",
```
bosilca, line 3160: Why was this not necessary before ?

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -4481,7 +4491,7 @@ static void jdf_generate_startup_hook( const jdf_t *jdf )
             "    parsec_task_class_t* tc = (parsec_task_class_t*)__parsec_tp->super.super.task_classes_array[i];\n"
             "    __parsec_chore_t* chores = (__parsec_chore_t*)tc->incarnations;\n"
             "    uint32_t idx = 0, j;\n"
-            "    for( j = 0; NULL != chores[j].hook; j++ ) {\n"
+            "    for( j = 0; PARSEC_DEV_NONE != chores[j].type; j++ ) {\n"
```
bosilca, line 4494: why ? It seems that the two lead to similar outcome, but I question the need for the change.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -4670,7 +4680,7 @@ static void jdf_generate_constructor( const jdf_t* jdf )
     coutput("  for( i = 0; i < __parsec_tp->super.super.nb_task_classes; i++ ) {\n"
             "    __parsec_tp->super.super.task_classes_array[i] = tc = malloc(sizeof(parsec_task_class_t));\n"
             "    memcpy(tc, %s_task_classes[i], sizeof(parsec_task_class_t));\n"
-            "    for( j = 0; NULL != tc->incarnations[j].hook; j++);  /* compute the number of incarnations */\n"
+            "    for( j = 0; PARSEC_DEV_NONE != tc->incarnations[j].type; j++);  /* compute the number of incarnations */\n"
```
bosilca, line 4683: same comment as above. why ?

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -6741,54 +6750,69 @@ static void jdf_generate_code_hook_gpu(const jdf_t *jdf,
...
+    /* Get the hint for static and/or external gpu scheduling */
+    jdf_find_property( body->properties, "device", &device_property );
+    if ( NULL != device_property ) {
+        device_expr = dump_expr((void**)device_property->expr, &info);
+        coutput("static int eval_%s_%s_select_device(%s *this_task)\n"
+                "{\n"
+                "  __parsec_%s_internal_taskpool_t *__parsec_tp = (__parsec_%s_internal_taskpool_t *)this_task->taskpool; (void)__parsec_tp;\n"
+                "  int dev_prop = %s;\n"
+                "  int nb_devs_of_type, dev_index;\n"
+                "  if( NULL != this_task->selected_device ) {\n"
```
bosilca, line 6767: how can it be possible that the task already has a `selected_device` ?

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -6741,54 +6750,69 @@ static void jdf_generate_code_hook_gpu(const jdf_t *jdf,
...
+                "    char tmp[MAX_TASK_STRLEN];\n"
+                "    parsec_warning(\"Task %%s property device (value %%d) has been ignored \"\n"
+                "                   \"because the task already has a selected device %%d %%s\",\n"
+                "       parsec_task_snprintf(tmp, MAX_TASK_STRLEN, (parsec_task_t *)this_task), dev_prop,\n"
+                "       this_task->selected_device->device_index, this_task->selected_device->name);\n"
+                "    return PARSEC_HOOK_RETURN_DONE;\n"
+                "  }\n"
+                "  for( dev_index = nb_devs_of_type = 0; dev_index < parsec_mca_device_enabled(); dev_index++ )\n"
```
bosilca, line 6775: This code should have been into a parsec internal function. That would have simplified the entire code, because you could have just dumped the device property as a function, then call use it's pointer to call it in validation function, before doing the checks below.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -7099,28 +7128,6 @@ jdf_generate_code_complete_hook(const jdf_t *jdf,
...
-        if(JDF_FLOW_TYPE_CTL & fl->flow_flags) continue;
-        if(fl->flow_flags & JDF_FLOW_TYPE_WRITE) {
-            /**
-             * The data_out might be NULL if we don't forward anything.
-             */
-            coutput("  if ( NULL != this_task->data._f_%s.data_out ) {\n"
-                    "#if defined(PARSEC_DEBUG_NOISIER)\n"
-                    "     char tmp[128];\n"
-                    "#endif\n"
-                    "     this_task->data._f_%s.data_out->version++;  /* %s */\n"
```
bosilca, line 7112: This is a significant change, you move the version management from `_complete`, aka. after the body execution into the `_hook`, aka. for each execution. How would that work if the body returns AGAIN or NEXT ?

PR #651, opened 2024-05-16
Fix computation of available memory on gpu (avoid truncation and conversions)

abouteiller:
Fix computation of available memory on gpu (avoid truncation and conversions) and add the amount of reserved mem to the device caps output

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -575,23 +575,15 @@ parsec_cuda_module_init( int dev_id, parsec_device_module_t** module )
     }
 
     if( show_caps ) {
-        parsec_inform("GPU Device %-8s: %s [capability %d.%d] %s\n"
-                      "\tLocation (PCI Bus/Device/Domain): %x:%x.%x\n"
-                      "\tSM                 : %d\n"
-                      "\tFrequency (GHz)    : %f\n"
-                      "\tpeak Tflop/s       : %4.2f fp64,\t%4.2f fp32,\t%4.2f tf32,\t%4.2f fp16\n"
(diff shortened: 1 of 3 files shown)


devreal (approved):
LGTM

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -575,23 +575,15 @@ parsec_cuda_module_init( int dev_id, parsec_device_module_t** module )
     }
 
     if( show_caps ) {
-        parsec_inform("GPU Device %-8s: %s [capability %d.%d] %s\n"
```
bosilca, line 578: not sure I understand the need to change the device info.

parsec/mca/device/device_gpu.c
```diff
@@ -597,9 +597,9 @@ parsec_device_memory_reserve( parsec_device_gpu_module_t* gpu_device,
     int rc;
     (void)eltsize;
 
-    size_t how_much_we_allocate;
+    size_t alloc_size;
     size_t total_mem, initial_free_mem;
-    uint32_t mem_elem_per_gpu = 0;
```
bosilca, line 602: Apart renaming and reordering the code, what is the real change here ? The shift for the local variable from uint32_t to size_t ? Why is this needed ?

abouteiller: the test for how_much_we_allocate > initial_free_mem was done twice, in similar but not exactly the same way. The second time, around line 699, the use of int32_t would cause the result to be random (truncated) on some large memory systems.

parsec/mca/device/cuda/device_cuda_module.c
abouteiller: The main goal was to add the amount of actually allocated memory (and the total capacity), the rest of the reformatting is just cosmetic to reduce the number of output lines, the old format was pretty annoying on Frontier with 8GPUs per node.

bosilca (approved):

PR #652, opened 2024-05-16
cmakery: let find_package find HIP v6 

abouteiller:
and check for HIP<v5 in another way as the cmake blessed way is broken.


bosilca (approved):

devreal (approved):
Thanks @abouteiller!

PR #653, opened 2024-05-17
reenable detection of AVX capabilities even if compiler would not generate vector code

abouteiller:
Reenable detection of AVX capabilities even if compiler would not generate vectorized AVX code (because libraries like mkl can still use it). With the current status, the AVX availability on the CPU is never detected because we don't pass the necessary -march=xyz that enables AVX instruction generation.

@bosilca you made the opposite change some time ago, and I have vague recollection that there were issues with compiling the builtin checks without the ifdefs on some systems?


bosilca:
The check for AVX512 was generating instructions that did not exists in some older architectures.

We can't assume what hardware capabilities the libraries we call can handle, this check should be related to what the runtime can handle.

abouteiller:
I see, that makes the builtin test pretty useless overall.

PR #654, opened 2024-05-17
bugfix: cpu loads are not computed, hence it makes no sense to target CPUs for load balancing

abouteiller:
In some cases CPU SYRKs would execute when there is a valid GPU hook, because the CPU load (which is never updated hence is always having an eta of 0+time_estimate) would be lower than the data AREAD preferred GPU load+time_estimate. 

Since the CPU loads are always 0 this is not accurate, and we should load balance only between GPUs.

### DIFF
--- a/parsec/mca/device/device.c
+++ b/parsec/mca/device/device.c
@@ -227,6 +227,8 @@ int parsec_select_best_device( parsec_task_t* this_task ) {
             if(!(dev->type & valid_types)) continue;
             /* Skip recursive devices: time estimates are computed on the associated CPU device */
             if(dev->type == PARSEC_DEV_RECURSIVE) continue;
+            /* Skip CPU devices: we do not compute loads on CPUs correctly anyway */
+            if(dev->type == PARSEC_DEV_CPU) continue;
 
             eta = dev->device_load + time_estimate(this_task, dev);
             if( best_eta > eta ) {


devreal:
Does this mean we would never select the CPU for a task that provides both device and host incarnations?

abouteiller:
Correct, this is reverting to behavior we had prior to #648. Indeed we would never select a CPU for a task that has both types of incarnations (at least not due to 'load balancing').

Users may direct what device is used with the `evaluate` hook, forcing the use of a particular GPU or CPU device. This is not automated, but the users can access the time_estimate functions to do some informed scheduling of their own. 

In most cases using CPU devices (even with correct load balance) is detrimental to performance overall (that may change with different architectures, but atm it is almost always a slowdown). This is the rationale we had when we disabled execution on CPU devices a long time ago. 

Enabling automatic CPU/GPU load balance is (and has been forever) impossible, because we don't have a queue for CPU work. We could rework the system to have select_best_device run way earlier (e.g., when the task is queued in the scheduler) and that would let us compute load for CPUs in advance. This is significant work though and I don't have time to do it now.

abouteiller:
As for recursive tasks, I do not remember the details, but my recollection is that we have to order them in the JDF so that they are first in the hook array, otherwise they wont be considered.

If they are first, I have honestly no idea what will happen with the current code, as the recursive is altogether broken and cannot be tested.

bosilca:
1. I'm not sure I understand what brokenness you are talking about, it works for the GB tests and we are using them for Expanse tests as well. 
2. The details were that they needed to be before the CPU incarnation or they would not be considered, but not before any accelerator implementation such that they could not always be selected. I guess with the evaluate, we can highlight that we would like to use the recursive version.
3. We did have some support for load balancing between accelerators and CPU, but apparently not anymore.
4. I would understand having an option to specifically disable load balancing with the CPU (simply by putting the CPU load very high from the beginning), but willingly discarding hand-picked chores deep inside the core of the scheduling algorithm is quite radical.

abouteiller:
Discussion on Friday: 
1. loads on cpu are always 0, and we now that we can execute on the CPU device NOW, because we -are- running the select_best_device on that EU.
2. even with the rump load computation, the fact that the time-estimate on the host is lower than the time-estimate of the queued work on GPUs is indicative that some benefit *may* be possible by running on cpu
3. we could use the `balance_skew` parameter to prefer GPUs anyway (because moving the data back on CPU would incur additional overhead that is not counted in time_estimate
4. have an mca param that can disable a particular DEV_TYPE from the command line (so one may force running all on GPU, even if the code has DEV_CPU hooks). 
5. Should 4 be: GPU only by default, CPU-GPU load balance can be *opt-in*?

I think that can be done quickly

abouteiller:
The plan discussed above implemented in #663

PR #656, opened 2024-05-17
bugfix: when hip is not found, its ok.

abouteiller:


bosilca (commented):
How can HIP_VERSION be set to anything if HIP was not found ?

abouteiller:
```
VERSION_LESS: Component-wise integer version number comparison (version format is major[.minor[.patch[.tweak]]], omitted components are treated as zero). 
```
so not found is less than 5.

PR #657, opened 2024-05-17
config: osx would not find bison on newer fink/brew

abouteiller:


contrib/platforms/macosx
```diff
@@ -20,15 +20,15 @@ fi
...
+  ENVVARS+=" BISON_ROOT=${BISON_ROOT:-/opt/homebrew/opt/bison}"
+elif [ -d /usr/local/opt/bison ]; then # brew (intel)
   ENVVARS+=" BISON_ROOT=${BISON_ROOT:-/usr/local/opt/bison}"
-fi
-# Try to get the 'MacPort' Bison if installed
-if [ -x /opt/local/bin/bison ]; then
+elif [ -x /opt/local/bin/bison ]; then # macports
   ENVVARS+=" BISON_ROOT=${BISON_ROOT:-/opt/local}"
-fi
-# Try to get the 'Fink' Bison if installed
```
devreal, line 30: Can we keep this comment? I think it would be helpful for future us to know what these directories refer to (also for the two brew lines above)

bosilca (commented):
I'm really not a fan of this, because we are bypassing what the user set as his environment (basically we skip $PATH). I don't think it is PaRSEC responsibility to fiddle with this, but instead we should report and let the use fix it.

contrib/platforms/macosx
abouteiller: each line us already commented

abouteiller:
You are free to not use the platform file, it's an explicit parameter on the configure line to add it.

devreal (approved):

bosilca:
This is just random stuff, you pick an off-PATH executable to be the main bison instead of deferring to the user to do the right thing. What if I have an old brew around (maybe because I switched to ports) ?

I might be old school on this but please point me to any other software that does similar selection in their configure helper ?

therault (approved):
LGTM, but I don't have a mac to test

PR #663, opened 2024-06-28
Add: mca control for cpu load balancing (and don't report Gflops figures for cpus we can't determine)

abouteiller:
Add an mca control that disable cpu-gpu load balancing (disabled by default). This can be overriden in multiple ways:
* setting mca param will enable cpu load balancing globally
* providing an evaluate function for the CPU type can set the selected device
* data_advise_on_device can be set to a CPU, etc.

Change reporting of flops for devices for which it cannot be determined, it will now print 0.0 if we can't figure it out.


parsec/mca/device/device.c
```diff
@@ -227,6 +232,8 @@ int parsec_select_best_device( parsec_task_t* this_task ) {
             if(!(dev->type & valid_types)) continue;
             /* Skip recursive devices: time estimates are computed on the associated CPU device */
             if(dev->type == PARSEC_DEV_RECURSIVE) continue;
+            /* Is automatic CPU / GPU load balancing enabled? */
```
bosilca, line 235: This should be the first among the validity checks.

parsec/mca/device/device.c
```diff
@@ -873,16 +846,68 @@ static int cpu_weights(parsec_device_module_t* device, int nstreams)
...
+        parsec_warning("CPU Name cannot be autodetected on this machine (OSX): %s", strerror(errno));
+        goto notfound;
+    }
+    /* is it an arm64 mac? */
+    rc = sysctlbyname("hw.optional.arm64", &val, &len, NULL, 0);
+    if( 0 == rc && val ) {
+        /* vector length */
+        fp_ipc = 16;
+        dp_ipc = 8;
+        // TODO: figure out a way to obtain cpu freq on arm64 macos
```
bosilca, line 873: `hw.cpufrequency` does not work ?

parsec/mca/device/device.c
```diff
@@ -873,16 +846,68 @@ static int cpu_weights(parsec_device_module_t* device, int nstreams)
...
+        rc = sysctlbyname("hw.cpufrequency", &val, &len, NULL, 0);
+        if( rc ) {
+            parsec_warning("CPU Frequency cannot be autodetected on this machine (OSX): %s", strerror(errno));
+            goto notfound;
+        }
+        freq = val * 1e-9f;
+    }
+#endif
+
+notfound:
```
bosilca, line 903: I can see that this code is adding support for arm64, but it also replace the cross-OS AVX detection with a single OSX specific detection. How are the different AVX flavors detected for Linux (the entire detection based on __builtin_cpu_supports being removed) ?

parsec/mca/device/device.c
abouteiller: not on arm cpus, there are multiple maximum cpu frequencies on arm chips (for e-cores and p-cores) and that cannot be queried through that simple interface, overall having this work was too much work for little usefulness

parsec/mca/device/device.c
abouteiller: The builtin_cpu_support always return no support because it requires we build parsec with specific flags that we never pass. 

The linux generic discovery is done by reading proc/cpuinfo above on lines 864/837 (old/new)

devreal (approved):
LGTM

parsec/mca/device/device.c
```diff
@@ -223,6 +228,8 @@ int parsec_select_best_device( parsec_task_t* this_task ) {
             /* Skip the device if it is disabled for the taskpool */
             if(!(tp->devices_index_mask & (1 << dev_index))) continue;
             dev = parsec_mca_device_get(dev_index);
+            /* Is automatic CPU / GPU load balancing enabled? */
+            if(dev->type == PARSEC_DEV_CPU && !parsec_device_load_balance_allow_cpu) continue;
```
bosilca, line 232: This is a terrible approach. You keep the CPU chore and evaluate it for every single task just such that at the end you drop the task because `parsec_device_load_balance_allow_cpu` does not allow rebalancing ?

If there is no rebalance remove the CPU chore, such that you never have to test it again. Or if you don't want to do that, move this test up beofre calling the `eval` function and update the `valid_types` accordingly.

parsec/mca/device/device.c
bosilca: I'm not sure I understand. `__builtin_cpu_support` returns based on the current processor capabilities, which is what we want. Anyway, the OSX support via `sysctlbyname` is valid and more generic (does not require GNU compilers) so it is a step in the right direction.

parsec/mca/device/device.c
bosilca: 1. An `eval` function, user-written or not, should never select an unallowed device. Moreover, without `parsec_device_load_balance_allow_cpu` CPU should not even be an option.
2. This is not what I suggested. I suggested to move the test before the `eval` is called, such that without `parsec_device_load_balance_allow_cpu` the `eval` for CPU will never be called.
3. It would be nice to have all devices similar, no special treatment for CPU
4. What is `advise_on_device` in this context ?

bosilca (approved):

parsec/mca/device/device.c
```diff
@@ -219,14 +224,15 @@ int parsec_select_best_device( parsec_task_t* this_task ) {
 
         /* Consider how adding the current task would change load balancing
          * between devices */
+        if(!parsec_device_load_balance_allow_cpu)
+            valid_types &= ~PARSEC_DEV_CPU; /* automatic CPU / GPU load balancing disabled, remove the CPU type */
+        valid_types &= ~PARSEC_DEV_RECURSIVE; /* Recursive device time estimates are computed on the associated CPU device */
```
QingleiCao, line 229: `Recursive` will be disabled now even if `parsec_device_load_balance_allow_cpu` is true, right?

parsec/mca/device/device.c
```diff
@@ -881,36 +855,94 @@ static int cpu_weights(parsec_device_module_t* device, int nstreams)
         fp_ipc = 16;
         dp_ipc = 8;
     }
-#endif
-#endif  /* defined(__x86_64__) || defined(__i386__) */
+    else {
+        fp_ipc = 8;
+        dp_ipc = 4;
```
QingleiCao, line 860: So, it assumes the IPC is 2? then dp_ipc = 2 * FMA (2), right?

parsec/mca/device/device.c
```diff
@@ -881,36 +855,94 @@ static int cpu_weights(parsec_device_module_t* device, int nstreams)
...
     device->gflops_fp32 = fp_ipc * freq;
     device->gflops_fp64 = dp_ipc * freq;
+    /* If unset, set to 1 to avoid dividing by 0 when computing the time Estimates
+     * and mark that the time_estimates are guesses */
+    if(device->gflops_fp32 == 0)
+      device->gflops_fp32 = device->gflops_guess = 1;
+    if(device->gflops_fp64 == 0)
+      device->gflops_fp64 = device->gflops_guess = 1;
+    /* CPUs emulate these using normal fp */
+    device->gflops_fp16 = device->gflops_tf32 = device->gflops_fp32;
```
QingleiCao, line 945: why `device->gflops_fp16 = device->gflops_tf32 = device->gflops_fp32;`?

parsec/mca/device/device.c
QingleiCao: I see. This is only for time estimation.

PR #664, opened 2024-07-12
Fix the user-triggered termination test

therault:
- Make sure the parsec context is started only after the taskpool is created as the taskpool will register active message handlers for the first time via the user-triggered option in PTG
 - Correctly compute the number of local tasks instead of assuming a run at a given world size; compute that number on all ranks and check that all ranks have a correct value
 - have all MPI rank return the same return code for consistency
 - Free the taskpool when it's not used anymore to remove assertion.


devreal (approved):

bosilca:
This is breaking existing capabilities. The termination detection capability should work in all cases, not only when then context is started after the taskpool creation.

bosilca:
This issue has been already fixed by inverting the order in which the termdet was triggered with respect to the task completion.

PR #665, opened 2024-08-07
Add a function to pre-set infos for all streams/devices/...

therault:
Lazy/dynamic allocation of infos can be convenient for the users, but also detrimental to predictive behaviors. Examples of drawbacks are when testing memory constrained setups (e.g., when the test uses zone_malloc to dynamically allocate objects in the info object constructor, that may fail nondeterministically because it depends how many streams call the constructor), or small scale performance runs where the number of cublas objects to initialize vary depending on the run. Calling the new function incurs deterministic overheads.

### DIFF
--- a/parsec/class/info.c
+++ b/parsec/class/info.c
@@ -318,3 +318,18 @@ void *parsec_info_get(parsec_info_object_array_t *oa, parsec_info_id_t iid)
     }
     return ret;
 }
+
+void parsec_info_set_all(parsec_info_t *nfo, parsec_info_id_t iid)
+{
+    parsec_list_item_t *li;
+    parsec_info_object_array_t *oa;
+
+    parsec_list_lock(&nfo->ioa_list);
+    for(li =  PARSEC_LIST_ITERATOR_FIRST(&nfo->ioa_list);
+        li != PARSEC_LIST_ITERATOR_END(&nfo->ioa_list);
+        li =  PARSEC_LIST_ITERATOR_NEXT(li)) {
+            oa = (parsec_info_object_array_t*)li;
+            parsec_info_get(oa, iid);
+    }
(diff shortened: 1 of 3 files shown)


parsec/class/info.c
```diff
@@ -311,10 +314,26 @@ void *parsec_info_get(parsec_info_object_array_t *oa, parsec_info_id_t iid)
...
+    if(NULL != oa->ctx_set) oa->ctx_set(oa->ctx_set_obj);
     nio = ie->constructor(oa->cons_obj, ie->cons_data);
     ret = parsec_info_test_and_set(oa, iid, nio, NULL);
     if(ret != nio && NULL != ie->destructor) {
         ie->destructor(nio, ie->des_data);
     }
     return ret;
 }
+
+void parsec_info_set_all(parsec_info_t *nfo, parsec_info_id_t iid)
```
devreal, line 326: Suggest a different name:
```suggestion
void parsec_info_populate(parsec_info_t *nfo, parsec_info_id_t iid)
```

bosilca: Let me question the need for this function. It is only use in a single location, and based on what it is doing I don't see the need for a capability to populate the entire info set. Second issue with this populate capability, is that all memory allocation will be done in the context of the caller of the populate, instead of allowing the allocated objects to belong to the scope of the thread/stream that would use them.

In fact my main issue is with the entire info API. Because `parsec_info_set` nicely allows you to set a value for a key, but then `parsec_info_get` will magically allocate one for you even if you never set one. And there is no way to distinguish between the case where the value was properly set and the case where it was auto-generated.

abouteiller:
Oct. 25. We decided this is a useful feature but it needs a more descriptive name and maybe have functionality to pre-populate everything that needs pre-allocated

PR #668, opened 2024-09-10
Add support for batched tasks.

bosilca:
The idea is the following:
- tasks incarnations (aka. BODY) can be marked with the "batch" property allowing the runtime to provide the task with the entire list of ready tasks of the execution stream instead of just extracting the head.
- this task kernel can acquire additional tasks from the stream's ready_list, and build a ring. While the batch group will be submitted for execution (user responsibility), the rest of the tasks, that remains in the stream's ready_list, will be added back into the stream pending list, in the order in which they were provided originally.
- upon completion the user will return the ring of tasks and the runtime will merge it properly for the next stage.


parsec/mca/device/device_gpu.c
```diff
@@ -1869,15 +1870,26 @@ parsec_device_progress_stream( parsec_device_gpu_module_t* gpu_device,
...
     }
-    PARSEC_LIST_ITEM_SINGLETON((parsec_list_item_t*)task);
+    /* Should we allow the tasks to be batched */
+    if (PARSEC_GPU_TASK_TYPE_KERNEL == task->task_type ) {
+        if( PARSEC_DEV_CHORE_ALLOW_BATCH & task->ec->task_class->incarnations[0].type ) {
+            /* Don't singleton the task, allowing the kernel to extract the tasks it wants 
+             * from the task ring, and singleton it or replace it with the aggregated tasks
+             * as necessary.
+             */
+            goto move_forward_with_this_task;
```
abouteiller, line 1887: can we avoid using a goto here?

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -6983,7 +6999,7 @@ static void jdf_generate_code_hook(const jdf_t *jdf,
     coutput("#if defined(PARSEC_HAVE_DEV_%s_SUPPORT)\n", type_upper);
     if( NULL != type_property) {
 
-        if (!strcasecmp(type_property->expr->jdf_var, "cuda")
+        if (!strncasecmp(type_property->expr->jdf_var, "cuda", 4)  /* for batched */
```
abouteiller, line 7002: Looks like a leftover from a prior iteration of that patchset that used the `type=cuda_batched` instead of adding a new `batched` property.

I assume the expectation is that we can have `batched` and `non batched` CUDA bodies simultaneously. Did you test this works?

bosilca: This was indeed a left over, I will remove it.

For your second question, no we can't have two incarnations of the same device, this would be rejected by the runtime. However, the batched property is just a hint for both the user and the runtime. The runtime does not have to fulfill it in which case it will call the task with a single, singletoned, task each time. If it chooses to provide a batched call, the task will have a chain of tasks as input, and will return a chain of tasks as output. If the user chooses to ignore the batched support it can treat just the current task, and then singleton it before returning.

parsec/mca/device/device.c
```diff
@@ -325,6 +382,23 @@ int parsec_mca_device_init(void)
...
+    (void)parsec_mca_param_reg_int_name("device_cuda", "enable_batching",
+                                        "Boolean to allow batched task execution on CUDA devices",
+                                        false, false, 1, &parsec_device_cuda_enable_batching);
+    (void)parsec_mca_param_reg_int_name("device_hip", "enable_batching",
+                                        "Boolean to allow batched task execution on HIP devices",
+                                        false, false, 1, &parsec_device_hip_enable_batching);
+    (void)parsec_mca_param_reg_int_name("device_level_zero", "enable_batching",
+                                        "Boolean to allow batched task execution on Level Zero devices",
+                                        false, false, 1, &parsec_device_level_zero_enable_batching);
+#endif
```
devreal: Why not enabled by default?

parsec/mca/device/device.h
```diff
@@ -65,10 +66,13 @@ typedef struct parsec_device_base_component_2_0_0 parsec_device_base_component_t
 #define PARSEC_DEV_CUDA       ((uint8_t)(1 << 2))
 #define PARSEC_DEV_HIP        ((uint8_t)(1 << 3))
 #define PARSEC_DEV_LEVEL_ZERO ((uint8_t)(1 << 4))
+#define PARSEC_DEV_CUDA_BATCH ((uint8_t)(1 << 5))
```
devreal: Any reason why we cannot batch on other architectures?

parsec/mca/device/device_gpu.c
```diff
@@ -1936,15 +1936,30 @@ parsec_device_progress_stream( parsec_device_gpu_module_t* gpu_device,
...
+            parsec_mca_device_type_supports_batch(task->ec->selected_device->type) ) {
+            /* Don't singleton the task, allowing the kernel to extract the tasks it wants
+             * from the task ring, and singleton it or replace it with the aggregated tasks
+             * as necessary.
+             */
+            goto move_forward_with_this_task;
+        }
+    }
+#endif
+    PARSEC_LIST_ITEM_SINGLETON((parsec_list_item_t *)task);
```
devreal: That goto is ugly and easy to avoid:

```suggestion
    /* Should we allow the tasks to be batched */
    bool do_batch = false;
#if defined(PARSEC_HAVE_DEV_CAPABILITY_BATCH)
    if (PARSEC_GPU_TASK_TYPE_KERNEL == task->task_type ) {
        assert(task->ec->selected_chore >= 0);
        if( (PARSEC_DEV_CHORE_ALLOW_BATCH & task->ec->task_class->incarnations[task->ec->selected_chore].type) &&
            parsec_mca_device_type_supports_batch(task->ec->selected_device->type) ) {
            /* Don't singleton the task, allowing the kernel to extract the tasks it wants
             * from the task ring, and singleton it or replace it with the aggregated tasks
             * as necessary.
             */
            do_batch = true;
        }
    }
#endif
    if (!do_batch) {
        PARSEC_LIST_ITEM_SINGLETON((parsec_list_item_t *)task);
    }
```

tests/dsl/dtd/dtd_test_simple_gemm.c
```diff
@@ -182,55 +184,149 @@ int initialize_matrix(parsec_context_t *parsec_context, int rank, parsec_matrix_
...
+    (void)gpu_stream;
+
+    /* The whole ring was submitted on one CUDA stream and completed under one
+     * event. Move every task in the ring to the output stream so the regular
+     * GPU pop/epilog path still handles ownership, pushout, and completion one
+     * task at a time.
+     */
+    parsec_list_item_ring_merge(output_stream_ghost, &(*gpu_task)->list_item);
+    (*gpu_task)->complete_stage = NULL;
+    *gpu_task = NULL;
```
devreal: I don't like this. The DSL should be required to modify internal data structures. This should be done by the GPU manager instead.

parsec/mca/device/device_gpu.c
```diff
@@ -1573,6 +1573,74 @@ static inline parsec_list_item_t* parsec_device_push_task_ordered( parsec_list_t
...
+        rc = callback((parsec_gpu_task_t *)item, batch_head, callback_data);
+        if( rc < 0 ) {
+            parsec_list_unlock(fifo_pending);
+            return rc;
+        }
+        if( 0 == rc ) {
+            (void)parsec_list_nolock_remove(fifo_pending, item);
+            (void)parsec_list_item_ring_push(&batch_head->list_item, item);
+            nb_tasks++;
+        }
```
devreal, line 1637: It would be good to have a STOP signal that can be returned by the callback. If I know I can batch up to 20 tasks but there are hundreds in the queue then I will want to stop after 20 and not look at all of them.

devreal (changes requested):

devreal:
I might have reviewed a stale version (before the last commit). I like the approach and I think we can integrate that into TTG. The docs that were just added should be good to work this out.

parsec/mca/device/device.c
bosilca: I'm not 100% sure, but maybe because I somehow feel this doesn't bring much unless the tasks are very small, which is not the default regime where parsec is used. I'm just looking for a reason to turn it on ...

parsec/mca/device/device.h
bosilca: it's a leftover from another approach. All devices support batching, CPU and recursive will never let the user do anything meaningful. Getting access to the list of ready_tasks is difficult on these devices, so for now the capability should not be used.

parsec/mca/device/device_gpu.c
bosilca: I have no issues with goto, it works great they way they were intended to. However, the code in question got removed, so no need to have this discussion anymore.

tests/dsl/dtd/dtd_test_simple_gemm.c
bosilca: I can't do that, the manager is agnostic to whatever "batching" means for the user.

parsec/mca/device/device_gpu.c
bosilca: It does have that. Return any negative value, it will be pushed back to the user code, and there you handled it as you want.

parsec/mca/device/device.c
bosilca: I will merge all these into one and set the default to true. I revisited the code and there is no drawback not cost to do so.

parsec/mca/device/device_gpu.c
```diff
@@ -1573,6 +1573,74 @@ static inline parsec_list_item_t* parsec_device_push_task_ordered( parsec_list_t
...
+    fifo_pending = gpu_stream->fifo_pending;
+    assert(NULL != fifo_pending);
+    parsec_list_item_singleton(&batch_head->list_item);
+
+    parsec_list_lock(fifo_pending);
+    for(item = (parsec_list_item_t *)fifo_pending->ghost_element.list_next;
+        item != &fifo_pending->ghost_element;
+        item = next) {
+        next = (parsec_list_item_t *)item->list_next;
+        rc = callback((parsec_gpu_task_t *)item, batch_head, callback_data);
```
devreal, line 1628: Should screen for the same task-class as `batch_head` here first? I'm not sure that it makes sense to mix task classes in a batch and it seems error-prone to let the user screen for it.

tests/dsl/dtd/dtd_test_simple_gemm.c
```diff
@@ -182,55 +184,104 @@ int initialize_matrix(parsec_context_t *parsec_context, int rank, parsec_matrix_
...
+        PARSEC_CUDA_CHECK_ERROR("cublasDgemm_v2", status,
+                                { return PARSEC_HOOK_RETURN_ERROR; });
+
+        current_gpu_task = (parsec_gpu_task_t *)current_gpu_task->list_item.list_next;
+    } while( current_gpu_task != gpu_task );
+
+    if( verbose && batch_count > 1 ) {
+        fprintf(stderr, "Submitted %d batched GEMM tasks on GPU stream %s\n",
+                batch_count, gpu_stream->name);
+    }
```
devreal, line 284: I think having an example that actually batches GEMM calls would be helpful. Here is one (untested, written in the browser)

```suggestion
    current_gpu_task = gpu_task;
    // NOTE: assuming all tasks have the same shape. Special handling needed otherwise.
    int m, n, k, mb, nb, kb;
    double** As = malloc(batch_count*sizeof(*As));
    double** Bs = malloc(batch_count*sizeof(*Bs));
    double** Cs = malloc(batch_count*sizeof(*Cs));
    for (int i = 0; i < batch_count; ++i) {
        double *A, *B, *C;
        parsec_task_t *this_task = current_gpu_task->ec;
        struct timeval start, end, diff;
        double delta;
        double *a_gpu, *b_gpu, *c_gpu;

        parsec_dtd_unpack_args(this_task,
                               &A, &B, &C,
                               &m, &n, &k,
                               &mb, &nb, &kb);
        (void)A; (void)B; (void)C;
        As[i] = parsec_dtd_get_dev_ptr(this_task, 0);
        Bs[i] = parsec_dtd_get_dev_ptr(this_task, 1);
        Cs[i] = parsec_dtd_get_dev_ptr(this_task, 2);

        current_gpu_task = (parsec_gpu_task_t *)current_gpu_task->list_item.list_next;
    } while( current_gpu_task != gpu_task );

    gettimeofday(&start, NULL);

    status = cublasDgemmBatched(handle,
                            CUBLAS_OP_N, CUBLAS_OP_N,
                            mb, nb, kb,
                            one_device, As, mb,
                            Bs, kb,
                            one_device, Cs, mb,
                            batch_count);

    gettimeofday(&end, NULL);
    timersub(&end, &start, &diff);
    delta = (double)diff.tv_sec + (double)diff.tv_usec/1e6;
    if(verbose) {
        fprintf(stderr, "GEMM(%d, %d, %d) with tiles of %dx%d, %dx%d, %dx%d on node %d, GPU %s submitted in %g s%s\n",
                m, n, k, mb, kb, kb, nb, mb, kb,
                this_task->taskpool->context->my_rank,
                gpu_stream->name, delta,
                batch_count > 1 ? " as part of a batch" : "");
    }

    PARSEC_CUDA_CHECK_ERROR("cublasDgemmBatched", status,
                            { return PARSEC_HOOK_RETURN_ERROR; });

    if( verbose && batch_count > 1 ) {
        fprintf(stderr, "Submitted %d batched GEMM tasks on GPU stream %s\n",
                batch_count, gpu_stream->name);
    }
```

docs/doxygen/task-batching.md
```diff
@@ -0,0 +1,189 @@
...
+
+For PTG-generated tasks, use the `batch = true` body property on a device body:
+
+```c
+BODY [type=CUDA
+      batch = true
+      dyld=cublasDgemm dyldtype=cublas_dgemm_t]
+{
+    /* GPU submit body. */
+}
```
devreal, line 23: It's not clear how PTG tasks can actually batch kernel invocations. Simply stringing kernels together on the same stream won't save much.

bosilca: The same way as DTD tasks, and the same way we did two years ago for the GB submission.

tests/dsl/dtd/dtd_test_simple_gemm.c
bosilca: I'm sorry but this is not parsec role to batch gemm tasks. There is a clear way to do it, an example on how to do it. That should be more than enough for any PR !

That being said, this would be a really nice addition to DPLASMA ! Looking forward to your PR.

parsec/mca/device/device_gpu.c
bosilca: Not the runtime job to do so. We can impose whatever rules we want, but my goal here is to let the user drive this.

tests/dsl/dtd/dtd_test_simple_gemm.c
devreal: What is the benefit of "batching" kernels one by one into the same stream vs submitting them the same way we do today? I don't see how the above will yield any benefits so it's not a good example.

docs/doxygen/task-batching.md
devreal: GB submissions are not part of the docs.

bosilca: This PR has been lingering here for a very long time. Let's get it in.

devreal (approved):
I guess we'll see benefits in DPLASMA once it's picked up there (not volunteering by any means...)

tests/dsl/dtd/dtd_test_simple_gemm.c
bosilca: In the worst case we are saving a CUDA event recording and tracking per "batched" task. Something you were very supportive for the PR that saved one event for tasks "with all input arguments available". And here we might save more than one, so even in the worst case still a positive outcome.

devreal: Not quite. Here we're saving one at most per task (the event on the execution stream) but still have the input and output ones for each task. In #681 we are potentially saving two per task (the input and the output events). Combined, this can significantly reduce the number of events overall.

bosilca: There is absolutely no requirement to have the input/output event for each task. This optimization is actually complementary to the one in #681.

PR #670, opened 2024-09-10
Allow selection of a particular GPU (via the mask).

bosilca:


bosilca:
Don't merge yet I want to add support for `CUDA_VISIBLE_DEVICES`.

abouteiller:
Why is this needed? 

On master (w/o this PR) Setting the environment variable already lets one fine control the allotment of devices to the parsec processes. 

What do we gain from having a parsec-level mask that is a subsetting of the cuda one?

bosilca:
I'm not sure what you are referring to. Are you saying that if `CUDA_VISIBLE_DEVICES` is set CUDA only make the specified devices visible to parsec so we can ignore the second part of the patch ?

bosilca:
I think @abouteiller is right, CUDA_VISIBLE_DEVICES is already fulfilling this, we don't need to duplicate it at the parsec level.

PR #672, opened 2024-09-13
show-caps: don't report flops for unknown cuda devs, report peer access

abouteiller:
show-caps:
* don't report flops for unknown cuda devs (report 0.0, like cpus in #663)
* report peer access mask
* report cpu avx/simd instruction if detected available

The diff looks bigger that what it actually is because I had to move the cuda show_caps to after `all_devices_attached` to be able to report peer-access, so its mostly copy pasting from `cuda_module_init` to `all_devices_attached`

* [x] undo the change that reorders the caps, after using this PR for a while the peer-access  mask is not super relevant info

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2023 The University of Tennessee and The University
+ * Copyright (c) 2010-2024 The University of Tennessee and The University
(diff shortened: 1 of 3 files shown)


bosilca:
Right, a whole lot of code move for a small benefit. I find it strange that we now query a device capabilities in `parsec_cuda_all_devices_attached` instead of the place where the device is actually initialized and made known to the runtime.

abouteiller:
I have a followup commit that will do the same for the `ze` component but I can't test atm so I'll bring it later.

PR #673, opened 2024-09-17
Remove superfluous checks for input data in kernel_pop()

devreal:
These checks have no relevance and cause problems in TTG because the input data pointers may change.

### DIFF
--- a/parsec/mca/device/device_gpu.c
+++ b/parsec/mca/device/device_gpu.c
@@ -2139,7 +2139,6 @@ parsec_device_kernel_pop( parsec_device_gpu_module_t   *gpu_device,
         original = gpu_copy->original;
         nb_elts = gpu_task->flow_nb_elts[i];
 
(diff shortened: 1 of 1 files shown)


parsec/mca/device/device_gpu.c
```diff
@@ -2126,7 +2126,6 @@ parsec_device_kernel_pop( parsec_device_gpu_module_t   *gpu_device,
         /* We need to manage all data that has been used as input, even if they were read only */
 
         /* Make sure data_in is not NULL */
-        if( NULL == this_task->data[i].data_in ) continue;
```
therault, line 2129: In PTG, we use this to detect that this flow is a control flow. Removing this line will probably create an issue with PTGs with control flows...

devreal: I removed the line from the PR. I have to check that this doesn't collide with TTG. If it does, I'll open an issue. Should not hold up this PR.

therault (approved):

devreal:
Closing as not needed.

PR #677, opened 2024-10-15
Upstream return code of progress function device_progress_stream

devreal:
The caller (parsec_device_kernel_scheduler) needs to know whether to initiate the eviction of data if we find tasks that return something other than RETURN_DONE.


bosilca (commented):
This function reports errors asynchronously and if you want the error to make sense you need it to be associated with the task that generated it. The error you are returning here is associated with what task ?

devreal:
Yes, the error is stored in the task itself but we will never start the eviction unless we return AGAIN (https://github.com/ICLDisco/parsec/blob/master/parsec/mca/device/device_gpu.c#L2520).

PR #683, opened 2024-10-18
w2r: skip transfer if no host data is available

devreal:
This is to protect against cases where the host data disappeared (because its life ended before the eviction). This is safe on a single device but could cause race conditions on multiple devices. It's a step forward for TTG.


devreal:
This doesn't work, because we won't release the memory. I will come up with a different PR.

PR #684, opened 2024-10-18
w2r: release copy immediately if no host copy is available

devreal:
And if no readers exist on the copy.

In TTG, host-side data can disappear once its life-time ended, detaching the CPU data copy from the data. On eviction, we have no host copy to evict to so we just drop the device copy. This is not perfect as it might lead to race conditions with multiple devices but it is a step forward.


devreal:
> where we create a copy that is not attached to a data.

That's not a thing we do in TTG. What we do though is to create a data that goes out of scope before the end of the program and we detach the host copy from the data. Now the data still exists (because it is ref-counted) and has device copies but no host copy anymore.

devreal:
This is replaced by https://github.com/ICLDisco/parsec/pull/695

PR #693, opened 2024-11-01
List ordered push: search from back if lower than pivot

devreal:
Determine a pivot point (mid point of priority range) and search from front or back depending on where the new item lies. Inserting elements with highest or lowest priority becomes a constant operation and elements within the current range will on average be O(N/2).

This works best with linear distribution of priorities. For non-linear distributions the insertion will be skewed to either front or back but and the worst case will still be O(N). Average complexity would still be closer to O(N/2) and insertion with highest and lowest priorities will still be O(1).


bosilca (approved):

devreal:
Force-pushed a small correction: use `!A_HIGHER_PRIORITY_THAN_B` instead of `A_LOWER_PRIORITY_THAN_B` to put elements with same priority in the back.

therault (approved):

Issue #696, opened 2024-11-08
DTD GPU can run out of incarnations


abouteiller:
## Describe the bug

`spotrf_dtd` compiled with CUDA but ran with -g0 will cause an error to be created where it ran out of incarnations.
Correct behavior is to run on CPU incarnations, even though GPU incarnations have been compiled in.

### To Reproduce
  * parsec 46f471a9 (HEAD -> bugfix/dtd-out-of-incarnations, origin/master, origin/HEAD, master) Merge pull request #574 from abo....Bouteiller
  * dplasma 12ee91b7 (origin/master, origin/HEAD, master) Merge pull request #132 from abouteiller/feature/icl-platform      Aurelien Bouteiller   9 days ago
  * `PMIX_MCA_psec='' SLURM_TIMELIMIT=100  srun -N1  -n1  -c10  --cpu-bind=v,ldoms --pty gdb --args   tests/testing_spotrf_dtd -N $((8*1024)) -t $((8*256)) --nruns 3  -v=3   -x  -g 0  -c 1 `

## Suspected root cause

in get best device
```
(gdb)
227             if(!parsec_device_load_balance_allow_cpu)
(gdb)
228                 valid_types &= ~PARSEC_DEV_CPU; /* automatic CPU / GPU load balancing disabled, remove the CPU type */
```

PR #697, opened 2024-11-08
bugfix: dtd would not run cpu hooks when compiled with cuda

abouteiller:
bugfix: if the tp disabled gpu devices that are valid types, we need tolet best device pick a cpu device. This is however a last resort when we do not allow cpu load balancing

fixes: #696


devreal (approved):

bosilca (changes requested):
This is a nice workaround, but it is difficult to understand how it works. One need to realize that the backward look is needed in order to reach the `best_index == -1` case for the CPU. It deserves a comment !

Issue #634, opened 2024-02-07
Termination detection fault with dtd


abouteiller:
## Describe the bug

Seen only once on #321,  need to see if it also happens on master

### To Reproduce

```
29302 Command: "/apps/spacks/2023-08-14/opt/spack/linux-rocky9-x86_64/gcc-11.3.1/openmpi-4.1.5-2rgaqk2wseegpmbdbbygvwrljccjaqsk/bin/mpiexec" "-n" "4" "dsl/did/dtd_test_task_insertion"
...
 29329 dtd_test_task_insertion: /home/bouteill/parsec/dplasma/parsec/parsec/mca/termdet/local/termdet_local_module.c:114: parsec_termdet_local_termination_dete
       cted: Assertion `tp->tdm.monitor == PARSEC_TERMDET_LOCAL_TERMINATED' failed.
 29330 [leconte:4113702] *** Process received signal ***
 29340 [leconte:4113702] [ 7] /home/bouteill/parsec/dplasma/build.cuda/parsec/parsec/libparsec.so.4(+0xb042f)[0x7fa93d0fc42f]
 29341 [leconte:4113702] [ 8] /home/bouteill/parsec/dplasma/build.cuda/parsec/parsec/libparsec.so.4(parsec_release_dtd_task_to_mempool+0x32)[0x7fa93d0ce596]
 29342 [leconte:4113702] [ 9] /home/bouteill/parsec/dplasma/build.cuda/parsec/parsec/libparsec.so.4(__parsec_complete_execution+0xc6)[0x7fa93d0b7f70]
 29343 [leconte:4113702] [10] /home/bouteill/parsec/dplasma/build.cuda/parsec/parsec/libparsec.so.4(__parsec_task_progress+0x12e)[0x7fa93d0b80ca]
 29344 [leconte:4113702] [11] /home/bouteill/parsec/dplasma/build.cuda/parsec/parsec/libparsec.so.4(__parsec_context_wait+0x2ee)[0x7fa93d0b8c0a]
 29345 [leconte:4113702] [12] /home/bouteill/parsec/dplasma/build.cuda/parsec/parsec/libparsec.so.4(+0x49343)[0x7fa93d095343]
```

abouteiller:
this is somewhat rare but can be reproduced for sure with `ctest --repeat until-fail:100`

Another variant
```
[...]

PR #698, opened 2024-11-13
Fix a race condition in DTD for the local termdet

therault:
DTD uses termination_detected() to destroy the taskpool, and on_leave_wait() to reset the taskpool and reinitialize the termination detector.

The 'local' termination detector module had a race condition: it would mark the taskpool as 'terminated', then call the termination_detected() callback.

As any thread can detect the termination, a worker thread could enter taskpool destruction in parallel with the main thread calling on_leave_wait() because taskpool_state() has returned TERMINATED already.

Thus, the taskpool could be destroyed *after* its termination detector is reset, leading to asserts.

Proposed solution consists of adding an intermediary state in the termination detector, 'TERMINATING'. TERMINATING is shown as 'BUSY' to the scheduler, so the main thread does not return from the main loop and does not call on_leave_wait() while the thread that detected termination is still executing termination_detected(). TERMINATING is still a necessary intermediary step in order to ensure that only one thread calls termination_detected().

Solves issue #634

### DIFF
--- a/parsec/mca/termdet/local/termdet_local_module.c
+++ b/parsec/mca/termdet/local/termdet_local_module.c
@@ -67,14 +67,28 @@ const parsec_termdet_module_t parsec_termdet_local_module = {
     }
 };
 
-/* The local detector does not need to allocate memory:
- * we use the constants below to keep track of the state.
- * There is no need for a constant for idle, as the termdet
- * transitions directly from busy to terminated.
+/* The local detector does not need to allocate memory: we use the constants
+ * below to keep track of the state. There is no need for a constant for idle,
+ * as the termdet transitions directly from busy to terminating, and then to
+ * terminated.
+ *
+ * The local termdet is in 'terminating' state during the call to the
+ * termination_detected() function. It shows the state as 'BUSY' to the user
+ * at this time. This is used to guarantee that a single thread calls
+ * termination_detected(), yet that the main thread is not allowed to exit the
+ * progress loop and call on_leave_wait() in parallel with
+ * termination_detected(). Calling on_leave_wait() and termination_detected()
+ * in parallel leads to race conditions for some DSLs (e.g. DTD), as they
+ * want to delete the monitor in termination_detected(), and create a new one
+ * in on_leave_wait(). As on_leave_wait() is only called by the master thread
+ * once taskpool_state() has returned TERMINATED, returning BUSY until *after*
+ * termination_detected() has returned (by any thread) ensures the proper
+ * order of execution.
  */
 #define PARSEC_TERMDET_LOCAL_TERMINATED NULL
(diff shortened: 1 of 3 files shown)


abouteiller (approved):

parsec/mca/termdet/local/termdet_local_module.c
```diff
@@ -111,7 +129,7 @@ static void parsec_termdet_local_termination_detected(parsec_taskpool_t *tp)
 {
     assert( tp->tdm.module != NULL);
     assert( tp->tdm.module == &parsec_termdet_local_module.module );
-    assert( tp->tdm.monitor == PARSEC_TERMDET_LOCAL_TERMINATED );
+    assert( tp->tdm.monitor == PARSEC_TERMDET_LOCAL_TERMINATING );
```
devreal, line 132: Should this check for `TERMINATING || TERMINATED`?

devreal: Ahh never mind, we're in the process of terminating here...

devreal (approved):

QingleiCao (approved):

parsec/mca/termdet/local/termdet_local_module.c
```diff
@@ -128,11 +160,11 @@ static int parsec_termdet_local_taskpool_ready(parsec_taskpool_t *tp)
     assert( tp->tdm.monitor == PARSEC_TERMDET_LOCAL_NOT_READY );
     parsec_atomic_cas_ptr(&tp->tdm.monitor, PARSEC_TERMDET_LOCAL_NOT_READY, PARSEC_TERMDET_LOCAL_BUSY);
     PARSEC_DEBUG_VERBOSE(10, parsec_debug_output, "TERMDET-LOCAL:\tTASKPOOL %p READY", tp);
-
+    PARSEC_OBJ_RETAIN(tp);
```
abouteiller, line 163: where is the corresponding release for this?

therault: The `PARSEC_OBJ_RELEASE(tp)` corresponding to `PARSEC_OBJ_RETAIN(tp)` in `parsec_termdet_local_taskpool_ready` is at line 153 (https://github.com/ICLDisco/parsec/pull/698/files#diff-c4a4ca9c4b7ae4be5d5f60191e9f84074e2bd5533851a68ebca23afdc8545f02R153): once a taskpool is marked as `READY`, it must be terminated. That's why the check in `parsec_termdet_local_unmonitor` checks that the taskpool is either `UNMONITORED` (we haven't started it with `parsec_termdet_local_taskpool_ready`) or `TERMINATED` (we have started it, and it reached completion).

abouteiller: thanks I missed it

PR #705, opened 2024-11-18
Fix function name for parsec_atomic_fetch_add_int64

devreal:
Found trying to compile PaRSEC with nvc. Yes, it falls back to emulating atomic add. Will investigate.


devreal:
Looks like nvc needs the `-Mint128` flag for `__int128_t`. Makes sense, or something.

abouteiller (approved):

PR #706, opened 2024-11-20
Use 64bit integer when computing the ordered list pivot

devreal:
High priorities might lead to overflows an we revert back to a linear search from the front.

### DIFF
--- a/parsec/class/list.h
+++ b/parsec/class/list.h
@@ -2,6 +2,7 @@
  * Copyright (c) 2010-2023 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
+ * Copyright (c) 2024      Stony Brook University.  All rights reserved.
  */
 
 /* This file contains functions to access doubly linked lists.
@@ -14,6 +15,8 @@
 
 #ifndef PARSEC_LIST_H_HAS_BEEN_INCLUDED
 #define PARSEC_LIST_H_HAS_BEEN_INCLUDED
+
+#include <stdbool.h>
 
 #include "parsec/parsec_config.h"
 #include "parsec/class/list_item.h"
@@ -870,7 +873,12 @@ parsec_list_nolock_push_sorted( parsec_list_t* list,
         /* take the range of priorities and decide whether to iterate forward or backward */
         int tail_val = COMPARISON_VAL(_TAIL(list), off);
         int head_val = COMPARISON_VAL(_HEAD(list), off);
-        if (COMPARISON_VAL(newel, off) > ((head_val + tail_val) / 2)) {
+        int comp_val = COMPARISON_VAL(newel, off);
+        /* compute the pivot without risking overflow
+         * first, we compute the half point from the head and tail
+         * second, we account for odd numbers by adding 1 if both were odd */
+        int pivot = (head_val/2) + (tail_val/2) + (((head_val%2) + (tail_val%2))) == 2 ? 1 : 0;
+        if (comp_val > pivot) {
             /* new element is in upper half of priority range */
             parsec_list_item_t* position = PARSEC_LIST_NOLOCK_ITERATOR(list, pos,
             {


therault (approved):

parsec/class/list.h
```diff
@@ -868,9 +868,10 @@ parsec_list_nolock_push_sorted( parsec_list_t* list,
         parsec_list_nolock_push_front(list, newel);
     } else {
         /* take the range of priorities and decide whether to iterate forward or backward */
-        int tail_val = COMPARISON_VAL(_TAIL(list), off);
-        int head_val = COMPARISON_VAL(_HEAD(list), off);
-        if (COMPARISON_VAL(newel, off) > ((head_val + tail_val) / 2)) {
+        uint64_t tail_val = COMPARISON_VAL(_TAIL(list), off);
+        uint64_t head_val = COMPARISON_VAL(_HEAD(list), off);
+        uint64_t comp_val = COMPARISON_VAL(newel, off);
```
bosilca, line 873: According to the use we do of this macro the `priority` is an `int` and can be both positive or negative. By converting it to an `uint64_t` you are not correctly computing the median point, so it is not worse than the truncation that resulted from the initial code.

devreal: Fixed to use signed integer.

abouteiller:
That only partially fix the problem, as priority fields remain int32 (lp64 int) and may still get truncated at the source.

devreal:
That's a different problem. This fix targets a shortcoming of #693. Truncation at the source is not relevant here.

parsec/class/list.h
```diff
@@ -868,9 +868,10 @@ parsec_list_nolock_push_sorted( parsec_list_t* list,
         parsec_list_nolock_push_front(list, newel);
     } else {
         /* take the range of priorities and decide whether to iterate forward or backward */
-        int tail_val = COMPARISON_VAL(_TAIL(list), off);
-        int head_val = COMPARISON_VAL(_HEAD(list), off);
-        if (COMPARISON_VAL(newel, off) > ((head_val + tail_val) / 2)) {
+        int64_t tail_val = COMPARISON_VAL(_TAIL(list), off);
+        int64_t head_val = COMPARISON_VAL(_HEAD(list), off);
+        int64_t comp_val = COMPARISON_VAL(newel, off);
+        if (comp_val > ((head_val + tail_val) / 2)) {
```
abouteiller, line 874:
```suggestion
        if (comp_val > ((head_val/2 + tail_val/2 + head_val%2 + tail_val%2)) {
```

devreal:
Put on hold until @abouteiller has refactored priorities to be `int64_t` across the board. The current approach would fail again once priorities are 64bit.

therault:
That's not the issue: many tasks use MAX_INT as their priority. When that
happens, a->priority + b->priority overflows, and the test degenerates to
always pushing them at the wrong place.

devreal:
> There is no need for priorities to be 64 bits. It's a hint anyway, an application should reduce their needs to fit into that 32 bits.

Apparently DPLASMA suffers from 32bit limitations (@abouteiller). If the authors of PaRSEC don't get it right we cannot assume that outside users can get it right.

bosilca:
This PR fixes the `+` by doing the priority computations in 64 bits. Changing the priority field to 64 bits will simply allow someone to use INT64_MAX instead of INT_MAX and put us back in the same mess.

I'm well aware of the issue with DPLASMA, but it is a user issue not a conceptual issue. If there was a critical need to have such finesse in the priorities I would understand, but 1) our current schedulers are not able to handle that; 2) our apps can use alternative approaches, and 3) there seems to be more important issues to be fixed than fiddling with a non-critical issue.

devreal:
Alright, this seems to be a larger discussion then. Please review this fix in the context of current PaRSEC. It's a correct fix for a very specific problem. I don't care about DPLASMA, I really don't. We cannot afford holding up progress through endless discussions. This has been open for 2 weeks. Two weeks! Other projects make progress, we argue about 4B. Debilitating.

bosilca:
I think @abouteiller has a very good point with the alternative approach, aka instead of dividing the sum by 2, add two halves. No need for int64_t.

abouteiller:
This PR is ready to go as it is now, but we will still need to revisit the priorities being 32 bit because dplasma has  priorities overflow for 'normal' MT/NT values, and changing the formulas to fit in 32 bits is not super obvious, while the cost of having a 64 bit compare is probably 0.

abouteiller (approved):

therault (commented):
(I (approve))

PR #708, opened 2024-11-22
bugfix: when ALLOW_OVERTAKE is not defined, we still need to dup the

abouteiller:
bugfix: when ALLOW_OVERTAKE is not defined, we still need to dup the comm_ctx

Or at least that is what we need to do to replicate prior v3 behavior.

### DIFF
--- a/parsec/parsec_mpi_funnelled.c
+++ b/parsec/parsec_mpi_funnelled.c
@@ -194,6 +194,8 @@ static int mpi_funnelled_num_recv_req_in_arr = 0;
 
 #if defined(PARSEC_HAVE_MPI_OVERTAKE)
 static int parsec_param_enable_mpi_overtake = 1;
+#else
+static int parsec_param_enable_mpi_overtake = 0;  /* Default to 0 if not supported to avoid complaints about the MCA */
 #endif
 
 /* List to hold pending requests */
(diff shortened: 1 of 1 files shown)


parsec/parsec_mpi_funnelled.c
```diff
@@ -1393,11 +1393,13 @@ mpi_no_thread_enable(parsec_comm_engine_t *ce)
         MPI_Info_set(no_order, "mpi_assert_allow_overtaking", "true");
         MPI_Comm_dup_with_info((MPI_Comm) context->comm_ctx, no_order, &parsec_ce_mpi_comm);
         MPI_Info_free(&no_order);
-    } else {
-        MPI_Comm_dup((MPI_Comm) context->comm_ctx, &parsec_ce_mpi_comm);
+        /* There is no need to enable overtake for the AM communicator */
     }
-    /* There is no need to enable overtake for the AM communicator */
+    else
```
devreal, line 1398: Why do we even have the conditional here? It's an info key, just set and forget it. No downside if it's not supported.

abouteiller: we could also dup_with_info with an empty info

devreal: Why? Just set the key and be done with it. Complexity leads to errors like this.

bosilca (approved):

PR #710, opened 2024-11-25
Use red-black-tree in zone_malloc

devreal:
Currently, zone_malloc iterates linearly over its blocks, leading to O(N) complexity and significant overhead if all blocks are taken. We now use a red-black-tree to order lists of free blocks in a binary search tree. Lookup, insert, and delete are log(N). If all blocks are taken lookup is constant.

When allocating memory, we look for blocks or the requested size or the next larger size. In the latter case, the block is split and put back into the tree, potentially allocating a new node.
When releasing a block, it is either inserted directly or merged with its predecessor/successor blocks, which are then removed and reinserted as a larger block.

In the previous scheme, freeing was of O(1) complexity, which is now O(log(N)). However, the better worst-case complexity will amortize this. Plus, using a binary search tree will help reuse of same-size fragments and avoids splitting larger segements when possible.

RB trees have lower average complexity than AVL trees and are widely used (C++ std::map, Linux kernel allocator).

### DIFF
--- a/parsec/class/parsec_rbtree.c
+++ b/parsec/class/parsec_rbtree.c
@@ -23,9 +23,17 @@ PARSEC_OBJ_CLASS_INSTANCE(parsec_rbtree_node_t, parsec_list_item_t,
                           parsec_rbtree_node_construct, NULL);
 
 void parsec_rbtree_init(parsec_rbtree_t* tree, size_t offset) {
-    tree->nil = PARSEC_OBJ_NEW(parsec_rbtree_node_t);
+    PARSEC_OBJ_CONSTRUCT(&tree->nil_element, parsec_rbtree_node_t);
+    tree->nil = &tree->nil_element;
     tree->root = tree->nil;
     tree->comp_offset = offset;
+}
+
+void parsec_rbtree_fini(parsec_rbtree_t* tree) {
+    PARSEC_OBJ_DESTRUCT(&tree->nil_element);
+    tree->nil = NULL;
+    tree->root = NULL;
+    tree->comp_offset = 0;
 }
 
 static inline
--- a/parsec/class/parsec_rbtree.c
+++ b/parsec/class/parsec_rbtree.c
@@ -1,0 +1,331 @@
+#include <stdbool.h>
+
+#include "parsec/parsec_config.h"
+
+#include "parsec/class/parsec_rbtree.h"
+#include "parsec/constants.h"
+
+/* left and right children, using list pointers */
+#define LEFT(node)  (*(parsec_rbtree_node_t**)&node->super.list_prev)
+#define RIGHT(node) (*(parsec_rbtree_node_t**)&node->super.list_next)
+
+
+/**
+ * The list_item object instance.
+ */
+static inline void
+parsec_rbtree_node_construct( parsec_rbtree_node_t* item )
+{
+    item->color = PARSEC_RBTREE_BLACK;
+}
+
+PARSEC_OBJ_CLASS_INSTANCE(parsec_rbtree_node_t, parsec_list_item_t,
+                          parsec_rbtree_node_construct, NULL);
+
+void parsec_rbtree_init(parsec_rbtree_t* tree, size_t offset) {
+    tree->nil = PARSEC_OBJ_NEW(parsec_rbtree_node_t);
+    tree->root = tree->nil;
+    tree->comp_offset = offset;
+}
+
+static inline
+void parsec_rbtree_left_rotate(parsec_rbtree_t *tree, parsec_rbtree_node_t *x) {
+    parsec_rbtree_node_t *y = RIGHT(x);
+    RIGHT(x) = LEFT(y);
+    if (LEFT(y) != tree->nil) {
+        LEFT(y)->parent = x;
+    }
+    y->parent = x->parent;
+    if (x->parent == tree->nil) {
+        tree->root = y;
+    } else if (x == LEFT(x->parent)) {
+        LEFT(x->parent) = y;
+    } else {
+        RIGHT(x->parent) = y;
+    }
+    LEFT(y) = x;
+    x->parent = y;
+}
+
+static inline
+void parsec_rbtree_right_rotate(parsec_rbtree_t *tree, parsec_rbtree_node_t *y) {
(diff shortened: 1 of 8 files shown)


parsec/class/parsec_rbtree.h
```diff
@@ -0,0 +1,41 @@
+#ifndef PARSEC_RBTREE_H
+#define PARSEC_RBTREE_H
+
+#include "parsec/class/list_item.h"
+
+
+typedef enum parsec_rbtree_color_e { PARSEC_RBTREE_RED, PARSEC_RBTREE_BLACK } parsec_rbtree_color_e;
+
+typedef struct parsec_rbtree_node_t {
```
bosilca, line 9: Don't derive it from `parsec_list_item_t`, there are too many fields that are not only unnecessary, but will make a badly aligned object.

I would suggest 
```c
typedef struct parsec_rbtree_node_s {
    parsec_object_t  super;                         /**< Base object */
    volatile struct parsec_list_item_s* rbtree_left;  /**< Pointer to the left item */
    volatile struct parsec_list_item_s* rbtree_right;  /**< Pointer to the right item */
    parsec_rbtree_color_e color;
} parsec_rbtree_node_t;
```

Even with this the color enum will take 8 bytes.

parsec/utils/zone_malloc.c
```diff
@@ -149,7 +216,11 @@ void zone_free(zone_malloc_t *gdata, void *add)
     next_tid = current_tid + current_segment->nb_units;
     next_segment = SEGMENT_AT_TID(gdata, next_tid);
 
-    if( NULL != prev_segment && prev_segment->status == SEGMENT_EMPTY ) {
+    printf("zone_free current tid %d segment %p prev tid %d segment %p next tid %d segment %p\n",
```
bosilca, line 219: nope.

parsec/class/parsec_rbtree.c
```diff
@@ -0,0 +1,331 @@
...
+parsec_rbtree_node_construct( parsec_rbtree_node_t* item )
+{
+    item->color = PARSEC_RBTREE_BLACK;
+}
+
+PARSEC_OBJ_CLASS_INSTANCE(parsec_rbtree_node_t, parsec_list_item_t,
+                          parsec_rbtree_node_construct, NULL);
+
+void parsec_rbtree_init(parsec_rbtree_t* tree, size_t offset) {
+    tree->nil = PARSEC_OBJ_NEW(parsec_rbtree_node_t);
```
bosilca, line 26: Instead of allocating the `nil` you can make it point to the `tree` itself. The code only uses it to compare for end, of branches, so it should work just fine.

bosilca:
It would be interesting to see a comparison with the current allocator.

parsec/class/parsec_rbtree.h
devreal: This layout was intentional. The current layout of `list_item_t` (without debugging fluff):

```C
typedef struct parsec_list_item_s {
    parsec_object_t  super;                         /**< A list item is a @ref parsec_internal_classes_object */
    volatile struct parsec_list_item_s* list_next;  /**< Pointer to the next item */
    volatile struct parsec_list_item_s* list_prev;  /**< Pointer to the previous item */
    int32_t aba_key;                                /**< This field is __very__ special and should be handled with extreme
                                                     *   care. It is used to avoid the ABA problem when atomic operations
                                                     *   are in use and we do not have support for 128 bits atomics.
                                                     */
} parsec_list_item_t;
```

The compiler will put the color right after the aba key, no gaps. The size is the same as what you suggest. By deriving from `parsec_list_item_t` these items can be put into a list or lifo without hacking around.

bosilca: The size of what I suggested will always be the same including in debug more, because all of the unused debugging information will not even be there. I don't see the rb item put in any lists as of now, but who knows maybe one day they will.

devreal: We use the list item to store them in a lifo of free nodes to avoid reallocating them: https://github.com/ICLDisco/parsec/pull/710/files#diff-16341bb496e9d455f94a4e49f22ff87cd1894643e29364a6827b351785560a18R141

bosilca (commented):
Is this PR ready ? If not please mark it as draft.

I'm asking because I see two major issues.
1. I don't see there the zone_malloc_chunk_list_t are release (I see where they are allocated).
2. The header is not properly protected (BEGIN_C_DECLS / END_C_DECLS).
3. Any performance results to show the benefits of the RB tree ?

devreal:
Added the BEGIN_C_DECLS/END_C_DECLS guards and and fixed the deallocation (thanks for the catch!) I did some measurements on my local system. The benchmark is part of the PR now. As a result, I added in-place update of nodes, which brought down the overhead somewhat.  

Below are the absolute numbers and the speedups. Observations:

- The rbtree does introduce some overhead for regular allocations, esp when allocating segments sequentially. That's expected since it involves a lookup in the tree. That's about a 2x slowdown. The in-place node update brought that down from 3x. However, that is almost negligible with 8ns vs 4ns per allocation.
- There are three obvious benefits of the rbtree: 
  - It reduces fragmentation, making more memory available by reusing smaller allocations instead of splitting the next best chunk. That is the case for pow2, many-distinct, and find-or-larger for the linear version, which is not able to allocate all segments the rbtree version can allocate.
  - For full OOM (all segments allocated), the rbtree is faster than linear (not visible in the plots).
  - In cases of extreme fragmentation, the linear allocation is terrible because it walks the whole fragment list trying to find a matching slot. I measured up to 16k segments and we're in microsecond range but with 64GB device memory and 128kB segments size, we're looking at 512k segments. The `many-distinct` benchmark shows similar signs because of the fragmentation it tests.
 
I think the downside is limited and the upside is significant for non-uniform allocations under high memory pressure. The fragmentation is unlikely in DPLASMA (with mostly uniform tile sizes) but is much more likely in TTG because of its mixed allocation patterns and task-temporary data.

[zonemalloc_nsop.pdf](https://github.com/user-attachments/files/28036860/zonemalloc_nsop.pdf)
[zonemalloc_speedup.pdf](https://github.com/user-attachments/files/28036861/zonemalloc_speedup.pdf)

bosilca (approved):

PR #722, opened 2024-12-23
Decouple device flow count from parameters

devreal:
Make PARSEC_MAX_DEVICE_FLOWS configurable and select a proper integer type, up to int128_t. Make sure the flow mask is properly checked.

This is an attempt at allowing an increase of the number of device inputs up to 128. The use of 128bit integers for flows requires proper handling of integer types so this PR introduces `parsec_flow_mask_t` and provides macros to set and test the flow mask.

Related: #694 where the flows are separated out of the gpu task structure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,7 @@ set(PARSEC_MAX_LOCAL_COUNT 20 CACHE STRING "Number of local variables for tasks 
 set(PARSEC_MAX_PARAM_COUNT 20 CACHE STRING "Number of parameters for tasks (default 20)")
 set(PARSEC_MAX_DEP_IN_COUNT 10 CACHE STRING "Number of input flows for each task (default 10)")
 set(PARSEC_MAX_DEP_OUT_COUNT 10 CACHE STRING "Number of output flows for each task (default 10)")
+set(PARSEC_MAX_DEVICE_FLOWS ${PARSEC_MAX_PARAM_COUNT} CACHE STRING "Number of parameters for tasks (default: same as PARSEC_MAX_PARAM_COUNT)")
 
 ### PaRSEC PP options
 set(PARSEC_PTGPP_FLAGS "--noline" CACHE STRING "Additional parsec-ptgpp precompiling flags (separate flags with ';')" )
--- a/parsec/include/parsec/parsec_config_bottom.h
+++ b/parsec/include/parsec/parsec_config_bottom.h
@@ -160,6 +160,24 @@ typedef int32_t parsec_dependency_t;
 typedef int32_t parsec_dependency_t;
 #endif
 
+#if ((MAX_PARAM_COUNT <= 16) && (PARSEC_MAX_DEVICE_FLOWS <= 16))
+typedef uint16_t parsec_flow_mask_t;
+#elif ((MAX_PARAM_COUNT <= 32) && (PARSEC_MAX_DEVICE_FLOWS <= 32))
+typedef uint32_t parsec_flow_mask_t;
+#elif ((MAX_PARAM_COUNT <= 64) && (PARSEC_MAX_DEVICE_FLOWS <= 64))
+typedef uint64_t parsec_flow_mask_t;
+#elif ((MAX_PARAM_COUNT <= 128) && (PARSEC_MAX_DEVICE_FLOWS <= 128)) && defined(PARSEC_HAVE_INT128)
+typedef __int128_t parsec_flow_mask_t;
+#else
+#error Failed to find proper type for PaRSEC flow mask type. \
+       Make sure MAX_PARAM_COUNT and PARSEC_MAX_DEVICE_FLOWS \
+       is max 128 or 64 if 128bit integer are not supported.
+#endif
+
+#define PARSEC_FLOW_MASK(_id) (((parsec_flow_mask_t)1) << _id)
+#define PARSEC_CHECK_FLOW_MASK(_mask, _id) (!!(_mask & PARSEC_FLOW_MASK(_id)))
+
+
 /*
  * A set of constants defining the capabilities of the underlying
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -151,6 +151,7 @@ set(PARSEC_MAX_LOCAL_COUNT 20 CACHE STRING "Number of local variables for tasks
 set(PARSEC_MAX_PARAM_COUNT 20 CACHE STRING "Number of parameters for tasks (default 20)")
 set(PARSEC_MAX_DEP_IN_COUNT 10 CACHE STRING "Number of input flows for each task (default 10)")
 set(PARSEC_MAX_DEP_OUT_COUNT 10 CACHE STRING "Number of output flows for each task (default 10)")
+set(PARSEC_MAX_DEVICE_FLOWS ${PARSEC_MAX_PARAM_COUNT} CACHE STRING "Number of parameters for tasks (default: same as PARSEC_MAX_PARAM_COUNT)")
```
abouteiller, line 154: discussion: 
1. do we really need a second parameter here? 
2. can MAX_DEVICE_FLOWS be smaller than MAX_PARAM_COUNT or is it erroneous?
3. changing the type of MAX_PARAM_COUNT could have some adverse effects on PTG minimal execution context, due to the fragile reliance on alignment in that structure, that would need to be checked if we want to go there

devreal: I wanted to go with a separate parameter here because `MAX_PARAM_COUNT` is used in many places that I didn't want to touch and where it is not needed for my use case, including the task structure and the communication backend. I didn't want to touch that, party because it blows up all the size of the tasks unnecessarily. But with #694 this issue would be moot and we could expand this to all flows.

parsec/include/parsec/parsec_config_bottom.h
```diff
@@ -160,6 +160,24 @@ typedef int32_t parsec_dependency_t;
 typedef int32_t parsec_dependency_t;
 #endif
 
+#if ((MAX_PARAM_COUNT <= 128) || (PARSEC_MAX_DEVICE_FLOWS <= 128)) && defined(PARSEC_HAVE_INT128)
```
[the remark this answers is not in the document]
devreal: I reversed the order in which we choose the type so we pick the smallest possible type.

parsec/include/parsec/parsec_config_bottom.h
```diff
@@ -160,6 +160,24 @@ typedef int32_t parsec_dependency_t;
 typedef int32_t parsec_dependency_t;
 #endif
 
+#if ((MAX_PARAM_COUNT <= 16) || (PARSEC_MAX_DEVICE_FLOWS <= 16))
```
abouteiller, line 163: Should be `&&`,  we don't expect `MAX_DEVICE_FLOW < MAX_PARAM_COUNT` to be correct.

devreal: Good catch, fixed :+1:

abouteiller (approved):

tests/runtime/cuda/stage_custom.jdf
```diff
@@ -39,7 +39,7 @@ stage_stride_in(parsec_gpu_task_t *gtask,
     int elem_sz;
     int i;
     for(i = 0; i < task->task_class->nb_flows; i++){
-        if(flow_mask & (1U << i)){
```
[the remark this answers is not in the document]
abouteiller: Yes, this changes public interfaces, will require bump to v5

abouteiller:
I am willing to let this go in v4.1 despite the in-principle API breakage since it is backward compatible as long as you don't compile with custom parameters. 

Note that in the end, even if we were to bump the API to v5, compiling with different sizes would make the ABI of stage_in/stage_out inconsistent between. But we don't have an ABI so we don't care(?). 

Please chime in.

parsec/mca/device/device_gpu.c
```diff
@@ -1497,7 +1497,7 @@ parsec_device_data_stage_in( parsec_device_gpu_module_t* gpu_device,
 #endif
     gpu_task->sources[flow->flow_index] = candidate;  /* save the candidate for release on transfer completion */
     /* Push data into the GPU from the source device */
-    int rc = gpu_task->stage_in ? gpu_task->stage_in(gpu_task, (1U << flow->flow_index), gpu_stream): PARSEC_SUCCESS;
+    int rc = gpu_task->stage_in ? gpu_task->stage_in(gpu_task, PARSEC_FLOW_MASK(flow->flow_index), gpu_stream): PARSEC_SUCCESS;
```
bosilca, line 1500: This is extremely problematic, as we are looking to call a function that the user potentially registered. This is well worst than an API breakage, this make the API dependent on parsec build type parameters.

bosilca (commented):
I don't think we should merge this because it creates an unstable API.

Moreover, I think it is unreasonable to expect a function call with 128 arguments because at the the a task is just a delegated function call. I could understand extreme cases, but over 64 arguments (and I'm being overly generous) it is costly to manage the copies and tranfers for such tasks.

Instead, we can change all the places we use dependency masks to allow for 64 bits masks, but then keep it there. This will give us a stable API and you a little more giggle rooms with tasks with large number of inputs.

devreal:
I can see that this the API change is a problem. I'll need to find a different way of expressing this then. The argument that this is in any way connected to the number of arguments to the task function is overly simplistic and not at all relevant for TTG (we can create them inside the task and pass them as array to the kernel as long as we stay below the 2kB argument cap). I'm not arguing for 128 inputs for the giggles but because I expect them to become relevant, e.g., for combining two levels of tree traversal in 3D MRA or for implementing 6D MRA, which will require ~70 separate device memory buffers.

Closing this, will think of something else.

PR #728, opened 2025-02-04
Introduce managed objects

devreal:
Managed objects derive from `parsec_object_t` and provide a custom release function that can be set by the application. PaRSEC will fall back to calling `free` if no explicit callback is set. Managed objects are tied into the type system because `PARSEC_OBJ_RELEASE` must be able to distinguish them from regular objects. The most common managed object would be `parsec_list_item_t` as that is the basis for all container elements in PaRSEC and so we can register a release callback that stashes an object into a container from where it can be retrieved and reused later.

This is an alternative to #726 . It's less intrusive but like #726  not ABI backwards compatible but does not change the API.


abouteiller:
We did #731 instead

PR #729, opened 2025-02-04
Add release callback to parsec_class_t

### LINKED ISSUE #724: Release function for parsec_object_t
## Description

With https://github.com/ICLDisco/parsec/pull/694 the GPU task structure becomes a proper PaRSEC object and carefully avoid calling `PARSEC_OBJ_RELEASE` on it to avoid freeing it explicitly and instead add a callback to free it. The regular task structure is already an object but we don't really utilize the object management macros because we want to utilize free lists, so we have a dedicated callback for its release. Data copies have to be heap allocated because they are freed so we cannot use free lists for them.

### Describe the solution you'd like
Instead of calling `free` in `PARSEC_OBJ_RELEASE` we should unify the above approaches and add a release callback to the `parsec_object_t` that allows the entity allocating it to determine the way it should be freed. Then we can `PARSEC_OBJ_RELEASE` the object and the object will go back where it came from, e.g., a free-list. This avoids the careful dance we do with `parsec_task_t` and `parsec_gpu_task_t` and would allow the use of free lists for `parsec_data_copy_t`.

### Describe alternatives you've considered
The current approach of type-specific deallocators is fragile and does not cover all objects that could benefit from non-standard allocation methods.

devreal:
Users can derive PaRSEC types and add a custom release callback that is invoked instead of free once an object has dropped all its references. If no callback is provided, free will be called.

Fixes #724 

See also #726 and #728 for alternatives.

### DIFF
--- a/parsec/class/parsec_object.c
+++ b/parsec/class/parsec_object.c
@@ -9,6 +9,7 @@
  *                         University of Stuttgart.  All rights reserved.
  * Copyright (c) 2004-2005 The Regents of the University of California.
  *                         All rights reserved.
+ * Copyright (c) 2025      Stony Brook University.  All rights reserved.
  * $COPYRIGHT$
  *
  * Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


abouteiller:
We did #731 instead

PR #515, opened 2023-04-05
Revamp of the binding layer.

bosilca:
Rework the bindings. The main idea is to inherit the bindings from the batch scheduler, and then work from there.

PR #730, opened 2025-02-07
Disable thread binding

devreal:
The resource manager does a much better job at distributing cores than whatever it is we are doing. And running multiple processes per node is not uncommon these days.

#515 is two years and it seems to be an intractable problem. @therault has at least one recent example where the forced binding to the same cores became a footgun for PaRSEC. We can re-enable the binding once someone comes up with a solution but until then we should play nice.


bosilca (approved):

abouteiller (approved):

abouteiller:
rationale for this change in icldisco/dplasma#9

PR #731, opened 2025-02-10
Store release callback in parsec_object_t

devreal:
By default, objects created through `PARSEC_OBJ_NEW` will  be free'd in the callback, objects constructed through `PARSEC_OBJ_CONSTRUCT` will only be destructed. Users/applications can provide their own release callback through
`PARSEC_OBJ_RELEASE_CONSTRUCT`.

Mempools break the PaRSEC object model. To store the mempool origin, a list item and an offset stored in the mempool that points to the mempool from where the item was allocated. We should consider adding a type `parsec_mempool_item_t` that stores the origin mempool. Then the release callback invoked from `PARSEC_OBJ_RELEASE` can put the object back into the mempool.

Yes, this adds 8B to the `parsec_object_t` but in today's world that is negligible and is outweighed by the benefits of having a more flexible object model.

Alternative to #726 and #728.

### DIFF
--- a/parsec/class/parsec_object.h
+++ b/parsec/class/parsec_object.h
@@ -137,6 +137,7 @@ typedef struct parsec_object_t parsec_object_t;
 typedef struct parsec_class_t parsec_class_t;
 typedef void (*parsec_construct_t) (parsec_object_t *);
 typedef void (*parsec_destruct_t) (parsec_object_t *);
+typedef void (*parsec_release_t) (parsec_object_t *);
 
 
 /* types **************************************************************/
@@ -185,6 +186,7 @@ struct parsec_object_t {
 #endif  /* defined(PARSEC_DEBUG_PARANOID) */
     parsec_class_t *obj_class;            /**< class descriptor */
     volatile int32_t obj_reference_count;   /**< reference count */
+    parsec_release_t obj_release;           /**< destruct and release */
 #if defined(PARSEC_DEBUG_PARANOID)
     const char* cls_init_file_name;        /**< In debug mode store the file where the object get contructed */
     int   cls_init_lineno;           /**< In debug mode store the line number where the object get contructed */
@@ -295,6 +297,9 @@ static inline parsec_object_t *parsec_obj_new_debug(parsec_class_t* type, const 
 #define PARSEC_OBJ_SET_MAGIC_ID( OBJECT, VALUE )
 #endif  /* defined(PARSEC_DEBUG_PARANOID) */
 
+void parsec_obj_destruct(parsec_object_t * object);
+void parsec_obj_destruct_and_free(parsec_object_t * object);
+
 /**
  * Release an object (by decrementing its reference count).  If the
  * reference count reaches zero, destruct (finalize) the object and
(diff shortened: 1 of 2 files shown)


parsec/class/parsec_object.h
```diff
@@ -311,25 +317,20 @@ static inline parsec_object_t *parsec_obj_new_debug(parsec_class_t* type, const
         assert(NULL != ((parsec_object_t *) (object))->obj_class);        \
         assert(PARSEC_OBJ_MAGIC_ID == ((parsec_object_t *) (object))->obj_magic_id); \
         if (0 == parsec_obj_update((parsec_object_t *) (object), -1)) {     \
-            parsec_obj_run_destructors((parsec_object_t *) (object));       \
-            PARSEC_OBJ_SET_MAGIC_ID((object), 0);                      \
-            PARSEC_OBJ_REMEMBER_FILE_AND_LINENO( object, __FILE__, __LINE__ ); \
```
abouteiller, line 316: It's a pity loosing that debugging ability, why can't we keep this here? 

I'd say we don't need the parsec_obj_release function, and just call object->obj_release(object) where the free was

parsec/class/parsec_object.h
```diff
@@ -339,17 +340,26 @@ static inline parsec_object_t *parsec_obj_new_debug(parsec_class_t* type, const
 
 #define PARSEC_OBJ_CONSTRUCT(object, type)                             \
 do {                                                            \
-    PARSEC_OBJ_CONSTRUCT_INTERNAL((object), PARSEC_OBJ_CLASS(type));          \
+    PARSEC_OBJ_RELEASE_CONSTRUCT_INTERNAL((object), PARSEC_OBJ_CLASS(type), &parsec_obj_destruct); \
```
abouteiller, line 343: `PARSEC_OBJ_CONSTRUCT_INTERNAL_WRELEASE` (or WITH_RELEASE?) the opposite direction is confusing.

parsec/class/parsec_object.h
```diff
@@ -470,6 +480,7 @@ static inline parsec_object_t *parsec_obj_new(parsec_class_t * cls)
     if (NULL != object) {
         object->obj_class = cls;
         object->obj_reference_count = 1;
+        object->obj_release = &parsec_obj_destruct_and_free;
```
abouteiller, line 483: Won't the run_constructor overwrite this with `parsec_obj_destruct` (w/o free)?

parsec/class/parsec_object.h
devreal: Done.

parsec/class/parsec_object.h
devreal: The `parsec_obect_t` has no dedicated constructor, it's all done here.

parsec/class/parsec_object.h
devreal: > I'd say we don't need the parsec_obj_release function, and just call object->obj_release(object) where the free was

Done.

abouteiller (approved):

bosilca:
Sorry but merging this was a bad idea. This is not a proper object approach, it basically destroy the beauty of the parsec object system for a single 'it works for me case'.

devreal:
Feel free to revert. TTG is on a fork now, I'll make it work there.

PR #734, opened 2025-02-28
Scheduling flush next task

therault:
Sometimes, MADNESS tasks schedule a task A (ready) from another one (B), and B then blocks on the completion of A. Since the introduction of the `next_task` shortcut in the general schedulers, that may lead to a deadlock, because A may end up in `next_task`, and no other free thread will ever execute A. MADNESS makes sure there is enough threads to always execute A, if A is stealable, and MADNESS can detect when this happen and call a function to ensure A becomes stealable. This PR just provides the API that MADNESS can call to unlock that situation.

### DIFF
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -378,6 +378,16 @@ int __parsec_schedule_vp(parsec_execution_stream_t* es,
     return ret;
 }
 
+int __parsec_schedule_flush_private( parsec_execution_stream_t* es )
+{
+    parsec_task_t* task = es->next_task;
+    if( NULL != task ) {
+        es->next_task = NULL;
+        return __parsec_schedule(es, task, 0);
+    }
+    return PARSEC_SUCCESS;
+}
+
 /**
  * @brief Reschedule a task on some resource.
  *
--- a/parsec/scheduling.h
+++ b/parsec/scheduling.h
@@ -69,6 +69,23 @@ int __parsec_schedule_vp( parsec_execution_stream_t*,
                           int32_t distance);
 
 /**
+ * Some runtime systems (e.g. MADNESS) that use PaRSEC as a task scheduler may detect
+ * late that a task that already scheduled new work may block on the
+ * completion of that work (too late to mark the new task as a potential
+ * blocker). That new task, however, may have been stored in the private
+ * queue of this execution stream, and since it's not going to be scheduled
+ * until the current task is completed, the system is unable to progress.
+ *
+ * This function tells the scheduler that any enabled tasks that were reserved
+ * for private execution must be scheduled by another thread.
+ *
+ * @param[in] es The execution stream whose private queue must be flushed.
+ * @return the same return codes as __parsec_schedule (PARSEC_SUCCESS in
+ *    case of success)
+ */
+int __parsec_schedule_flush_private( parsec_execution_stream_t* es );
+
+/**
  * @brief Reschedule a task on the most appropriate resource.
  *
  * @details The function reschedules a task, by trying to locate it as closer


evaleev:
@therault thanks, will test.

evaleev:
@therault success, thanks!

bosilca (commented):
I don't like this for at least two reasons. First and foremost, the underlying assumptions of a task programming model is that in order to avoid potential deadlocks nothing should be blocking. Breaking this assumption will always lead to deadlock, whatever layers of bandaid you are trying to add. Imagine the device thread doing this, or a complicated scenario in which all CPU threads are all in blocking tasks.

Second, from a purely practical point of view why is there a need for another internal API ? Based on the PR description I assume that you add the tasks into parsec via `__parsec_schedule_vp` (because `__parsec_schedule` does not add to the `next_task`). Instead of adding a new API you could just alter `__parsec_schedule_vp` to prevent it from adding a task to `next_task` when the distance is not 0, and call it with a non-zero distance. This makes more sense as the current thread will not be able to execute it, the distance should not be 0.

```diff
diff --git a/parsec/scheduling.c b/parsec/scheduling.c
index 55657cc36..a19645f0d 100644
--- a/parsec/scheduling.c
+++ b/parsec/scheduling.c
@@ -319,7 +319,7 @@ __parsec_schedule(parsec_execution_stream_t* es,
  * If the provided execution stream is NULL, all tasks are delivered to their
  * respective vp.
  *
- * Beware, as the manipulation of next_task is not protected, an exeuction
+ * Beware, as the manipulation of next_task is not protected, an execution
  * stream should never be used concurrently in two call to this function (or
  * a thread should never `borrow` an execution stream for this call).
  */
@@ -337,7 +337,7 @@ int __parsec_schedule_vp(parsec_execution_stream_t* es,
     assert( (NULL == es) || (parsec_my_execution_stream() == es) );
 #endif  /* defined(PARSEC_DEBUG_PARANOID) */
 
-    if( NULL == es || !parsec_runtime_keep_highest_priority_task || NULL == es->scheduler_object) {
+    if( NULL == es || !parsec_runtime_keep_highest_priority_task || distance || NULL == es->scheduler_object) {
         for(int vp = 0; vp < es->virtual_process->parsec_context->nb_vp; vp++ ) {
             parsec_task_t* ring = task_rings[vp];
             if( NULL == ring ) continue;
```

evaleev:
> I don't like this for at least two reasons. First and foremost, the underlying assumptions of a task programming model is that in order to avoid potential deadlocks nothing should be blocking. Breaking this assumption will always lead to deadlock, whatever layers of bandaid you are trying to add. Imagine the device thread doing this, or a complicated scenario in which all CPU threads are all in blocking tasks.

In the app the issue occurs in the main thread. Because main can create tasks it becomes involved in the machinery without knowing it. Then it calls MPI_Comm_split and we need to ensure that we flush the sticky task (if any).

It's reasonable to mandate that tasks do not block. I don't think it's reasonable to mandate that any thread that creates tasks  cannot make blocking function calls.

> Second, from a purely practical point of view why is there a need for another internal API ? Based on the PR description I assume that you add the tasks into parsec via `__parsec_schedule_vp` (because `__parsec_schedule` does not add to the `next_task`). Instead of adding a new API you could just alter `__parsec_schedule_vp` to prevent it from adding a task to `next_task` when the distance is not 0, and call it with a non-zero distance. This makes more sense as the current thread will not be able to execute it, the distance should not be 0.

The choice of how MADNESS drives PaRSEC can be adjusted. I assume, however, that the current default is the recommended way. Is it not?

bosilca:
@evaleev that's a different use case than what @therault described in the PR.

If we are talking about the main thread then the correct answer is to always push the task away from its execution context, especially when there is no guarantee that it will stay around to execute them. The solution I proposed above will cover this case, assuming the main thread pushes the tasks at distance at least 1.

parsec does not mandate anything about the main thread. In fact we use a similar approach when inserting asynchronous dags (which are generated starting from a single task), with the difference that we push the generator task onto the system queue (which saves us from the next_task optimization). However, that is internal to parsec itself, we need a better way for external users, and I think that what I propose here make sense and remains more user friendly than adding a new API.

evaleev:
> If we are talking about the main thread then the correct answer is to always push the task away from its execution context, especially when there is no guarantee that it will stay around to execute them. The solution I proposed above will cover this case, assuming the main thread pushes the tasks at distance at least 1.

So replacing `0` with `1` in https://github.com/m-a-d-n-e-s-s/madness/blob/master/src/madness/world/parsec.cc#L296 when invoked from main would do? It feels like a hack though, why should task creation from the main thread be different and possibly be _less_ efficient than from other threads? Only very specific parts of code need to call blocking MPI collectives, most of the code in MADNESS/TiledArray only contains nonblocking code, interspersed with fences (nonblocking barriers). Having the ability to "oops, I really need to make a (collective) call" and flush the sticky tasks is the right solution, in my opinion.

P.S. MADNESS also can use TBB and its own thread pool ("Pthreads"). The latter buffers tasks before submission, and similarly need buffer flushing before collective calls. TBB has per-thread queues but steals without limits as far as I know so I don't think it needs the flush mechanism.

evaleev:
For the record, even in TBB it seems if one knows that thread will not execute tasks it's faster to use different API for submission which avoids thread-local q: https://github.com/uxlfoundation/oneTBB/issues/1354

bosilca:
Yes the solution you propose (aka. changing 0 to 1 when called from main thread) is not a hack but the correct approach, at least when you know the main thread will follow with some blocking operation. Another approach would be to call `__parsec_schedule_vp` with a NULL execution stream, to push the new tasks away from the current thread.

Adding a new API that needs to be used following the existing API translates to : "here is a task do some smart scheduling followed by undo that smartness just give it to someone else". Having control over the distance where you insert the task allows for flexible user-driven solutions. Clean, no hack.

evaleev:
> The latency for the main thread waiting for others to work steal would be high, as they have their own worrk queues and will prioritize whatever is in there. From that perspective pushing the tasks at `distance > 0` would make more sense as the tasks will not need to be stolen, they will directly be added to the some other thread queue (potentially based on their priority).

OK, that actually probably makes much sense for main thread anyway.

> Sorry, but the flush as implemented here makes little sense, it is a bandaid for the runtime to solve a user-introduced deadlock. Because, the current bandaid only pushes away the `next_task` leaving all the other queues intact, waiting for other threads to eventually steal the work. If you really intend to have a flush, all the queues of the main thread execution stream should be emptied, to leave no opportunity for high priority tasks to linger around until some starvation kicks in forcing the work stealing.

Just will note that this did fix the deadlocks due to the next_task never having a chance to execute. In terms of what the flush needs to do, I'm not sure honestly what the best practice is. It does appear that TBB does not have unstealable tasks, so at least it's not possible to deadlock due to thread blocking while having a nonnull next_task.

> How is that the main thread suddenly discovers that it has to do a blocking communication ?

instead of main thread, it's me who runs into a deadlock and eventually figured out what's going on. Luckily the Pthreads ("default") backend in MADNESS requires manual flushing of prebuffer, and I figured PaRSEC perhaps had somethhing similar (i.e., next_task)

devreal:
The next_task implementation is broken. It's another level in the hierarchy of queues but completely separate so no other thread can steal from it. It's also for a very narrow view of how tasks are discovered (at the very end of predecessor tasks), which may be true for PTG but does not hold in general. 

Either remove it (to be safe) or make it a proper level in the scheduler hierarchy so that other threads can steal from it (after going through the other levels). Did anyone show significant performance benefits from next_task in the first place?

evaleev:
> Yes the solution you propose (aka. changing 0 to 1 when called from main thread) is not a hack but the correct approach, at least when you know the main thread will follow with some blocking operation. Another approach would be to call `__parsec_schedule_vp` with a NULL execution stream, to push the new tasks away from the current thread.

tried both, neither works:
- calling `__parsec_schedule_vp` with NULL execution stream segfaults due to null ptr dereferencing https://github.com/ICLDisco/parsec/blob/master/parsec/scheduling.c#L331 ... seems like this is never tested anymore.
- setting `distance` to 1 still causes the use of `next_task`, via https://github.com/ICLDisco/parsec/blob/master/parsec/scheduling.c#L363 (as `next_task` is NULL)

bosilca:
@devreal almost correct, the intent behind `next_task` was to prioritize the data reuse as much as possible by preventing other threads from stealing a task that will reuse the data generated by the predecessor. And yes it assumes that all successors that are ready to be executed are generated in a single block, because it is more efficient than discovering them one by one. I guess this is part of the set of optimizations that set parsec apart from the other runtimes, allowing it to focus on being able to support tasks with very small granularity. Anyway, we did look into the performance benefit of this feature, and while I don't recall the exact numbers it was significant enough to keep it.

@evaleev using the distance set to 1 requires the small patch I posted above.

Anyway, I take back my complaint about this feature, as after a deeper introspection I can see other cases where it can be necessary, such as when a thread start driving an accelerator or other devices (the IO device).

evaleev:
Ok, thanks everyone!

evaleev:
> @evaleev using the distance set to 1 requires the small patch I posted above.

Missed that. Will make a patch.

Q: is it the recommended way then for the main thread to always submit tasks with `distance=1` to reduce the latency of task execution since the main thread is only used for task processing during fences?

bosilca:
Based on this I think the correct solution for TTG is to never use `__parsec_schedule_vp` but instead use `__parsec_schedule`. With this approach (which is also the case for DTD task insertion) we don't even need #736 to have it work.

PR #736, opened 2025-03-06
Cleanup the __parsec_schedule_vp to make it do what it was supposed to.

bosilca:
Basically, the next_task can only be set if the submission stream is not NULL and the distance is 0. In all other cases, the tasks will be pushed on the first execution stream of each VP, with the exception of the local VP (the VP of the submission execution stream) where the tasks are pushed locally on the execution stream.

This PR is a followup to the discussion in #734.


therault (approved):

therault (approved):

bosilca:
I should not merge my own PR without the proper vetting. This PR needs a second review.

abouteiller (approved):

PR #739, opened 2025-03-12
Issue parsec_fatal when the datatype_arenas have not been set (rather

abouteiller:
than crashing in a cryptic way), examplified by  icldisco/dplasma#138

### DIFF
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2024 The University of Tennessee and The University
+ * Copyright (c) 2009-2025 The University of Tennessee and The University
  *                         of Tennessee Research Foundation.  All rights
  *                         reserved.
  * Copyright (c) 2023      NVIDIA CORPORATION. All rights reserved.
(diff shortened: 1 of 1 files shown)


therault (approved):

devreal (approved):

bosilca (changes requested):
Why here in the iterator ? This will bail out on the first flow without a datatype, and you should only bail out if at the end of the get_datatype you don't get a proper datatype, aka in remote_dep_mpi.c:934.

bosilca:
It should not matter until the end of the iterator. We don't have to police every dep type, only that the end we have one.

devreal:
Is it ever valid to have a non-zero count with an invalid datatype? How? Otherwise this error seems legit.

bosilca:
Define valid in this particular context ? From the communication perspective it is invalid to have a non-zero count with an invalid datatype, but the communication is bound to a flow in the receive case not to any particular deps inside the flow. This PR checks for each dep, even the ones that will not finally be associated with the communication, instead of checking only once at the end to make sure the flow (and therefore the communication to be issued) is correct.

Issue #458, opened 2022-11-11
Handle kernels where [type=CPU]


cflinto:
## Description
In a jdf file, you can specify the type of kernel you want by adding [type=ARCH].
If you declare [type=CPU], the generated program will crash with very little available info.
I.e. in __parsec_execute(scheduling.c:179) because the chore is declared as so:
```
static const __parsec_chore_t __TP_MyTask_chores[] ={
#if defined(PARSEC_HAVE_CPU)
    { .type     = PARSEC_DEV_CPU,
      .dyld     = NULL,
      .evaluate = NULL,
      .hook     = (parsec_hook_t*)hook_of_LBM_WriteBack_CPU },
#endif  /* defined(PARSEC_HAVE_CPU) */
    { .type     = PARSEC_DEV_NONE,
      .evaluate = NULL,
      .hook     = (parsec_hook_t*)NULL },  /* End marker */
};
```
And PARSEC_HAVE_CPU is not defined.
The main problem is that it is very natural for new users to declare [type=CPU], yet they have very few clues to help them understand the cause of the error.

### Describe the solution you'd like
We should define PARSEC_HAVE_CPU or have some special case in parsec-ptgpp.
There should be an error of no chore has been declared.

PR #740, opened 2025-03-13
feature: support [type=CPU] bodies

abouteiller:
fixes: #458

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -3946,12 +3946,12 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
         jdf_find_property(body->properties, "type", &type_property);
         jdf_find_property(body->properties, "dyld", &dyld_property);
         jdf_find_property(body->properties, JDF_BODY_PROP_EVALUATE, &evaluate_property);
-        if( NULL == type_property) {
+        if( NULL == type_property || !strcasecmp("CPU", type_property->expr->jdf_var)) {
             string_arena_add_string(sa, "#if defined(PARSEC_HAVE_DEV_CPU_SUPPORT)\n");
             string_arena_add_string(sa, "    { .type     = PARSEC_DEV_CPU,\n");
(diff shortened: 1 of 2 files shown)


parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -6997,12 +6997,8 @@ static void jdf_generate_code_hook(const jdf_t *jdf,
     ai.holder = "this_task->locals.";
     ai.expr = NULL;
 
-    if(NULL == type_property)
-        coutput("static int %s(parsec_execution_stream_t *es, %s *this_task)\n",
-                name, parsec_get_name(jdf, f, "task_t"));
-    else
-        coutput("static int %s_%s(parsec_execution_stream_t *es, %s *this_task)\n",
-                name, type_property->expr->jdf_var, parsec_get_name(jdf, f, "task_t"));
+    coutput("static int %s_%s(parsec_execution_stream_t *es, %s *this_task)\n",
```
devreal, line 7000: Is this assuming that we always have a CPU body?

bosilca (commented):
We already have support for CPU. Look at tests/dsl/ptg/recursive.jdf. Look at the generated code and you will see the incarnation is properly protected using PARSEC_HAVE_DEV_CPU_SUPPORT, which is defined in the parsec_options.h.

abouteiller:
I saw that, but when testing (adding the case in stage), we were generating different functions signatures depending on the way we create the cpu body: 
* when [type=cpu] would be passed, we would create a function body_of_xyz_CPU, and look for 'dyld' functions etc, as-if it where a GPU body, that would not compile because the function names in `.hook=body_of_xyz` and the  actually declareed function `body_of_zyx_CPU` would not match. (recursive tester is disabled for other reasons, as recursive is broken, hence why we didn't see the issue in CI). 
* when no type is provided, we would generate a different signature (as describe above), also we do not read the dyld, and some different behavior w.r.t. evaluate functions etc. 

Overall, it is cleaner that we generate the same body form for both [type=CPU] and no-type cases, instead of going through the GPU body generation in one case and the CPU body generation in the other.

devreal (approved):
Users seem to ask for `[type=CPU]` so let's just provide it.

PR #747, opened 2025-03-28
BODY without a type will be considered as CPU.

### LINKED ISSUE #458: Handle kernels where [type=CPU]

bosilca:
This was already the case in most of the code but some corner cases where skipped such as:
  - don't allow for two bodies with the same tyoe (including the case where one of the bodies is untyped)
  - protect all the inlined functions with the BODY type.

Provides a more complete alternative approach to #740.

Fixes #458.

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -3939,6 +3939,7 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
     jdf_def_list_t* dyld_property;
     jdf_def_list_t* evaluate_property = NULL;
     jdf_def_list_t* device_property = NULL;
+     char* dev_upper;
 
     (void)jdf;
     string_arena_add_string(sa, "static const __parsec_chore_t __%s_chores[] ={\n", base_name);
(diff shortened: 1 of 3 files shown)


parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -3947,45 +3948,40 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
...
+        if( NULL == dyld_property ) {
+            string_arena_add_string(sa, "      .dyld     = NULL,\n");
+        } else {
+            jdf_def_list_t* dyld_proptotype_property;
+            jdf_find_property(body->properties, "dyldtype", &dyld_proptotype_property);
+            if ( NULL == dyld_proptotype_property ) {
+                fprintf(stderr,
+                        "Internal Error: function prototype (dyldtype) of dyld function (%s) is not defined in %s body of task %s at line %d\n",
+                        dyld_property->expr->jdf_var, type_property->expr->jdf_var, f->fname, JDF_OBJECT_LINENO( body ) );
+                assert( NULL != dyld_proptotype_property );
```
QingleiCao, line 3966: It's not for this PR, but I was wondering what the purpose of dyld and dyld_fn is? For my understanding, type, evaluate and hook are enough. 😀

bosilca: They have different meaning.
- `dyld_fn` only enables the BODY it is associated with if that particular symbol exists in the application space when the JDF is loaded. It allowed us to make distinction between different versions of the CUDA library.
- `dyld` is a replacement function for the BODY. Same API but provided by another shared library. At execution time instead of executing the current BODY it will execute the function specified in this `dyld`. This features allows for empty BODY, that are provided dynamically.

abouteiller (approved):

PR #759, opened 2026-02-23
Add AUTO in PTG which dynamically allocate suitable buffer to receive…

QingleiCao:
… data

### DIFF
--- a/parsec/remote_dep_mpi.c
+++ b/parsec/remote_dep_mpi.c
@@ -19,6 +19,8 @@
 #include "parsec/interfaces/dtd/insert_function_internal.h"
 #include "parsec/remote_dep.h"
 #include "parsec/class/dequeue.h"
+#include "parsec/mca/device/device.h"
+#include "parsec/mca/device/device_gpu.h"
 
 #include "parsec/parsec_binary_profile.h"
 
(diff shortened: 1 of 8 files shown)


parsec/remote_dep_mpi.c
```diff
@@ -571,10 +616,152 @@ void parsec_remote_dep_memcpy(parsec_execution_stream_t* es,
...
 }
 
+static void remote_dep_auto_fallback_copy_release(parsec_data_copy_t *copy, int device)
+{
+    (void)device;
+    free(copy->device_private);
+    copy->device_private = NULL;
+}
+
+static void remote_dep_auto_gpu_copy_release(parsec_data_copy_t *copy, int device)
```
bosilca: The name is confusing as it implies you are releasing the data copy but instead you are just releasing the private memory owned by it.

Why do you need the `device` ? A coipy is always associated with a device, and it would not make sense to release the memory of a different copy than the one you pass as argument.

parsec/remote_dep_mpi.c
```diff
@@ -760,6 +947,18 @@ remote_dep_mpi_retrieve_datatype(parsec_execution_stream_t *eu,
...
 
+    if( PARSEC_REMOTE_DEP_AUTO_ALLOC == output->data.remote.arena ) {
+        uint64_t remote_size = output->data.remote.src_count;
+        output->data.remote.src_datatype = PARSEC_DATATYPE_PACKED;
+        output->data.remote.dst_datatype = PARSEC_DATATYPE_PACKED;
+        output->data.remote.src_count = remote_size;
+        output->data.remote.dst_count = remote_size;
+        output->data.remote.src_displ = 0;
+        output->data.remote.dst_displ = 0;
+        output->data.remote.device_index = remote_dep_auto_pick_gpu_device(newcontext);
```
bosilca, line 979: We have a proper accessor for the task class to identify the device it wants to use. You should be calling the `newcontext->taskclass.data_affinity`

parsec/remote_dep_mpi.c
```diff
@@ -571,10 +616,152 @@ void parsec_remote_dep_memcpy(parsec_execution_stream_t* es,
...
+}
+
+static int remote_dep_auto_pick_gpu_device(const parsec_task_t *task)
+{
+    if( 0 == parsec_param_auto_gpu_enable ) {
+        return 0;
+    }
+    parsec_task_t tmp = *task;
+    tmp.selected_device = NULL;
+    if( PARSEC_SUCCESS != parsec_select_best_device(&tmp) ) {
```
bosilca: Are you sure this returns what you expect ? For me `parsec_select_best_device` only works once the task is properly initialized and has all it's input data copies properly set. Before that it will basically count the possible incarnations and returns the ownership of any of the copies. In other words it is mostly ransom or maybe always zero if the input copies are not yet set (which is the most likely).

bosilca (commented):
I understand what you are trying to do here, and I think it is a good step in the right direction. However, I have two issues with the proposed approach:
1. it loses the type of the incoming data. You are forced to drop the type and rollback to receive data as packed (potentially breaking heterogeneity). I understand that the arena must be set to `AUTO` but I think in that case the remote_type should be set to the expected type of the data instead of allowing it to be `void*`.
2. I am really not sure about the way you compute the device_index ? Without running the code, I would think you should mostly get 0, as the task's inputs are not set during the get_datatype function.

QingleiCao:
> I understand what you are trying to do here, and I think it is a good step in the right direction. However, I have two issues with the proposed approach:
> 
> 1. it loses the type of the incoming data. You are forced to drop the type and rollback to receive data as packed (potentially breaking heterogeneity). I understand that the arena must be set to `AUTO` but I think in that case the remote_type should be set to the expected type of the data instead of allowing it to be `void*`.
> 2. I am really not sure about the way you compute the device_index ? Without running the code, I would think you should mostly get 0, as the task's inputs are not set during the get_datatype function.

Thank you for your comments @bosilca. This is still a draft, and I’ll be making updates intermittently. 😄

PR #762, opened 2026-03-28
Shift GPU task completion

devreal:
Completion is potential costly (discovering new tasks) and might hit the network so we better move that away from the device management thread.

### DIFF
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -4497,6 +4497,7 @@ static void jdf_generate_startup_hook( const jdf_t *jdf )
             "      idx++;\n"
             "    }\n"
             "    chores[idx].type     = PARSEC_DEV_NONE;\n"
+            "    chores[idx].flags    = PARSEC_CHORE_FLAG_NONE;\n"
             "    chores[idx].evaluate = NULL;\n"
             "    chores[idx].hook     = NULL;\n"
             "    /* Create the initialization tasks for each taskclass */\n"
--- a/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
+++ b/parsec/interfaces/ptg/ptg-compiler/jdf2c.c
@@ -3954,6 +3954,7 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
         }
         string_arena_add_string(sa, "#if defined(PARSEC_HAVE_DEV_%s_SUPPORT)\n", dev_upper);
         string_arena_add_string(sa, "    { .type     = PARSEC_DEV_%s,\n", dev_upper);
+        string_arena_add_string(sa, "      .flags    = PARSEC_CHORE_FLAG_NONE,\n");
         if( NULL == dyld_property ) {
             string_arena_add_string(sa, "      .dyld     = NULL,\n");
         } else {
@@ -3986,7 +3987,7 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
     } while (NULL != body);
     string_arena_add_string(sa,
                             "    { .type     = PARSEC_DEV_NONE,\n"
-                            "      .evaluate = NULL,\n"
+                            "      .flags    = PARSEC_CHORE_FLAG_NONE,\n"
                             "      .hook     = (parsec_hook_t*)NULL },  /* End marker */\n"
                             "};\n\n");
 }
--- a/parsec/mca/device/device_gpu.c
+++ b/parsec/mca/device/device_gpu.c
@@ -2499,6 +2499,16 @@ parsec_device_kernel_cleanout( parsec_device_gpu_module_t *gpu_device,
 }
 
 /**
+ * Returns true if the task's completion should be shifted to worker threads.
+ */
+static bool shift_completed_task(parsec_device_gpu_module_t* gpu_device, parsec_gpu_task_t* gpu_task)
+{
+    parsec_task_t* this_task = gpu_task->ec;
+    const __parsec_chore_t *chore = &this_task->task_class->incarnations[gpu_device->super.device_index];
+    return (bool)(chore->flags & PARSEC_CHORE_FLAG_SHIFT_COMPLETION);
+}
+
+/**
  * This version is based on 4 streams: one for transfers from the memory to
  * the GPU, 2 for kernel executions and one for transfers from the GPU into
  * the main memory. The synchronization on each stream is based on GPU events,
@@ -2703,11 +2713,17 @@ parsec_device_kernel_scheduler( parsec_device_module_t *module,
         goto remove_gpu_task;
     }
     parsec_device_kernel_epilog( gpu_device, gpu_task );
-    // ship the task to other threads to complete its execution
-    gpu_task->ec->status = PARSEC_TASK_STATUS_COMPLETE;
-    PARSEC_LIST_ITEM_SINGLETON(gpu_task->ec);
-    __parsec_schedule(es, gpu_task->ec, 1);
+
+    if (shift_completed_task(gpu_device, gpu_task)) {
+        // ship the task to other threads to complete its execution
+        gpu_task->ec->status = PARSEC_TASK_STATUS_COMPLETE;
+        PARSEC_LIST_ITEM_SINGLETON(gpu_task->ec);
+        __parsec_schedule(es, gpu_task->ec, 1);
+    } else {
+        __parsec_complete_execution( es, gpu_task->ec );
+    }
     gpu_device->super.executed_tasks++;
+
  remove_gpu_task:
     PARSEC_DEBUG_VERBOSE(10, parsec_gpu_output_stream, "GPU[%d:%s]: gpu_task %p freed",
                          gpu_device->super.device_index, gpu_device->super.name,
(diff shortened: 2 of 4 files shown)


bosilca:
Do you have any numbers that validate this need ? We tried this in the past, and even for short tasks with a lot of descendents didn't indicate there was a real need for shifting the completion to another thread (which comes with it's own set of problems).

devreal:
I have not measured the impact, no. In TTG we run multi-threaded MPI so the GPU thread ends up calling MPI. That's not great. Also, in practice the worker threads are idle most of the time, since they only select the device and then pass the task on to the device manager. Any task we push out of the manager will be picked up relatively quickly because there is not much else the worker threads are doing. Without it, in a single device run we're effectively serializing everything onto the device manager thread.

devreal:
@bosilca Can you elaborate on the problems you see with thread-shifting the completion?

bosilca:
None in the runtime, most in the profiling (which assumes the completion thread is the same as the submission or it does some O(N ^3) searches.

Last time we did something similar we had a cleaner approach, we reserved some hyper-threads as helpers for the task preparation and cleanup, and they were bound to the same resources as the threads they were helping. This allowed us to take advantage of the core-level cache coherence so the thread and its helpers were able to collaborate without locks and atomics. Despite that we were never able to show that the device in charge of the GPU was a starvation point, which is the only valid reason to try to delegate the completions.

devreal:
In DPLASMA, the GPU thread does not hit the network because DPLASMA runs with serialized MPI. In TTG, every thread can call into MPI, so the GPU thread will eventually end up there. Then all bets are off. It's also safe to assume that the worker threads are mostly idle so there are plenty of cycles to spare and the completion will be picked up quickly. SMT is not enabled everywhere so it's not an option we can rely on.

FWIW, the approach here follows the suggestion you made in https://github.com/ICLDisco/parsec/pull/687#issuecomment-2462576802.

devreal:
We can add YACK (yet another configuation knob) to disable the offloading if it becomes a problem with profiling.

bosilca:
Because SMT was not enabled everywhere and because we could not figure out a valid case where it was needed, the changes I mentionned above are still pending in one of the stale PRs imported from bitbucket.

I understand why you would need this in TTG, and I want to give you an escape route but not a global knob. Let's add a flag onto the taskclass incarnation to mark it as completion delegation. Then in the GPU code, if the completion is marked as delegate-ready, do what you do here, otherwise fallback on the traditional path. With this all PTG and DTD would keep working as before, and all TTG will be able to take this different path. In fact it will even allow you different behaviors for different incarnations.

devreal:
@bosilca I like the idea of controling the thread-shift per chore. Please check if this is what you had in mind.

parsec/mca/device/device_gpu.c
```diff
@@ -2573,6 +2498,16 @@ parsec_device_kernel_cleanout( parsec_device_gpu_module_t *gpu_device,
     return 0;
 }
 
+/**
+ * Returns true if the task's completion should be shifted to worker threads.
+ */
+static bool shift_completed_task(parsec_device_gpu_module_t* gpu_device, parsec_gpu_task_t* gpu_task)
+{
+    parsec_task_t* this_task = gpu_task->ec;
+    const __parsec_chore_t *chore = &this_task->task_class->incarnations[gpu_device->super.device_index];
```
bosilca: I don't think this is correct, the incarnations are ordered by device type not by device index. You can find the current chore in `this_task->selected_chore`, or you can parse the incarnations array looking for a chore with the same device_type as the gpu_device.

bosilca (commented):
This is great. Fix the chore accessor and it will be ready to be merged.

parsec/mca/device/device_gpu.c
bosilca:
```suggestion
    const __parsec_chore_t *chore = &this_task->task_class->incarnations[this_task->selected_chore];
```

parsec/mca/device/device_gpu.c
```diff
@@ -2782,8 +2713,17 @@ parsec_device_kernel_scheduler( parsec_device_module_t *module,
         goto remove_gpu_task;
     }
     parsec_device_kernel_epilog( gpu_device, gpu_task );
-    __parsec_complete_execution( es, gpu_task->ec );
+
```
bosilca, line 2717:
```suggestion
+#if defined(PARSEC_DEBUG_PARANOID)
+    /* Batched submissions are split back into per-task GPU completions before
+     * this point. If this ever triggers, completion shifting needs to iterate
+     * the whole gpu_task ring instead of handling only the head.
+     */
+    assert(parsec_gpu_task_is_singleton(gpu_task));
+#endif
```

parsec/mca/device/device_gpu.c
```diff
@@ -2782,8 +2713,17 @@ parsec_device_kernel_scheduler( parsec_device_module_t *module,
...
+        gpu_task->ec->status = PARSEC_TASK_STATUS_COMPLETE;
+        PARSEC_LIST_ITEM_SINGLETON(gpu_task->ec);
+        __parsec_schedule(es, gpu_task->ec, 1);
+    } else {
+        __parsec_complete_execution( es, gpu_task->ec );
+    }
     gpu_device->super.executed_tasks++;
+
  remove_gpu_task:
     PARSEC_DEBUG_VERBOSE(10, parsec_gpu_output_stream, "GPU[%d:%s]: gpu_task %p freed",
```
bosilca, line 2739: this comment, while accurate in itself, is not reflective of the execution path. We need to let the user know if a task completion has been delegated to some other place.

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
```diff
@@ -3951,29 +3939,21 @@ jdf_generate_function_incarnation_list( const jdf_t *jdf,
...
         string_arena_add_string(sa, "#if defined(PARSEC_HAVE_DEV_%s_SUPPORT)\n", dev_upper);
-        string_arena_add_string(sa, "    { .type     = PARSEC_DEV_%s", dev_upper);
-        if( NULL != batch_property) {
-#if PARSEC_HAVE_DEV_CAPABILITY_BATCH
-            string_arena_add_string(sa, " | PARSEC_DEV_CHORE_ALLOW_BATCH");
-#endif  /* PARSEC_HAVE_DEV_CAPABILITY_BATCH */
-        }
-        string_arena_add_string(sa, ",\n");
+        string_arena_add_string(sa, "    { .type     = PARSEC_DEV_%s,\n", dev_upper);
         if( NULL == dyld_property ) {
```
bosilca, line 3958:
```suggestion
        string_arena_add_string(sa, "      .flags    = PARSEC_CHORE_FLAG_NONE,\n");
        if( NULL == dyld_property ) {
```

parsec/interfaces/ptg/ptg-compiler/jdf2c.c
bosilca:
```suggestion
                            "      .flags    = PARSEC_CHORE_FLAG_NONE,\n"
```

bosilca (changes requested):
The current PR branch is based before the merged batching code, so the device path has changed around it. 

What happens today with batching:

- A GPU submit hook can call parsec_gpu_task_collect_batch() and create a ring of parsec_gpu_task_t objects.
- The execution event tracks that ring as one submitted GPU operation.
- When the event completes, any complete_stage is called first. For example, the batched GEMM test uses this only to release the batch slot/state, not to complete PaRSEC tasks.
- The ring is then pushed into the next GPU stream FIFO with parsec_gpu_stream_push_pending().
- The pop/stage-out stream drains the ring back into individual GPU tasks.
- Only after each individual task has passed parsec_device_kernel_pop() and parsec_device_kernel_epilog() does it reach the final complete_task block.

We need to make sure we do not shift completion at the batch event-completion level. That would risk only completing the head of the ring, or bypassing the per-task pop/epilog work.

Few other small changes are needed to fix all the tasks generation point in the PTG compiler. The should also be a point in the insert_task where the task's .flags  is set to PARSEC_CHORE_FLAG_NONE

parsec/mca/device/device_gpu.c
devreal: Added a comment in the shift-branch to log that the task was scheduled for completion somewhere else.

parsec/mca/device/device_gpu.c
devreal: Somehow the this is a diff of a diff. Added in a separate commit.

bosilca:
I was reading you comment today "In TTG, every thread can call into MPI, so the GPU thread will eventually end up there." and I wonder if this really solves your problem or just alleviate it in some cases. I think the root issue is that you have a thread that is in charge of the GPU, owns the lock, when it goes doing some MPI calls. For as long as it remains in MPI, there is no possible progress on the GPU, the context has been locked by a thread that is now away. One can assume that most MPI calls that can be issued there are non-blocking sends, which means most of the time the duration of the MPI call will be small, but that's not always the case (burst of small messages toward the same destination might overrun the available credits). Maybe that's the point where the GPU manager needs help, with handling external data movements, not local completions.

devreal:
@bosilca #668 introduced `PARSEC_DEV_CHORE_ALLOW_BATCH` that is OR'd into the chore's `.type`. Here we introduce chore flags. Should they be merged somehow? Seems like `PARSEC_DEV_CHORE_ALLOW_BATCH` should actually be a flag, or I can `PARSEC_CHORE_FLAG_SHIFT_COMPLETION` be OR'd into the `.type`?

PR #763, opened 2026-03-28
Throttle evictions to one w2r task at a time

devreal:
No need to go berserk and have multiple concurrent evictions going.

### DIFF
--- a/parsec/mca/device/device_gpu.c
+++ b/parsec/mca/device/device_gpu.c
@@ -18,6 +18,7 @@
 #include "parsec/scheduling.h"
 
 #include <limits.h>
+#include <threads.h>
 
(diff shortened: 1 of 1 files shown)


bosilca (changes requested):
A single w2r task can generate at most MAX_PARAM_COUNT transfers, which is a small number compared with the total number of times that exists on modern GPUs. If we only generate a single w2r at the time, we are likely to pay the w2r latency all the time, because MAX_PARAM_COUNT data are quickly reused.

Let's do instead a middleground. We allow up to MAX_W2R tasks to coexists (where MAX_W2R is an MCA parameter). Same logic as here, except that we allow more than one by default, but users can set MAX_W2R to 1 to get this sequential-looking data eviction.

devreal:
Superseded by #773, closing.

PR #765, opened 2026-04-20
Add Python bindings for PaRSEC (py-parsec)

AddiThurston:
Cython-based Python bindings exposing PaRSEC's Dynamic Task Discovery (DTD) and tiled matrix APIs to Python. Includes:

- py_parsec package: core, dtd, matrix, runtime, and tasks modules with both .pyx (Cython) and .py (ctypes) implementations
- Build system: setuptools-based build with pyproject.toml, integrated into CMakeLists.txt via optional PARSEC_PYTHON_BINDINGS flag
- Examples: stencil_1D, DTD GEMM, merge sort, redistribute (DTD & PTG)
- Tests: installation verification, DGEMM DTD accuracy, stencil 1D accuracy with pytest fixtures and MPI support
- Stencil test utilities: result printing and element access helpers in the C stencil test for cross-validation with Python implementation

Made-with: Cursor
Signed-off-by Addison Thurston on behalf of SLU's Omni Computing Lab


python/src/py_parsec/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+PARSEC_DEV_CPU  = _dtd.PARSEC_DEV_CPU
+PARSEC_DEV_CUDA = _dtd.PARSEC_DEV_CUDA
+
+parsec_redistribute_dtd = _lazy_get(_dtd, "parsec_redistribute_dtd")
+parsec_redistribute = _lazy_get(_dtd, "parsec_redistribute")
+
+# --- Matrix module (what you actually have today) ---
+from .matrix import ParsecMatrixBlockCyclic
+
+# 给你一个“名字兼容”，先别再导出 None
```
bosilca, line 44: english please.

python/src/py_parsec/dtd.py
```diff
@@ -0,0 +1,649 @@
...
+# Return codes
+PARSEC_HOOK_RETURN_DONE = 0
+PARSEC_HOOK_RETURN_ERROR = -1
+
+# Matrix types
+PARSEC_MATRIX_DOUBLE = 1
+PARSEC_MATRIX_TILE = 0
+
+
+class ParsecDTDContext:
```
bosilca, line 36: I wonder why we need a specialized class for the DTDcontext ? I know we have a type for the dtd_taskpool, but I don't think it is in use anywhere. I would like to keep the clean design where all supported DSL work in the same context and taskpool space.

python/src/py_parsec/core.py
```diff
@@ -0,0 +1,69 @@
+# Core PaRSEC functionality - Python wrapper
```
bosilca, line 1: All these do nothing. If the goal was to have them as fallback in case the Cython stubs are not build/loaded then they should include a guarded import.

python/src/py_parsec/dtd.py
```diff
@@ -0,0 +1,649 @@
...
+        return self._n
+    
+    @property
+    def _data_collection(self):
+        """Get data collection pointer for internal use"""
+        return self
+
+
+# Utility functions
+def parsec_dtd_unpack_args(task, *args):
```
bosilca, line 296: These capabilities do not have equivalent in the cython file.

bosilca (commented):
A quick review that was pending for a long time. I will take another stab at this later today.

QingleiCao:
Just saw messages here. Thank you for your comments. @bosilca

PR #766, opened 2026-05-08
Topic/dtd cleanups

bosilca:
A bunch of DTD cleanups and modernize the code, add more examples, and some documentation.


parsec/interfaces/dtd/insert_function.c
```diff
@@ -201,6 +214,139 @@ inline int parsec_dtd_task_is_local(parsec_dtd_task_t *task)
...
+
+    /*
+     * The per-flow remote-send bitmap is sized from the communicator, while a
+     * task class can be described before the taskpool has a context. Construct
+     * the mempools exactly once, at the first point where the taskpool context
+     * is known, so each local task carries only the rank words it really needs.
+     */
+    parsec_atomic_lock(&dtd_tp->task_class_lock);
+    if( !parsec_dtd_task_class_has_mempools(dtd_tc) ) {
+        parsec_dtd_task_class_construct_mempools(dtd_tp, dtd_tc);
```
devreal, line 312: This code is never reached because we are bailing out above

parsec/mca/device/device_gpu.c
```diff
@@ -898,7 +898,7 @@ parsec_device_data_reserve_space( parsec_device_gpu_module_t* gpu_device,
                                         gpu_device->super.device_index, gpu_device->super.name, task_name,
                                         gpu_elem);
                     for (j = 0; j <= i; ++j) {
-                        if (temp_loc[i] == NULL) continue;
+                        if (temp_loc[j] == NULL) continue;
```
devreal, line 900: Yuck!

.gitignore
```diff
@@ -1,5 +1,10 @@
+# Copyright (c) 2026 NVIDIA Corporation.  All rights reserved.
```
devreal: Not sure this should be here (since there is no UTK copyright header)

parsec/mca/device/device_gpu.c
bosilca: and it's been here for quite some time.

parsec/interfaces/dtd/insert_function.c
bosilca: We can use a taskclass across multiple taskpools, for as long as we are in the same parsec context. The first test is a quick way out, not protected just checking if the tasks mempools are there. The second one we do it only once we got the lock, because the creation can be done by multiple threads (because we allow task insertion from multiple context for as long as they don't use the same DTD tiles).

PR #769, opened 2026-05-10
Add installation instructions

Rabab53:


devreal:
@Rabab53 Can you say a bit about what this change does? Is this a tutorial of sorts?

PR #773, opened 2026-05-11
device/gpu: proactive eviction with adaptive percentage threshold

devreal:
Introduce a two-tier proactive GPU memory eviction mechanism to hide eviction latency and avoid task stalls on allocation failure.

Tier-1 (pre-flight in parsec_device_data_reserve_space): before walking each task's data flows, free clean LRU entries while zone utilisation exceeds mem_evict_threshold percent of total capacity.  Uses the new parsec_device_try_evict_lru_one() helper which handles the readers / ref-count / trylock / CAS-readers race conditions that the reactive path already handles, and detects full-cycle scans via a cycling_sentinel.

Tier-2 (parsec_device_kernel_scheduler): when the clean LRU is empty and zone pressure is above the threshold, proactively enqueue a D2H writeback task on exec_stream[1] so dirty-page eviction latency overlaps with the upcoming H2D and kernel stages rather than blocking the critical path.

Adaptive threshold: each device carries mem_evict_threshold, initialised to parsec_gpu_mem_evict_upper (default 95%).  When parsec_device_progress_stream returns PARSEC_HOOK_RETURN_NEXT for the kernel-push stream -- meaning every queued task failed to acquire memory, a true stall -- the threshold is stepped down by 5 percentage points toward parsec_gpu_mem_evict_lower (default 80%).  This is intentionally coarser than per-task adjustment: a single task failing does not signal a stall; only a full pass of the pending queue failing does.

New MCA parameters registered by the CUDA and Level Zero components:
  device_{cuda,level_zero}_mem_evict_upper  (default 95)
  device_{cuda,level_zero}_mem_evict_lower  (default 80)

All zone-pressure checks are guarded with #if !defined(PARSEC_GPU_ALLOC_PER_TILE) since that mode has no zone allocator; tier-2 falls back to the simple clean-LRU-empty condition in that mode.

Supersedes https://github.com/ICLDisco/parsec/pull/763


parsec/mca/device/cuda/device_cuda_component.c
```diff
@@ -161,6 +161,12 @@ static int device_cuda_component_register(void)
     (void)parsec_mca_param_reg_int_name("device_cuda", "max_number_of_ejected_data",
                                         "Sets up the maximum number of blocks that can be ejected from GPU memory",
                                         false, false, MAX_PARAM_COUNT, &parsec_gpu_d2h_max_flows);
+    (void)parsec_mca_param_reg_int_name("device_cuda", "mem_evict_upper",
+                                        "Upper threshold (percentage of total GPU zone capacity) at which proactive clean-LRU eviction and D2H writeback begin. When a task stalls waiting for memory, the per-device threshold is stepped down by 5 points toward mem_evict_lower.",
+                                        false, false, 95, &parsec_gpu_mem_evict_upper);
```
bosilca: I guess we assume there will never be a run where two devices will be enabled in same time (because you overwrite parsec_gpu_mem_evict_upper and parsec_gpu_mem_evict_lower). Why do you even register them at the device level instead of globally to apply to all devices ? I would move them out of here into the device.c

parsec/mca/device/device_gpu.c
```diff
@@ -2594,6 +2737,13 @@ parsec_device_kernel_scheduler( parsec_device_module_t *module,
         }
         assert(NULL == progress_task);
 
+        /* Every task in the pending queue failed to acquire GPU memory on this iteration:
+         * we are truly stalled.  Step the eviction threshold down so tier-1 evicts more
+         * aggressively starting from the next scheduling iteration. */
+        if( PARSEC_HOOK_RETURN_NEXT == rc &&
+            gpu_device->mem_evict_threshold - 5 >= parsec_gpu_mem_evict_lower )
+            gpu_device->mem_evict_threshold -= 5;
+
```
bosilca: Why "every task" ? This tests one task at the time, I don't see how you make the global assumption that all tasks failed to acquire. Thus, if I understand the code correctly, this will drive the eviction threshold quite rapidly down to the minimum.

parsec/mca/device/device_gpu.c
```diff
@@ -854,6 +944,24 @@ parsec_device_data_reserve_space( parsec_device_gpu_module_t* gpu_device,
...
+     * steps (floor: parsec_gpu_mem_evict_lower, default 80%) whenever the device stalled
+     * because the reactive path also failed to find memory. */
+    {
+        size_t total_capacity = (size_t)gpu_device->mem_nb_blocks * gpu_device->mem_block_size;
+        parsec_gpu_data_copy_t *cycling = NULL;
+        while( zone_in_use(gpu_device->memory) * 100 >
+               (size_t)gpu_device->mem_evict_threshold * total_capacity ) {
+            if( !parsec_device_try_evict_lru_one(gpu_device, &cycling) )
+                break;
+            data_avail_epoch = 1;
```
bosilca: I don't think this is correct because:
1. The epoch is never supposed to go backward, only monotonically increasing, and to be honest I don't even think we handle the case where it rolls over. 
2. going back to a stable value means that some tasks that would fail to acquire memory the second time will not be marked with epoch 1, so all future rollback of the epoch to 1 will never allow them to recheck the availability.
3. This will stall some tasks until some other external event dumps the device data_avail_epoch leading to random/unexpected/undesirable behaviors.

parsec/mca/device/cuda/device_cuda_component.c
devreal: Moved to global scope.

parsec/mca/device/device_gpu.c
devreal: Fixed, replaced with `data_avail_epoch++`

parsec/mca/device/device_gpu.c
```diff
@@ -2576,6 +2684,62 @@ parsec_device_kernel_scheduler( parsec_device_module_t *module,
...
+     * memory pressure exceeds the watermark, queue a D2H transfer on exec_stream[1]
+     * now so its latency overlaps with the upcoming H2D stage and kernel execution.
+     * This converts a potential blocking wait (dirty page eviction on the critical
+     * path) into an overlapped background transfer.
+     *
+     * In PARSEC_GPU_ALLOC_PER_TILE mode there is no zone allocator; fall back to the
+     * simple condition of clean LRU being empty. */
+    if( !parsec_list_nolock_is_empty(&gpu_device->gpu_mem_owned_lru) &&
+         parsec_list_nolock_is_empty(&gpu_device->gpu_mem_lru) ) {
+#if !defined(PARSEC_GPU_ALLOC_PER_TILE)
```
bosilca, line 3578: These w2r tasks are not counted as part of the device->mutex, so the runtime does not track their existence. I can imagine a scenario where there is a single user task pending, and we decide to add one or more w2r tasks. The defive mutex will be one so as soon as the user task completes, the thread can consider its work on the GPU done and bail out, leaving the w2r tasks in limbo. If that happens in the middle of the execution, another thread will pick up the GPU and handle the w2r tasks, but if we are at the end of an application, we will now have tasks in flight that would never complete, because without user tasks there will be no GPU manager.

Maybe the solution is to count these w2r tasks as part of the device->mutex.

parsec/mca/device/device_gpu.c
```diff
@@ -2595,8 +2759,19 @@ parsec_device_kernel_scheduler( parsec_device_module_t *module,
...
-        gpu_task = parsec_gpu_create_w2r_task(gpu_device, es);
+        /* If we can extract data go for it, otherwise try to drain the pending tasks.
+         * Skip if evictions are already in flight to avoid a storm of D2H tasks.
+         * If there are no in-flight evictions and nothing left to queue from the dirty
+         * LRU, we are truly stuck: step the eviction threshold down so tier-1 starts
+         * freeing pages earlier on the next attempt. */
+        if( 0 == gpu_device->mem_evict_in_flight ) {
+            size_t _sel = 0;
+            gpu_task = parsec_gpu_create_w2r_task(gpu_device, es, SIZE_MAX, &_sel);
+            if(gpu_device->mem_evict_threshold - 5 >= parsec_gpu_mem_evict_lower) {
```
bosilca, line 3658: The threshold manipulation happens irrespective if the parsec_gpu_create_w2r_task was able to create some tasks or not. This is very aggressive, because you will lower the threshold to the minimum very quickly, way before there is any real pressure on the memory.  Why lowering it so aggressively ?

parsec/mca/device/device.c
```diff
@@ -313,6 +319,18 @@ int parsec_mca_device_init(void)
...
+                                        "Upper threshold (percentage of total GPU zone capacity) at which proactive "
+                                        "clean-LRU eviction and D2H writeback begin. When the device is truly stalled "
+                                        "(no in-flight evictions and no dirty pages left to queue), the per-device "
+                                        "threshold is stepped down by 5 points toward device_mem_evict_lower.",
+                                        false, false, 95, &parsec_gpu_mem_evict_upper);
+    (void)parsec_mca_param_reg_int_name("device", "mem_evict_lower",
+                                        "Lower bound (percentage of total GPU zone capacity) to which the adaptive "
+                                        "eviction threshold may be reduced after repeated stalls.",
+                                        false, false, 80, &parsec_gpu_mem_evict_lower);
+#endif  /* PARSEC_HAVE_CUDA || PARSEC_HAVE_HIP || PARSEC_HAVE_LEVEL_ZERO */
```
bosilca: you should check that parsec_gpu_mem_evict_lower and parsec_gpu_mem_evict_upper are in the valid range [0..100%]. I would also register them all the time, but mark them as read-only (the second `false` in the call) when support does not exists.

parsec/mca/device/cuda/device_cuda_component.c
devreal: This is fixed, moved out of the device-specific components.

parsec/mca/device/device_gpu.c
devreal: Fixed, we're now accounting for proactive eviction tasks. Reactive eviction tasks are not accounted since they are only scheduled if there are other tasks that depend on the memory to become available.

parsec/mca/device/device_gpu.c
```diff
@@ -2748,6 +2856,20 @@ parsec_device_kernel_cleanout( parsec_device_gpu_module_t *gpu_device,
     return 0;
 }
 
+/**
+ * Returns false if at least one of the execution stream fifos has pending tasks.
+ * Otherwise, returns true, meaning that the GPU has no new work to schedule into the stream.
+ */
+static bool gpu_device_exec_streams_fifo_empty( parsec_device_gpu_module_t *gpu_device )
+{
+    for (int i = 2; i < gpu_device->nb_exec_streams; i++) {
```
bosilca:
```suggestion
    for (int i = 2; i < gpu_device->num_exec_streams; i++) {
```

parsec/mca/device/device_gpu.c
```diff
@@ -2748,6 +2856,20 @@ parsec_device_kernel_cleanout( parsec_device_gpu_module_t *gpu_device,
...
+        if( !parsec_list_nolock_is_empty(gpu_device->exec_streams[i].fifo_pending) ) {
```
bosilca:
```suggestion
        if( !parsec_list_nolock_is_empty(gpu_device->exec_stream[i].fifo_pending) ) {
```

parsec/mca/device/transfer_gpu.c
```diff
@@ -260,17 +268,22 @@ parsec_gpu_create_w2r_task(parsec_device_gpu_module_t *gpu_device,
             PARSEC_DEBUG_VERBOSE(10, parsec_gpu_output_stream,  "D2H[%d:%s] task %p:\tdata %d -> %p [%p] readers %d",
                                  gpu_device->super.device_index, gpu_device->super.name, (void*)d2h_task,
                                  nb_cleaned, gpu_copy, gpu_copy->original, gpu_copy->readers);
+            _selected += gpu_copy->original->nb_elts;
             nb_cleaned++;
-            if (MAX_PARAM_COUNT == nb_cleaned)
+            if( MAX_PARAM_COUNT == nb_cleaned || _selected >= required_size )
```
bosilca, line 287: required_size is in bytes, but if I recall correctly nb_elts is in number of datatypes. There is another similar check in the completion path. Please check both.

devreal:
Merged in master. Will test it later this week.

bosilca:
You scared me here for a moment. "Merged in master" for a PR that does not build properly ... I guess you meant "rebased on master"

PR #783, opened 2026-06-04
Make PTG compiler optional

devreal:
Not all consumers of PaRSEC require the PTG compiler. Setting `PARSEC_BUILD_PTGPP=OFF` will build only the runtime component needed for DTD and direct users of the PaRSEC runtime.

Also fixes a duplicated installation of apply.h.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,8 @@ mark_as_advanced(BUILD_TOOLS)
 option(BUILD_PARSEC
        "Compile the PaRSEC framework" ON)
 mark_as_advanced(BUILD_PARSEC)
+option(PARSEC_BUILD_PTGPP
+       "Build the parsec-ptgpp JDF compiler" ON)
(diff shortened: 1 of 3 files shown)


bosilca (commented):
I think you also need to remove it from the generated PaRSECConfig.cmake to let any potential users know the ptgpp is missing in this install.

My big problem with this is that you now get a crippled data helper library, everything based on JDF will not exists, such as basic collectives, reduce/broadcast, or more elaborate features, the data redistribution.

So I guess the real question is what is the harm of builbin the ptgcc most of the time, even if you don't plan to use it ?

devreal:
Mhh, I was having issues with executing the PTG compiler during build when enabling address sanitizer. Not because of an issue in PaRSEC but because ASAN is stuck in init. Then I thought that I don't really need the compiler. But having an incomplete PaRSEC library also isn't great. Closing for now.

PR #784, opened 2026-06-04
Sort incoming device tasks using heap

devreal:
Currently, sorting of device tasks is broken (they don't actually contain a priority field) and linear over a slice of the input list. Instead, we can use a max-heap to keep a sorted array of device tasks, push all new tasks into the heap, and take out the highest priority task.

~Complexity of inserting the ring into the heap is O(N+M) or O(N*logM), depending on the whether we just insert elements one by one or rebuild from scratch (beneficial if we insert more tasks than exist already).~

We can detach the full lifo at once to avoid repeated atomic operations by the management thread.

Also, use nolock variants on stream-local lists and change to lifo for the shared queue to avoid the lock of the fifo. They are only modified by the device management thread. Inter-thread communication happens through the pending queue.

~Once tasks are trickling through the system they do not need to be sorted anymore.~


devreal:
Looking into merging the two heap implementations. Didn't see the existing one because it wasn't in the class/ directory.

parsec/maxheap.h
```diff
@@ -2,42 +2,58 @@
...
-/* main struct holding size info and ID */
-typedef struct parsec_heap_s {
-    parsec_list_item_t list_item; /* to be compatible with the lists */
-    unsigned int size;
-    unsigned int priority;
-    parsec_task_t * top;
-} parsec_heap_t;
+/* forward declaration — full definition is in parsec/parsec_internal.h */
+struct parsec_task_s;
+typedef struct parsec_task_s parsec_task_t;
```
bosilca: This is already defined in runtime.h, I don't like redeclaration if I can stay away from them. I was about to suggest to use 'struct parsec_task_s' in the rest of the file, but I just noticed this is not used in the rest of this file. I think it is safe to remove.

parsec/class/lifo.h
```diff
@@ -232,6 +233,35 @@ LIFO_STATIC_INLINE int parsec_lifo_nolock_is_empty( parsec_lifo_t* lifo ) {
...
 
+/* Convert a NULL-terminated singly-linked list (list_next chain, as stored in a
+ * LIFO) into a doubly-linked ring.  Calls PARSEC_ITEM_DETACH on each item to
+ * balance the PARSEC_ITEM_ATTACH done by parsec_lifo_push.
+ *
+ * In PARSEC_DEBUG_PARANOID mode PARSEC_ITEM_DETACH overwrites both list_prev
+ * and list_next with 0xdeadbeef, so each pointer must be (re)set AFTER the
+ * detach call rather than before. */
+static inline parsec_list_item_t *
+parsec_lifo_chain_to_ring(parsec_list_item_t *head)
```
bosilca: I don't understand why we need to reconstruct the entire ring. The difference between a chain and a ring is that the chain has the two heads disconnected. Thus, transforming a chain into a ring should not alter any intermediary elements, but just find the two extremes and link them together.

The refcount transfers, the only thing that needs to be updated is the belongs_to field but only in debug mode.

parsec/maxheap.c
```diff
@@ -13,8 +14,14 @@
 #include "parsec/maxheap.h"
 
 #include <stdlib.h>
+#include <stddef.h>
```
bosilca, line 17: This maxheap is only used in the ltq scheduler. As we have a good heap replacement, maybe we should move this data structure near the only place it is used (aka. in the ltq scheduler).

bosilca (commented):
Overall a good addition, and a significant fix.

Few nitpicks:
- You remove spaces at the end of the lines. This is uncalled for in the context of another commit.
- I just realized that the stream fifo_pending is not ordered by priority anymore. Indeed, now that tasks can bypass the input stream (which would have guaranteed the original, ordered by priority, order), they will land in the execution streams unordered. I don't think this is a problem, but your last point in the PR description is incorrect.
- you mention O(N+M) cost, but I don't see where that would be. The current push_ring code is o(N log(N+M)).

parsec/maxheap.c
devreal: It's also used in the hbbuffer. Leaving it where it is for now.

parsec/class/lifo.h
devreal: I thought it was safer to make it a ring but the path from the lifo to the heap is short, so probably not needed.

parsec/maxheap.h
devreal: Fixed.

devreal:
I replaced the ring with a chain. To avoid iterating over it, we can set the priority when scheduling the device task. We only set the priority if the device task priority is negative (set in constructor), otherwise we take what the caller has set.

devreal:
The complexity was a leftover from when I had this implemented using an array. I removed references to the complexity.

parsec/class/lifo.h
```diff
@@ -356,6 +371,29 @@ LIFO_STATIC_INLINE parsec_list_item_t* parsec_lifo_try_pop( parsec_lifo_t* lifo
...
+    parsec_atomic_wmb();
+    parsec_list_item_t *item = old_head.data.item;
+#if defined(PARSEC_DEBUG_PARANOID)
+    while (item != NULL) {
+        parsec_list_item_t *next = (parsec_list_item_t *) item->list_next;
+        PARSEC_ITEM_DETACH(item);
+        item = next;
+    }
+#endif
+    PARSEC_CHAIN_DETACH(item);
```
bosilca, line 390: **P1 — this breaks `PARSEC_DEBUG_PARANOID` chain detachment.** In the CAS128 implementation, the preceding loop advances `item` to `NULL`, so this call receives `NULL` and the function returns `NULL` after atomically removing every queued task. In the LLSC and lock-backed implementations, `PARSEC_CHAIN_DETACH` calls `PARSEC_ITEM_DETACH` on each element, which poisons `list_next/list_prev` with `0xdeadbeef`; `parsec_heap_push_chain()` then attempts to traverse those poisoned links.

Please preserve the chain links until heap insertion has saved each `next` pointer, while still performing the paranoid ownership/refcount bookkeeping. The duplicate CAS128 detach loop should also be removed. A paranoid-build test that detaches and inserts a multi-item chain would catch both failure modes.

parsec/class/parsec_heap.c
```diff
@@ -0,0 +1,204 @@
...
+
+    /* Sift down: swap last with the larger child until heap order is restored.
+     * Track parent and which side we came from (no extra allocation needed). */
+    parsec_list_item_t *bubbler = last;
+    parsec_list_item_t *par = NULL;
+    int from_right = 0;
+    while (1) {
+        parsec_list_item_t *left  = HLEFT(bubbler);
+        parsec_list_item_t *right = HRIGHT(bubbler);
+        int go_left  = (left  && heap_cmp(heap, left,  bubbler) > 0 &&
```
bosilca, line 175: **P1 — equal-priority tasks can starve indefinitely.** Equal elements neither sift up nor sift down. After a pop, the newest last leaf replaces the root and remains there because neither child is strictly greater. If one or more equal-priority tasks arrive before each subsequent pop, newly arrived work keeps cycling through the root while the older backlog remains below it indefinitely. Most device tasks use the same default priority, so this is a common arrival pattern, not just an edge case.

Please provide stable/FIFO ordering within a priority, for example with a monotonic enqueue sequence used as the secondary comparison key or per-priority FIFO queues. Add a test that continuously inserts and pops equal-priority elements while verifying that the initial backlog drains.

parsec/mca/device/device_gpu.c
```diff
@@ -2352,7 +2285,7 @@ parsec_device_send_transfercomplete_cmd_to_device(parsec_data_copy_t *copy,
                          current_dev->device_index, current_dev->name, gpu_task->ec->data[0].data_out,
                          gpu_task->ec->data[0].data_out->super.super.obj_reference_count,
                          dst_dev->device_index, dst_dev->name);
-    parsec_fifo_push( &(((parsec_device_gpu_module_t*)dst_dev)->pending), (parsec_list_item_t*)gpu_task );
+    parsec_lifo_push( &(((parsec_device_gpu_module_t*)dst_dev)->pending), (parsec_list_item_t*)gpu_task );
```
bosilca, line 2293: **P1 — this direct enqueue bypasses priority inheritance.** The wrapper is constructed with `gpu_task->priority == -1`; only `gpu_task->ec->priority` is assigned `INT32_MAX`. The heap compares the wrapper field, and inheritance occurs only in `parsec_device_kernel_scheduler()`, which this path does not call. Consequently this supposedly urgent D2D-completion command enters at the lowest priority and can starve behind normal work, delaying source-reader release and GPU-memory reclamation.

Please set `gpu_task->priority = INT32_MAX` here, or centralize priority resolution in a common pending-enqueue helper used by all direct and scheduler enqueue paths. This needs a test covering the priority of directly enqueued D2D completion work.

parsec/maxheap.h
```diff
@@ -2,42 +2,55 @@
...
  *                         reserved.
+ * Copyright (c) 2026      Stony Brook University. All rights reserved.
  */
 
 #ifndef MAXHEAP_H_HAS_BEEN_INCLUDED
 #define MAXHEAP_H_HAS_BEEN_INCLUDED
 
 #include "parsec/parsec_config.h"
-#include "parsec/class/list_item.h"
+#include "parsec/class/parsec_heap.h"
```
bosilca, line 12: **P1 — the default development-header installation is now unusable.** `parsec/maxheap.h` is installed when `PARSEC_WITH_DEVEL_HEADERS` is enabled (the default), but the newly included `parsec/class/parsec_heap.h` is not present in any install list. External compilation including `parsec/maxheap.h` will therefore fail.

Please install the new dependency and add an install-tree compilation test. This header also replaces the externally exposed `parsec_heap_t` with `parsec_task_heap_t`; that source/API break should either be avoided (for example by giving the new generic heap a non-conflicting name) or handled as an intentional compatibility/version change.

bosilca (changes requested):
The core heap preserves max-priority ordering in ordinary operation, and the current CI is green. I am requesting changes because the current head still has four blockers: paranoid builds lose or crash on detached task chains, equal-priority intake can indefinitely starve existing work, directly enqueued D2D-completion tasks receive the wrong priority, and the default development-header installation is broken. The inline comments include suggested fixes and targeted regression tests.

devreal:
@bosilca All previous comments should be addressed. Please check again.

PR #786, opened 2026-07-17
Add span_alloc for the size of gpu data allocation

QingleiCao:
This is useful when the GPU copy memory grows during runtime. It may reserve more memory than currently needed, but it avoids the overhead of GPU memory reallocations. 

It's not the best solution, as it's not good for low-rank on GPU.


QingleiCao:
Not a good solution for distributed env

QingleiCao:
It seems this PR is not needed, as we can reallocate memory in the BODY on GPU for both CPU and GPU.

PR #790, opened 2026-07-21
Don't singleton batched tasks rings before executing them

### LINKED ISSUE #788: Batched kernel and PARSEC_HOOK_RETURN_AGAIN
## Describe the bug
If the last return of the head of a batched task ring was `PARSEC_HOOK_RETURN_AGAIN` then we drop the ring and execute the task. That way we lose the non-lead members of the ring so they will never be touched again.

See https://github.com/ICLDisco/parsec/blob/master/parsec/mca/device/device_gpu.c#L2640 where we jump to the `schedule_task` label and the callback will make the task a singleton. Instead, we should splice the ring into the pending_tasks list and get the head out.

devreal:
If the task returns PARSEC_HOOK_RETURN_AGAIN, we must not break the ring and instead hand it back to the execution step.

Alternative is to feed the ring back into the pending queue but that potentially increases the number of events since now every task is treated separately. It is up to the application to handle the case where the state of tasks within a ring diverge.

Fixes #788

### DIFF
--- a/parsec/mca/device/device_gpu.c
+++ b/parsec/mca/device/device_gpu.c
@@ -2918,8 +2918,13 @@ parsec_device_kernel_exec( parsec_device_gpu_module_t      *gpu_device,
     /* The submit hook may turn gpu_task into a batch ring. Start from a clean
      * singleton so stale list links left by release-mode list operations cannot
      * be mistaken for a preexisting ring.
+     * However, if the task returned PARSEC_HOOK_RETURN_AGAIN, we must preserve the
+     * ring as it may contain the follower tasks of the batch.
      */
-    PARSEC_LIST_ITEM_SINGLETON(&gpu_task->list_item);
+    if (gpu_task->last_status != PARSEC_HOOK_RETURN_AGAIN) {
+        /* The task is being rescheduled, we need to reset the status */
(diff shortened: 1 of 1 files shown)


bosilca:
First, I don't see why a task that batches other tasks should be allowed to return AGAIN. What is the conceptual meaning behind this ? 

That being said if a task does return AGAIN, it should lose all the batched tasks, because nobody wants a hot task with a bunch of potentially batched tasks lingering around.

I assume you have a use case for this ? Can you detail please.

devreal:
In TTG, all device tasks return AGAIN. That's the design we settled on, so that we can suspend the task and submit additional kernels and/or transfers. We hardly ever do submit additional work, but it's what we support today.

We can splice the ring back into the pending queue (my initial approach). We have to assume that all tasks returned AGAIN, and schedule them again individually, possibly as part of another batch. In that case, having a separate return value of DONE would be helpful to avoid scheduling another event for the task even if that particular task completed. Either way, the application has to deal with the fact that some tasks coming back at it may already be complete.

bosilca:
Because of the coroutines ? I need to think a little more about this it is different than the model we used so far.

devreal:
Superseded by #792, closing

PR #791, opened 2026-07-23
Batched device memcpy for CUDA/HIP

devreal:
Since CUDA 12.8 we can use batched memory copy to perform multiple operations in a single CUDA call. For tasks with multiple stage in/out flows this can be beneficial. Prominent example
are eviction tasks. There is no reason to order pushouts of a single task.

In practice, I am observing higher bandwidth for small batched data (9x2GB/s vs 1x12GB/s for 9 8KB chunks).

### DIFF
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -356,7 +356,9 @@ static int parsec_cuda_memcpy_async(struct parsec_device_gpu_module_s *gpu, stru
  *    the semantics of the individual cudaMemcpyAsync calls this replaces); location
  *    hints are left unset since they only matter for managed memory, which PaRSEC
  *    does not use for these buffers. The cudaMemcpyBatchAsync signature dropped its
- *    failIdx parameter in CUDA 13.0; guard both forms.
+ *    failIdx parameter in CUDA 13.0; guard both forms. Neither form is asked to
+ *    report failIdx: this function's signature (parsec_device_memcpy_multi_async_fn_t)
+ *    has no partial-submission out-parameter, so there is nowhere to forward it to.
  */
 static int parsec_cuda_memcpy_multi_async(struct parsec_device_gpu_module_s *gpu, struct parsec_gpu_exec_stream_s *gpu_stream,
                                           void **dsts, void **srcs, size_t *sizes,
@@ -385,7 +387,7 @@ static int parsec_cuda_memcpy_multi_async(struct parsec_device_gpu_module_s *gpu
                                       &attrs, &attrsIdx, 1 /* numAttrs */,
                                       cuda_stream->cuda_stream);
 #else
-    cudaStatus = cudaMemcpyBatchAsync(dsts, srcs, sizes, (size_t)nb_items,
+    cudaStatus = cudaMemcpyBatchAsync(dsts, (const void *const *)srcs, sizes, (size_t)nb_items,
                                       &attrs, &attrsIdx, 1 /* numAttrs */,
                                       NULL /* failIdx: unused, we don't need to know which item failed */,
                                       cuda_stream->cuda_stream);
--- a/parsec/mca/device/cuda/device_cuda_module.c
+++ b/parsec/mca/device/cuda/device_cuda_module.c
@@ -343,6 +343,58 @@ static int parsec_cuda_memcpy_async(struct parsec_device_gpu_module_s *gpu, stru
     return PARSEC_SUCCESS;
 }
 
+#if CUDART_VERSION >= 12080
+/**
+ * @brief Native implementation of parsec_device_memcpy_multi_async_fn_t using
+ *    cudaMemcpyBatchAsync (introduced in CUDA 12.8). @p dsts/@p srcs/@p sizes are
+ *    forwarded to cudaMemcpyBatchAsync as-is -- no repacking -- since its parameter
+ *    shape is exactly the parallel-array layout parsec_device_memcpy_multi_async_fn_t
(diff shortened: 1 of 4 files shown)


parsec/mca/device/device_gpu.c
```diff
@@ -1611,9 +1611,30 @@ parsec_device_data_reserve_space( parsec_device_gpu_module_t* gpu_device,
...
+ *    Used by backends that do not (yet) provide a native batched copy primitive.
+ */
+int
+parsec_device_generic_memcpy_multi_async(parsec_device_gpu_module_t *gpu, parsec_gpu_exec_stream_t *gpu_stream,
+                                         void **dsts, void **srcs, size_t *sizes,
+                                         parsec_device_transfer_direction_t *directions, int nb_items)
+{
+    int ret, rc = PARSEC_SUCCESS;
+
+    for(int i = 0; i < nb_items; i++) {
```
bosilca, line 1630: This loop is too carefree. I know that if any error is returned everything will fail, but if we try to expose the faults at least we should do it properly. Here if anything fails we return the error but we continue to submit all the others copies. However, once we return from this function with an error we call parsec_device_kernel_push_release_readers_on_failure directly, releasing the readers for the entire mask (and that's one problem), and then return without enquing an event (which allow the data use right away on another stream). 

The interface needs either atomic submission semantics, a returned submitted mask, or the old per-flow path for backends without native batching.

parsec/mca/device/device_gpu.c
```diff
@@ -2730,6 +2759,31 @@ parsec_device_progress_stream( parsec_device_gpu_module_t* gpu_device,
...
+    for(uint32_t i = 0; i < gpu_task->nb_flows; i++) {
+        if( !(transfer_mask & (1U << i)) ) continue;
+        parsec_data_copy_t *src = gpu_task->flow_info[i].source;
+        const parsec_flow_t *flow = gpu_task->flow_info[i].flow;
+        if( (NULL == src) || (NULL == flow) ) continue;
+        if( !(flow->flow_flags & PARSEC_FLOW_ACCESS_READ) ) continue;
+        parsec_device_module_t *src_dev_mod = parsec_mca_device_get(src->device_index);
+        if( (NULL == src_dev_mod) || !PARSEC_DEV_IS_GPU(src_dev_mod->type) ) continue;
+        int readers = parsec_gpu_data_copy_release_reader((parsec_device_gpu_module_t*)src_dev_mod, src, 1);
+        assert(readers >= 0);
```
bosilca: releases a GPU reader without locking its original or coordinating with the source GPU manager. When this removes the last reader, parsec_gpu_data_copy_release_reader() can relink the copy into a source-device LRU.
The normal D2D completion path serializes with the source manager, locks the original, and updates data_avail_epoch. The failure path should reuse that protocol.

bosilca (commented):
This is a good idea, but the scope of the current implementation is narrow and only benefit to tasks with small data and only if the copy latency is important, which means for tasks that cannot keep the GPU busy (aka. not compute intensive). It also work per task, so at best it batches MAX_PARAM items.

If we can merge this with the task batching we could have a much stronger support for small compute-trivial tasks.

parsec/mca/device/cuda/device_cuda_module.c
```diff
@@ -343,6 +343,58 @@ static int parsec_cuda_memcpy_async(struct parsec_device_gpu_module_s *gpu, stru
...
+#if CUDART_VERSION >= 13000
+    cudaStatus = cudaMemcpyBatchAsync(dsts, (const void *const *)srcs, sizes, (size_t)nb_items,
+                                      &attrs, &attrsIdx, 1 /* numAttrs */,
+                                      cuda_stream->cuda_stream);
+#else
+    cudaStatus = cudaMemcpyBatchAsync(dsts, srcs, sizes, (size_t)nb_items,
+                                      &attrs, &attrsIdx, 1 /* numAttrs */,
+                                      NULL /* failIdx: unused, we don't need to know which item failed */,
+                                      cuda_stream->cuda_stream);
+#endif
```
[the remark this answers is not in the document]
devreal: Fixed.

parsec/mca/device/device_gpu.c
devreal: Factored out the d2d release and used it here. Please recheck

parsec/mca/device/device_gpu.c
devreal: Bailing out on error now. It's hard to reliably report which operation failed though (the new CUDA API doesn't report it anymore). I'm not sure it's important to know that. Repeating all transfers would do no harm in terms of correctness.

devreal:
> If we can merge this with the task batching we could have a much stronger support for small compute-trivial tasks.

I'm torn on this. I agree that batching memcpy for batched tasks could be helpful. However, for tasks with large inputs we don't gain much (since a single transfer can saturate the bandwidth) already and in fact may harm performance because the tasks are delayed collectively waiting for larger transfers. 

For small transfers and kernels this would be helpful, although we probably want to cap the maximum number of batched transfers (for the same reason as above).

I would suggest we take what we have here as a first step and explore more batching later. Small kernel are a problem that is hitting us in MADNESS so it is something I'm looking at. If we can benefit from task-level batched inputs I will explore that further.

devreal:
One way this would be helpful for us is if we could transfer the ring of batched tasks from execute to pushout and batch their outputs. Not sure if #792 covers that. 

Alternatively, we could just collect a bounded number of `PARSEC_DEV_CHORE_ALLOW_BATCH` tasks in the pushout and combine them that way. That should be easy to do but runs the risk of delaying successor releases because of large transfers.

PR #792, opened 2026-07-23
Allow AGAIN while batching

bosilca:
As an alternative to #790 

This PR fixes GPU batch ownership across all submit-hook results.

**AGAIN** preserves the complete ring for coroutine-style continuation.
**NEXT** restores followers and retries the singleton head.
**ASYNC** transfers every execution context to the hook.
**ERROR** and **DISABLE** clean the complete failed batch.

 - Stream insertion preserves task priority.
 - Batch callbacks gain explicit ACCEPT, REJECT, and STOP actions.
 - Profiling emits one start/end pair per logical task across repeated AGAIN.
 - GPU manager election preserves next_task locality for non-manager workers.
 - Documentation now includes AGAIN, ASYNC, and manual-disbanding examples.
 - A deterministic CUDA regression covers repeated AGAIN, ASYNC, NEXT, priority ordering, callback STOP, profiling, and exact-once completion.

Device-wide **DISABLE** recovery remains intentionally out of scope.

### DIFF
--- a/docs/doxygen/task-batching.md
+++ b/docs/doxygen/task-batching.md
@@ -1,3 +1,7 @@
+<!--
+Copyright (c) 2026 NVIDIA Corporation.  All rights reserved.
+-->
+
 Task Batching {#task_batching}
 ==============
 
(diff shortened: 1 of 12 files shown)


GPU_BATCH_TASKS_TODO.md
```diff
@@ -0,0 +1,351 @@
+<!--
```
devreal: Should this file be in here? Seems like an AI scratchpad. All items seem to be ticked off.

devreal (commented):
Thanks @bosilca! Just a question whether the TODO file should be part of the PR?

GPU_BATCH_TASKS_TODO.md
bosilca: That was the work plan, it does not need to be in.

docs/doxygen/task-batching.md
```diff
@@ -129,20 +144,183 @@ gemm_kernel_cuda(parsec_device_gpu_module_t *gpu_device,
...
+`parsec_list_item_ring_chop()` reconnects the followers but deliberately leaves
+the removed head's links invalid. Restore the head as a singleton, refresh the
+followers' priority snapshots, and return the follower ring to the pending
+stream with its configured ordering policy. Call this helper from the submit
+hook after the collector returns, never from the collector callback while the
+pending FIFO is locked:
+
+```c
+static void
+requeue_batch_followers(parsec_gpu_exec_stream_t *gpu_stream,
```
devreal, line 272: Could PaRSEC provide us with a function to do that? I don't like asking people to deal with internal data structures. Something like `parsec_device_split_batch`?

PR #796, opened 2026-09-16
Add parsec_device_show_capabilities_start and parsec_device_show_capa…

QingleiCao:
Add parsec_device_show_capabilities_start() and parsec_device_show_capabilities_end() to report device statistics for an explicit execution interval.

Preserve whole-execution reporting when the APIs are unused. Print an unfinished interval during finalization, allow a second start to restart the interval, warn on an unmatched end, and suppress duplicate reports after a completed interval.

Snapshot cumulative device counters at the beginning of an interval and report their deltas at the end. Rebase the snapshot when statistics are reset to prevent counter underflow, and correctly handle MCA parameter index zero.

Add a PTG regression using block-cyclic matrix descriptors and real GPU data movement. Cover explicit, default, unfinished, restarted, reset, unmatched-end, and disabled reporting for both CUDA and HIP, including exact validation of the reported task and transfer statistics.


devreal:
I like the idea of providing an API for that. However, resetting the global counter is not intuitive. I would suggest adding an argument into which the current counters are read during start and then taking the diff to that snapshot during end, and freeing the memory that the counters occupy.

bosilca (changes requested):
1. The public names collide with a different MCA feature. `device_show_capabilities` already means “print hardware info at init.” These APIs control `device_show_statistics`. 
2. if you really want such an API it should be called show_statistics or something.
3. What does it give you that the already existing API, parsec_device_show_capabilities_end, does not already provides ?
4. The current API allows multiple statistics interval ongoing. This API allows only for one.
5. Rename the capability and the test.

